PR #15, opened 2015-08-06
Fixed bug in setting of segment_z_bin

adfrawley:


blackcathj:
Sorry for introducing this bug in the first place. Merging to the master branch

PR #25, opened 2015-08-21
Add property tag to SvtxTrack

blackcathj:
Attn: @mccumbermike 

Following Mike's request, I have added the property tagging capability to the SvtxTrack object. The goal would be avoiding multi-versioning of this object, while flexibly support adding more property tags onto the track. These new code are based on our experience of property map on PHG4Hit. 

This pull request only proposes the changes. I will leave it up to tracking expert to decide whether merge it or continue the development based on this branch (blackcathj/coresoftware:PropertyTagOnTrack). 

I moved a few scalar variable to the property map as a start. It will be trickier to move arrays into the property map. However we should be able to do so, if we define the interface function right and keep it transparent to the users. Again I will leave it to tracking experts to decide. 

If I did it right, these change should be transparent to users of SvtxTrack. First test looks good, which is easy to do in the ROOT command lines: 

```
[jinhuang@rcas2066 macro]$ root
root [0]   gSystem->Load("libg4hough.so");
root [1]  SvtxTrack tr
root [2] tr.identify()
SvtxTrack Object id: -1 
	Charge = 	1
	NDF = 	0
	Vertex x = 	0
	Vertex y = 	0
	Vertex z = 	0

# OK, init looks good. Go on to add more properties:
root [3] tr.setMomentum(23)
root [4] tr.setCharge(-1)
root [5] tr.setQuality(1)
root [6] tr.setChisq(2)
root [7] tr.setChisqv(3)
root [8] tr.setNDF(4)
root [9] tr.setDCA(5)
root [10] tr.setDCA2D(6)
root [11] tr.setDCA2Dsigma(7)
root [12] tr.set_x(8)
root [13] tr.set_y(9)
root [14] tr.set_z(10)
root [15] tr.identify()
SvtxTrack Object id: -1 chisq/dof: 0.5 
	Momentum = 	23
	Charge = 	-1
	Quality = 	1
	Chi Square = 	2
	Chi Square @ vertex = 	3
	NDF = 	4
	DCA = 	5
	DCA 2D = 	6
	DCA 2D Sigma = 	7
	Vertex x = 	8
	Vertex y = 	9
	Vertex z = 	10

```

### DIFF
--- a/simulation/g4simulation/g4hough/SvtxTrack.C
+++ b/simulation/g4simulation/g4hough/SvtxTrack.C
@@ -6,7 +6,7 @@ ClassImp(SvtxTrack)
 
 using namespace std;
 
-SvtxTrack::SvtxTrack() : phi(0.),d(0.),kappa(0.),z0(0.),dzdl(0.), x(0.), y(0.), z(0.), covariance(6,6)
+SvtxTrack::SvtxTrack() : phi(0.),d(0.),kappa(0.),z0(0.),dzdl(0.), covariance(6,6)
 {
   Reset();
 }
@@ -29,22 +29,22 @@ SvtxTrack::SvtxTrack(SvtxTrack *track) : covariance( *(track->getCovariance()) )
   }
   
   trackID = track->getTrackID();
-  momentum = track->getMomentum();
+//  momentum = track->getMomentum();
   for(int j=0;j<3;j++){
     mom3[j] = track->get3Momentum(j);
   }
 
-  charge = track->getCharge();
+//  charge = track->getCharge();
   isprimary = track->getPrimary();
   ispositive = track->getPositive();
-  quality = track->getQuality();
+//  quality = track->getQuality();
   
-  DCA = track->getDCA();
-  DCA2D = track->getDCA2D();
-
-  chisq = track->getChisq();
-  chisqv = track->getChisqv();
-  ndf = track->getNDF();
+//  DCA = track->getDCA();
+//  DCA2D = track->getDCA2D();
+//
+//  chisq = track->getChisq();
+//  chisqv = track->getChisqv();
+//  ndf = track->getNDF();
   
   for(int i=0;i<9;i++){
     scatter[i] = track->getScatter(i);
@@ -84,22 +84,22 @@ SvtxTrack::SvtxTrack(const SvtxTrack& track) : covariance( *(track.getCovariance
   }
   
   trackID = track.getTrackID();
-  momentum = track.getMomentum();
+//  momentum = track.getMomentum();
   for(int j=0;j<3;j++){
     mom3[j] = track.get3Momentum(j);
   }
 
-  charge = track.getCharge();
+//  charge = track.getCharge();
   isprimary = track.getPrimary();
   ispositive = track.getPositive();
-  quality = track.getQuality();
+//  quality = track.getQuality();
   
-  DCA = track.getDCA();
-  DCA2D = track.getDCA2D();
+//  DCA = track.getDCA();
+//  DCA2D = track.getDCA2D();
 
-  chisq = track.getChisq();
-  chisqv = track.getChisqv();
-  ndf = track.getNDF();
+//  chisq = track.getChisq();
+//  chisqv = track.getChisqv();
+//  ndf = track.getNDF();
   
   for(int i=0;i<9;i++){
     scatter[i] = track.getScatter(i);
@@ -113,9 +113,10 @@ SvtxTrack::SvtxTrack(const SvtxTrack& track) : covariance( *(track.getCovariance
     cal_cluster_e[i] = track.get_cal_cluster_e(i);
   }
 
-  x = track.get_x();
-  y = track.get_y();
-  z = track.get_z();
+  //! TODO delete these lines. They are now part of property map
+//  x = track.get_x();
+//  y = track.get_y();
+//  z = track.get_z();
 
 
   for (prop_map_t::const_iterator i = track.prop_map.begin();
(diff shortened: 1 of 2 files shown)


mccumbermike:
Thanks Jin,

I still want to clean up the SvtxTrack interface at some point before a
reco DST code freeze (things like bringing too much of the perigee
coordinate system forward instead of a more generic 4-vector, splitting
storage nodes for dca-tracks and primary tracks), but getting Alan's merge
from TPC land in first should come first and has priority. Map-ifying the
layer => cluster id should also be done as that's too storage wasteful to
persist permanantly.

While I don't think this particular merge should be disruptive to that,
let's keep this off master until he is in.

The SvtxHit and SvtxCluster should be more stable moving forward, so are
more ready for this kind of treatment. A revised SvtxCluster like this
could also allow the TPC storage of additional ints and floats without
necessitating a new object---that would be a nice feature.

Mike

--
Michael P. McCumber, PhD
505-709-8161

pinkenburg:
Hi Mike, Jin,

we should be a bit careful to not go overboard with those properties. 
For the hits we have a real technical reason to implement them - we do 
have really different fields depending on which detector type generates 
those hits and need flexibility to accommodate this (rather than adding 
many different versions). This is not the case for track objects and 
most likely for calorimeter towers. I am not opposed to add the 
properties so one can easily propagate values for testing or if one has 
really forgotten something but this should not be the way we keep most 
of the content of our objects. Those properties are more expensive 
computationally and definitely more opaque to other people.

For the tracks - it might be better to have really 2 versions for the 
silicon and the tpc tracking instead of merging two object which don't 
have too much in common. For the design of those objects we also have to 
think about what constitutes experimental data and what is added 
information from the sims so we can later (5 years from now...) have a 
seamless integration of real data and simulations.

Chris

On 8/22/2015 2:46 AM, Michael P. McCumber wrote:
> Thanks Jin,
>
> I still want to clean up the SvtxTrack interface at some point before a
> reco DST code freeze (things like bringing too much of the perigee
> coordinate system forward instead of a more generic 4-vector, splitting
> storage nodes for dca-tracks and primary tracks), but getting Alan's merge
> from TPC land in first should come first and has priority. Map-ifying the
> layer => cluster id should also be done as that's too storage wasteful to
> persist permanantly.
>
> While I don't think this particular merge should be disruptive to that,
> let's keep this off master until he is in.
>
> The SvtxHit and SvtxCluster should be more stable moving forward, so are
> more ready for this kind of treatment. A revised SvtxCluster like this
> could also allow the TPC storage of additional ints and floats without
> necessitating a new object---that would be a nice feature.
>
> Mike
>
> --
> Michael P. McCumber, PhD
> Los Alamos National Laboratory
> 505-709-8161
>
> On Fri, Aug 21, 2015 at 9:21 PM, Jin Huang <notifications@github.com> 
> wrote:
>
> > Assigned #25
> > <https://github.com/sPHENIX-Collaboration/coresoftware/pull/25> to
> > @mccumbermike <https://github.com/mccumbermike>.
> >
> > —
> > Reply to this email directly or view it on GitHub
> > 
> <https://github.com/sPHENIX-Collaboration/coresoftware/pull/25#event-389083314>
> > .
> >
>
> —
> <https://github.com/sPHENIX-Collaboration/coresoftware/pull/25#issuecomment-133647330>.
>

-- 

Upton, NY 11973-5000

mccumbermike:
I agree with Chris that we should use this technique only in a limited way.

But I disagree that the SvtxTrack should be replaced with N-objects for
every kind of track we will need. For instance I would like to have a node
of SvtxTrackMap and SvtxPrimaryTrackMap. A primary track would eliminate
the DCA fields (since it will be fit through the vertex) but be extended by
a vertex id (to allow proper backtracing). They will however share much in
common (4-vector + covariance + charge + many other things) and many
analyzers would probably like to switch between the two without caring
deeply about those fields.

So that requires a common interface---which we can implement in a number of
ways. One could solve this with a base class and two daughter classes.
Fine. But then we also have the issue of supporting both full silicon and
the TPC. Again most users won't care in detail how the track was measured.
They will only want a 4-vector near/on the vertex and to get on with their
lives. Okay so we decide that in order to store the dE/dx from the TPC on
the track, because someone wants to play with low pt PID. Now we need yet
another factor of two-objects--yuck. Oh and we need two separate classes of
stored clusters because the TPC might want to write out an additional field
or two--double yuck. Oh and the eval will have to be split between those
too--triple yuck. Then when I get around to writing a FastSim for the
charged particles, yay, N-times the number of FastSims just for the charged
particle tracking. Heavy use of tempting can reduce the number of lines of
code, but not the fundamental complexity of having so many different
objects.

Anyhow I'm pretty sure that will quickly get us into an unsupportable or at
least inconvenient mis-mash of objects--it is at the least a pain to
contemplate. If we don't go overboard by using the property tags for every
field but instead we utilize this flexibility only for fields that will not
overlap between usages (DCA, vertexid, dedx), I think we will be better
off. But of course I could have a blindspot in this vision.

When the tracking system down-selects to a particular technical design, we
can return to the v1 objects (still need to version these buggers), pare
down the property fields in a v2 implementation, and recover the <10-20%
storage that the property maps require for the non-overlapping fields.

How does this sound?

PS: I like the discussion these pull requests generate---discussions like
this one will improve our code. And the earlier exposure to ideas/pushback
will get us down the correct path faster.

--
Michael P. McCumber, PhD
505-709-8161

mccumbermike:
Jin I appreciate the example here. As discussed I plan to make a version of these ideas with a limited implementation and will bring it into master on a different pull request. I'm working now on revising storage of the SvtxTrack on a development branch. I'm going to close this one since we've decided to go a different direction.

PR #27, opened 2015-08-24
New Evaluators and Eval Objects

mccumbermike:
I think it is time to bring in the new evaluators as they are now ready for wider testing. Since these live in new SubsysReco objects, there is no effect on the old code until we swap them into the macros. Users can incorporate the Detector*Object*Eval objects into their own code to migrate between reconstructed information and Truth Information. The ntuple output has changed somewhat to be more consistent between ntuples (for instance g* always means truth) and put full reco information into the forward eval (truth => best reco) rows. 

The evaluators should be somewhat slower than reco code alone, but much faster than GEANT4. I've run them through call grind and taken care of obvious things. They cache answers for the lifetime of the object or until next_event() is called. This prevents repeated deep-inspection of the ancestry. I also added flags to turn of low level ntuple generation (like hits or clusters) in case users only want higher level output (like tracks). This will make smaller files, but probably not speed things up too much.

One issue I ran into that is rare enough not to be significant issue is the zero energy hits that I've already mentioned in a trouble ticket. But---if I'm correct about the underlying cause---requires a solution outside of the eval.

Anyhow, have a look and see if there are glaring design errors that I should handle before we merge.

EDIT: I should add that I also added a few non-const retrieval methods to the Raw*Container classes.


mccumbermike:
The patch from Chris for the zero edep hits seems to have fully resolved the runtime issue in the evaluation. We should be good to go now.

mccumbermike:
So... what's the policy here? Do other core developers expect me to authorize my own merge? I could do that but this way of working would violate the principle of code review---do we care? I'm trying to play by the fork-and-pull rules, but what exactly are our polices?

Also, should we be merging new features into an official pro-2 development branch? Or do we plan on just pushing these straight into master as I've requested here?

blackcathj:
Hi, @mccumbermike ,

I think when a pull request comes in, the package manager(s) should review, reach concensors and approve the change. Here is the list of manager that we can use as baseline:
https://github.com/sPHENIX-Collaboration/utilities/blob/master/utils/rebuild/packages.txt 

Therefore, this pull request #27  involves two packages, g4cemc managed by @pinkenburg and g4eval managed by @mccumbermike . Therefore, if both you agree, one of you should merge it as soon as possible. 

Meanwhile, I do like your proposed changes too and I think we should merge them in now. 

PS: two more point:

I think we should push to master. @pinkenburg is maintaining the pro-2 branch. And he would merge new features into pro build as the test production proceeds. 

I think there are many advantages for @sPHENIX-Collaboration/coresoftware-developpers to also use pull request to commit major new features: 1) it keep a record of a group of commit for specific purpose. 2) encourage a discussion through GitHub 3) keep a tag of the software through the pull request number.

PR #30, opened 2015-09-01
Adding BbcVertexMap and GlobalVertexMap tools

mccumbermike:
Ok, this will be one we should discuss at this week's meeting.

I needed a vertex position for the jet reco, but I also didn't really care if it was a precision vertex from the SVTX or a starter vertex from the BBC. We didn't have code to blend vertexes into a "master" global vertex, nor did we have a fast sim to smear truth vertexes for the BBC (which we presume we will also have in sPHENIX).

This merge adds both features. It has storage vertexes for BbcVertexMap and GlobalVertexMap and reco modules BbcVertexFastSimReco and GlobalVertexReco to fill both. Unlike PHENIX, it is up the to reco module to blend the various vertexes since this isn't a simple matter of prioritizing the sources, but actually computing a new averaged or other wise blended vertex. Both storage units support multiple collision reconstructions.

I think this is a better approach that with the fewer sPHENIX subsystems will work to our advantage, but since this will be used widely by sPHENIX analyzers and nearly everyone in PHENIX has used a global vertex of some kind or the other, I expect there will be opinions on how this task should be best accomplished---This is our chance to get this part *right* for sPHENIX so lets make use of the opportunity.

Ideas on this merge?

EDIT: This merge will require a simultaneous update to utils to add the new g4vertex & g4vertex_io library to the build which is a new dependency of g4jets.


simulation/g4simulation/g4vertex/BbcVertexMap.h
```diff
@@ -0,0 +1,49 @@
...
+public:
+
+  typedef std::map<unsigned int, BbcVertex*>::const_iterator ConstIter;
+  typedef std::map<unsigned int, BbcVertex*>::iterator            Iter;
+  
+  virtual ~BbcVertexMap() {}
+
+  virtual void identify(std::ostream &os = std::cout) const {os << "BbcVertexMap base class" << std::endl;}
+  virtual void Reset()                                      {}
+  virtual int  IsValid() const                              {return 0;}
```
blackcathj: @mccumbermike  are you trying to override PHObject::int isValid() const;? Or it is a new IsValid() function?

simulation/g4simulation/g4vertex/GlobalVertexReco.C
```diff
@@ -0,0 +1,165 @@
...
+      const BbcVertex* bbc = bbciter->second;      
+      GlobalVertex* vertex = new GlobalVertex_v1();
+
+      vertex->set_t(bbc->get_t());
+      vertex->set_t_err(bbc->get_t_err());
+      vertex->set_z(bbc->get_z());
+      vertex->set_error(2,0,0.0);
+      vertex->set_error(2,1,0.0);
+      vertex->set_error(2,2,pow(bbc->get_z_err(),2));
+      vertex->set_error(1,2,0.0);
```
blackcathj: @mccumbermike , a minor point: it appears GlobalVertex_v1 enforce that corvariant matrix is symmetric. Therefore, would you suggest a conversion that user code only need to fill half of the matrix or to fill the full matrix as this?

simulation/g4simulation/g4vertex/GlobalVertexReco.C
```diff
@@ -0,0 +1,165 @@
...
+
+    for (BbcVertexMap::ConstIter bbciter = bbcmap->begin();
+	 bbciter != bbcmap->end();
+	 ++bbciter) {
+      const BbcVertex* bbc = bbciter->second;      
+      GlobalVertex* vertex = new GlobalVertex_v1();
+
+      vertex->set_t(bbc->get_t());
+      vertex->set_t_err(bbc->get_t_err());
+      vertex->set_z(bbc->get_z());
```
blackcathj: @mccumbermike It appears x,y would be NAN for BBC-only vertex. As an alternative option, would you like to do so, or give a zero central value with a few mm sigma?
The advantage for a zero X,Y central value, rather than NAN, is that a user code can be ignorant of the source of vertex, and still use it with a rought x,y to estimate, for example tower eta,phi.

blackcathj:
@mccumbermike  I quickly checked through the code, it looks nice to me. It also compiles on my area too. I do not have an example macro to test run it though. Nevertheless, there was few questions when I go through the code, and I added them inline as shown above. 
Let's talk more in the meeting today and try to merge it soon.

simulation/g4simulation/g4vertex/BbcVertexMap.h
mccumbermike: This is a bug, I will have to fix it.

simulation/g4simulation/g4vertex/GlobalVertexReco.C
mccumbermike: Filling half (either half) is all that is necessary. A few of these fills are redundant, but they aren't expensive.

simulation/g4simulation/g4vertex/GlobalVertexReco.C
mccumbermike: As we discussed in the meeting, I will added x,y to the BBC vertex and change a bit how the global vertex is filled under a few of these various scenarios.

mccumbermike:
Okay, I've patched the isValid issue. I've added a simple truth vertex smearing module in GlobalVertexFastSim (settings for x,y,z,t included). And I've revised the algorithm for the GlobalVertexReco to better handle cases where the BBC finds a vertex but the SVTX did not (say a large z vertex values). There is still the open issue of patching utils simultaneously for the build or pulling those scripts into coresoftware.

PR #35, opened 2015-09-08
replace eta/phi bin based tower key in cluster by towerid from towers, store tower energy for easier access

pinkenburg:
Now the clusters follow the same scheme as our hits, cells and towers. No change in the clusters themselves. This made a small change in the evaluator necessary.


mccumbermike:
Cloned, compiled, and ran against the full stack of eval. No run-time errors, and the eval output looks similar for the clusters, so I think it is fine to merge. I'm taking this as a positive indication that the new evals are indeed easier to support.

Before: ntp_cluster->Draw("e:gedep","efromtruth/e > 0.01")
before

After: ntp_cluster->Draw("e:gedep","efromtruth/e > 0.01")
after

PR #36, opened 2015-09-08
Verbosity Threshold Updates

mccumbermike:
I decided to make this a formal pull request since I introduce a new object into g4main to handle the G4 streams, PHG4UIsession.h. At the moment I'm using this object to quench the GEANT4 initialization messages (those to G4out) for verbosity=0, but it would let us mask other chatter from GEANT4 as needed later if this proves to be an issue. All error messages are always passed.

Other changes are some verbosity if statements and verbosity += 1 style mods.

I haven't produced a global verbosity setting, we can handle that next or later, but this patch equalizes the settings for verbosity=0 to rather quiet running.


blackcathj:
Checked compile and output for multi-electron simulation with the default macro. It looks good to me. Merging.

PR #38, opened 2015-09-10
Proposed Merge from TPC branch

mccumbermike:
This pull request will bring in the underlying changes to HelixHough and some of the specialized modules for TPC reconstruction. The merge of PHG4HoughTransform is residing in PHG4HoughTransformTPC at least until the issue of dropped tracks has been further diagnosed. I may work on that issue tomorrow, but for now most of the commits can be brought together.

The performance of the SVTX alone is unchanged by the proposed merge (open symbols are from master, closed from svtx_tpc_merge):
crosscheck_20150910

EDIT: Alan if you would look at the diff between the two Subsystem recos and comment that would be helpful.


alandion:
OK, looks good to me.  FYI HelixHough is getting some changes soon and we'll need to test those as well

mccumbermike:
Not a problem. Make a pull request, assign it to me, I'll make the
comparison for the silicon option.

--
Michael P. McCumber, PhD
505-709-8161

PR #40, opened 2015-09-14
SvtxTrack storage and interface revision

mccumbermike:
This merge will rework the SvtxTrack object so that it is suitable for additional modularization of the tracking reconstruction. The track object can now store a projection of the track fit at multiple locations along the path. The default state vector (at the point of closest approach) is exposed through svtxtrack->get_??() methods. Some analysis modules will need minor updates to follow the new interface. I also fix some minor inconsistencies in some method definitions and a bug that swapped the x and y PCA values.

In reworking the object's internal storage I have recovered 2 kilobytes per track for silicon-only tracking. So central event storage is down over 1 MB per event. There is also space savings for the TPC tracks, but not as large a change.

The g4hough's i/o objects will still need to be versioned. My plan is to do that in a future pull request. I have an idea to reduce the storage size of the SvtxHit and SvtxCluster by ~10% as well by utilizing the covariance matrix packed array I have implemented here in SvtxTrack. I believe those changes can hide behind the new interface brought in on this request so further disruption cause by interface changes should now be very minimal. We should pull this new interface while the TPC reconstruction can be fast forwarded with these changes as well. 

I also verified the performance in central events for this merge (solid points) against yesterday's new build (open points):
crosscheck

### DIFF
--- a/simulation/g4simulation/g4eval/JetRecoEval.C
+++ b/simulation/g4simulation/g4eval/JetRecoEval.C
@@ -408,7 +408,7 @@ float JetRecoEval::get_energy_contribution(Jet* recojet, Jet* truthjet) {
 	SvtxTrack* track = _trackmap->get(index);
 	PHG4Particle* maxtruthparticle = get_svtx_eval_stack()->get_track_eval()->max_truth_particle_by_nclusters(track);
 	if (maxtruthparticle->get_track_id() == truthparticle->get_track_id()) { 
-	  energy = track->getMomentum();
+	  energy = track->get_p();
 	}
       } else if (source == Jet::CEMC_TOWER) {
 	RawTower* tower = _cemctowers->getTower(index);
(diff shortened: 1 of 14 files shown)


blackcathj:
@mccumbermike , thanks for the update. I like these improvement a lot in general. 

Meanwhile, during cross checking, I found a compile error at 

'''
std::vector<Jet*> TrackJetInput::get_input(PHCompositeNode *topNode) {
  
    Jet *jet = new JetV1();
    jet->set_px(track->get3Momentum(0));
    jet->set_py(track->get3Momentum(1));
    jet->set_pz(track->get3Momentum(2));
    jet->set_e(track->getMomentum());
'''

which appears not updated with the track object.

mccumbermike:
Yes, this is my oversight. I've patched that on the branch. I also
discovered an issue with the new field (it might be reversed from the last
version) so I've reversed the sense of helicity to charge. The track eval
also had a bug preventing the look up of the truth exit point, that is
patched as well.

Mike

--
Michael P. McCumber, PhD
505-709-8161

blackcathj:
Tested couple events. Looks right. Merging.

Issue #44, opened 2015-09-16
segfault in jet evaluator for hijing hits files


pinkenburg:
using the standard Fun4All_G4_sPHENIX.C macro with readhits=true and the input file
/gpfs02/phenix/prod/sPHENIX/preCDR/pro.1-beta.3/sHijing/spacal2d/G4Hits_sPHENIX_sHijing-0-4.4fm_0000.root
It crashes in
JetTruthEval.C:63
with the truthjet being a null pointer.

mccumbermike:
I found a few hours ago a limitation in the tracking and calorimeter evaluator when there is more than one vertex in phg4truthinfo. This was based on my misunderstanding of the PHG4TruthInfo storage where I had to guess at what was stored on what map and when---I think I understand that now. The jet eval has a similar problem, but there it seems I will need to store the indexes of secondary and primary truth particles separately since the tracing should be done on different maps.

mccumbermike:
I'm pretty

mccumbermike:
Sorry finger slipped. The new pull request should resolve this issue. Let me know if it does not.

mccumbermike:
There is still a problem with the latest new build. I'm looking into it this morning.

mccumbermike:
I still need to work on the TPC branch eval to bring that under eval support---I'll be doing this very soon now---but the default detector config is running well for the jet eval to the extent that I can test as of commit aaf394a. Closing issue.

PR #45, opened 2015-09-16
Eval modifications + new random numbers for PHPythia8

mccumbermike:
This merge will bring in the modifications to PHPythia8 to get it read for use (random number seeding mostly). In testing PHPythia8, I discovered all the ways the copied primaries & primary vertexes are not equivalent to the input primary storage and modified the eval to route around this issues. I also solved the problem in issue #44 where the truth eval expected a truth particle to always be in a truth jet, but this won't always be the case given eta acceptance and jet definitions.

Jin I'm assigning this to you. I think having a second runtime check would be helpful. I did see some reco jets being traced back to similar kinematic truth jets in the eval ntuple---that's an encouraging sign.

### DIFF
--- a/generators/PHPythia8/Makefile.am
+++ b/generators/PHPythia8/Makefile.am
@@ -23,6 +23,8 @@ libPHPythia8_la_LDFLAGS = \
   `root-config --libs`
 
 libPHPythia8_la_LIBADD = \
+  -lgsl \
+  -lgslcblas \
   -lphool \
   -lSubsysReco \
   -lfun4all \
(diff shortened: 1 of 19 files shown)


blackcathj:
Crosscheck by testing with some calorimeter events. The output seems reasonable. I will merge it in the master then. 

Meanwhile, it is probably a good time to add
```
coresoftware/generators/PHPythia8
```
to the nightly build then. @pinkenburg

pinkenburg:
I added it to the build and it succeeded just now.

On 9/16/2015 11:30 PM, Jin Huang wrote:
>
> Crosscheck by testing with some calorimeter events. The output seems 
> reasonable. I will merge it in the master then.
>
> Meanwhile, it is probably a good time to add
>
> |coresoftware/generators/PHPythia8 |
>
> to the nightly build then. @pinkenburg <https://github.com/pinkenburg>
>
> —
> <https://github.com/sPHENIX-Collaboration/coresoftware/pull/45#issuecomment-140959501>.
>

-- 

Upton, NY 11973-5000

PR #54, opened 2015-09-19
Fully configurable outer hcal

pinkenburg:
The code started a copy of the inner hcal with the added cutout needed for the magnet. I preserved the field map (field only in steel plates) and the handling for the light collection/generation. Pretty much all the old code was retired.


pinkenburg:
I merge this in - I hope to get the next pass of sims started tonight. For the hcals the known issues in the geometry are fixed with this.

pinkenburg:
I'll kick off a new build and then will adjust the outer hcal reference macro

Issue #52, opened 2015-09-17
infinite loop in calo evaluator for inner hcal in hijing


pinkenburg:
I started to run over /gpfs02/phenix/prod/sPHENIX/preCDR/pro.1-beta.3/sHijing/spacal2d/G4Hits_sPHENIX_sHijing-0-4.4fm_0200.root using the standard Fun4All_G4_sPHENIX.C macro with readhits=true and the jeteval disabled. It runs now for 10 hours, attaching with the debugger indicates it is still at the first event. It seems to be in an infinite loop in CaloTruthEval.C line 171:
  for (PHG4HitContainer::ConstIterator g4iter = _g4hits->getHits().first;
       g4iter != _g4hits->getHits().second;
       ++g4iter) {

    PHG4Hit* g4hit = g4iter->second;
    PHG4Particle* candidate = get_primary_particle(g4hit);
    if (candidate->get_track_id() != primary->get_track_id()) continue;
    truth_hits.insert(g4hit);
  }

the code doesn't seem to find a match for the candidate and keeps "continuing"

mccumbermike:
I've been poking at this issue for the past hour or so. I don't think it
isn't truly infinite, but semi-infinite for sure. It doesn't spend much
time here until you throw hijing at it. I'm sure this has to do with
redefining the sense of primary to include both primaries on the PrimaryMap
and the copies on the Map. Which was needed to get the vertex retrieval
working.

More later.

Mike

--
Michael P. McCumber, PhD
Los Alamos National Laboratory
505-709-8161

pinkenburg:
yes it is not infinite. I set a breakpoint after this loop and the code eventually made it there

pinkenburg:
I timed the first event in this file and it needs 15 minutes (without jet eval). That's comparable with simulating a hijing event. Therefore we should consider to run this either as part of the hijing simulation production or better in a separate centralized pass.

mccumbermike:
I'll have some patches soon. I had to develop a correctness test procedure
so I can follow when the eval is making incorrect decisions under load.
Throw in a lot of noise towers and things get difficult to debug by eye,
but dealing with those is also essential.

Some of the speed issues in central events came from tracing primary
"copies" stored in Map separately from primaries in the Primary Map. I
think by extending the trace all the way to the PrimaryMap (which is
smaller and has faster look up) I can improve both the caching and do my
searches on a smaller map. Apologies for the inconvenience I hadn't
[...]

PR #57, opened 2015-09-19
Calorimeter evaluator modifications

mccumbermike:
This merge will bring in a couple of things. Most importantly it improves the correctness of the energy contribution calculations. I found these by embedding 40 GeV electrons in central events and watching that the evaluator was not getting confused by the additional trace possibilities.

It also mitigates the speed issues by improving the merging of sets. Full eval of the CEMC in a central event should run 10 minutes or so. For single particle traces within central events these commits resulted in a ~x10 improvement. I've added methods to reduce the forward tracing to particular embed flags and reverse tracing only above some energy threshold. This will buy us some time in many circumstances. 

I need to do more profiling to guide further optimization. One possibility is I might have to swap to unordered_sets, that is a pain to do with tr1, but is still possible if we are still holding off on c++11. I'll let the profiler tell me on Monday at the moment the bottlenecks have moved away from the insert cost / merge cost.

Given the logic repair on the energy contributions and some strides on the optimization, I'm making this pull request available now.

Partially addresses issue #52

### DIFF
--- a/simulation/g4simulation/g4eval/CaloEvaluator.C
+++ b/simulation/g4simulation/g4eval/CaloEvaluator.C
@@ -33,6 +33,8 @@ CaloEvaluator::CaloEvaluator(const string &name, const string &caloname, const s
   : SubsysReco(name),
     _caloname(caloname),
     _ievent(0),
+    _truth_trace_embed_flag(0), // truth => reco trace all primaries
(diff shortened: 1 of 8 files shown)


blackcathj:
The update looks great to me. Tried compile and run on single electron simulation, the results look fine. Merging.

PR #59, opened 2015-09-21
Additional CaloEvalStack Optimization

mccumbermike:
I've leaned a bit more on the caching for an additional minor speed improvement. I also provide an energy cutoff for the truth eval tracing. I suggest setting both truth and reco tracing to >2.0 GeV/c for the calorimeter macros. That should run at about half time of a full event.

Most of the time now is spent tracing the between g4hits and g4particles with lots of calls going to the truthinfocontainer. Short of storing the primary truth id on the g4hit id itself (which would explode memory) it is pretty tough to go faster than this now.


blackcathj:
Thanks for the update. 
Have run simple test compile and single particle analysis. 
Merging....

PR #65, opened 2015-09-25
Allow Geant4 Embedding into existing DSTs (e.g. Hijing)

blackcathj:
It turns out some efforts are required to allow Geant4 events to get embedded to an existing DST stream. This procedure is used to reuse the Hijing production DST and embed more single particles into it via a new Geant4 cycle. 

Example Fun4All_sPHENIX.C macro to drive the embedding production is proposed to a fork of the macro repository:
https://github.com/blackcathj/macros/blob/SinglePart_master_embed_prod/macros/g4simulations/Fun4All_G4_sPHENIX.C

Major changes in coresoftware to allow G4->DST embedding are discussed in this pull request. It will be great if @pinkenburg and @mccumbermike can cross check these updates.  And the proposed changes are open for discussion.

## Add InitRun into PHG4Reco

The most significant change would be in initialization order for PHG4Reco. Previously both InitRun and Init part of the G4 subsystems are called in the Init() function of PHG4Reco (strange to me). That's out-of-order for Fun4All. And it would lead to the G4 subsystem init the node tree prior to loading of the input manager (e.g. DST manager when loading another DST stream). In that case, the DST node from the input manager would get dropped. Also, calling InitRun for G4 subsystem in PHG4Reco::Init() would prevent us from using run number/timestamp based database access in the future too. 

Therefore, I have split PHG4Reco::Init() into two parts of Init() and InitRun(). Only call G4 subsystem InitRun in the later part. 

Most G4 subsystem already use InitRun() for initialization. Two HCal subsystems are not though. Therefore, they were updated to use InitRun().

## Treatment for embedded primary particle. 

When embedding particle into a DST stream containing an existing TruthContainer, TruthContainer would append the new primary particle to a shorter primary particle list and its G4 tracks to the longer particle list. Since the two list sizes are different and the primary particle ID is controlled by TruthContainer, the primary particle's IDs in PHG4TruthInfoContainer::hitmap and in PHG4TruthInfoContainer::primary_particle_map would end up different (which PHG4TruthInfoContainer allows). 

The truth subsystems reco part were updated to support so too, sort out the mapping correctly. 

Now, the truth tracing route is consistent with non-embedding analysis, i.e.:
1. Start with PHG4TruthInfoContainer::hitmap via interface functions
2. locate a PHG4Particle in it 
3. Ask for its primary from with primary_id= PHG4Particle::get_primary_id() 
4. Ask for the primary particle with PHG4TruthInfoContainer::primary_particle_map[primary_id] via interface functions

## Evaluators

@mccumbermike , I think it deserve another round of check to make sure the evaluator is conform to the above standard. Especially that, NO implicit assumption that key->particle would be the same for PHG4TruthInfoContainer::primary_particle_map and PHG4TruthInfoContainer::hitmap, since they are not for newly embedded particles. The right way to go should be via the function CaloTruthEval::get_primary_particle(), which is already used in many part of the evaluator. 

Few minor changes were introduced : 
* CaloTruthEval: pointer protection in CaloTruthEval, in case the primary mapping not working. 
* PHG4DSTReader was updated to conform with the standard on primary particle list. 

## Reuse embedded parimary vertex in PHG4SimpleEventGenerator

Now PHG4SimpleEventGenerator would search for the primary vertex available in both InEvent (e.g. for embed into event generator) and primary vertexes in TruthContainer (e.g. for DST embedding)

### DIFF
--- a/simulation/g4simulation/g4main/PHG4TruthEventAction.cc
+++ b/simulation/g4simulation/g4main/PHG4TruthEventAction.cc
@@ -207,7 +207,8 @@ void PHG4TruthEventAction::EndOfEventAction(const G4Event* evt)
       {
 	if (userdata->get_embed())
 	  {
-            truthInfoList_->AddEmbededTrkId(part->GetTrackID()+ trackidoffset);
+//      truthInfoList_->AddEmbededTrkId(part->GetTrackID()+ trackidoffset); // use G4 particle list ID for the embedded list
+      truthInfoList_->AddEmbededTrkId(part->GetTrackID()+ parimarytrackidoffset); // use primary ID for the embedded list
 	  }
       }
       part = part->GetNext();
--- a/simulation/g4simulation/g4main/PHG4TruthEventAction.cc
+++ b/simulation/g4simulation/g4main/PHG4TruthEventAction.cc
@@ -25,6 +25,7 @@ using namespace std;
 PHG4TruthEventAction::PHG4TruthEventAction( void ):
   truthInfoList_( 0 ),
   trackidoffset(0),
+  parimarytrackidoffset(0),
   vertexid_(0)
 {}
 
@@ -167,6 +168,10 @@ void PHG4TruthEventAction::EndOfEventAction(const G4Event* evt)
           if (particle->get_parent_id() <= 0)
             {
               primaryid = particle->get_track_id(); // this particle is the primary truth
+              assert (primaryid > trackidoffset);
+              primaryid -= trackidoffset; // recovery the Geant4 track ID = inEvent track ID
+              primaryid += parimarytrackidoffset; // ID for the primary track in truth container
+
               break;
             }
 
--- a/simulation/g4simulation/g4detectors/PHG4InnerHcalSubsystem.cc
+++ b/simulation/g4simulation/g4detectors/PHG4InnerHcalSubsystem.cc
@@ -35,7 +35,7 @@ PHG4InnerHcalSubsystem::PHG4InnerHcalSubsystem( const std::string &name, const i
 }
 
 //_______________________________________________________________________
-int PHG4InnerHcalSubsystem::Init( PHCompositeNode* topNode )
+int PHG4InnerHcalSubsystem::InitRun( PHCompositeNode* topNode )
 {
   PHNodeIterator iter( topNode );
   PHCompositeNode *dstNode = dynamic_cast<PHCompositeNode*>(iter.findFirst("PHCompositeNode", "DST" ));
--- a/simulation/g4simulation/g4detectors/PHG4InnerHcalSubsystem.h
+++ b/simulation/g4simulation/g4detectors/PHG4InnerHcalSubsystem.h
@@ -29,7 +29,7 @@ class PHG4InnerHcalSubsystem: public PHG4Subsystem
   reates the stepping action and place it on the node tree, under "ACTIONS" node
   creates relevant hit nodes that will be populated by the stepping action and stored in the output DST
   */
-  int Init(PHCompositeNode *);
+  int InitRun(PHCompositeNode *);
 
   //! event processing
   /*!
--- a/simulation/g4simulation/g4detectors/PHG4OuterHcalSubsystem.cc
+++ b/simulation/g4simulation/g4detectors/PHG4OuterHcalSubsystem.cc
@@ -33,7 +33,7 @@ PHG4OuterHcalSubsystem::PHG4OuterHcalSubsystem( const std::string &name, const i
 }
 
 //_______________________________________________________________________
-int PHG4OuterHcalSubsystem::Init( PHCompositeNode* topNode )
+int PHG4OuterHcalSubsystem::InitRun( PHCompositeNode* topNode )
 {
   PHNodeIterator iter( topNode );
   PHCompositeNode *dstNode = dynamic_cast<PHCompositeNode*>(iter.findFirst("PHCompositeNode", "DST" ));
--- a/simulation/g4simulation/g4detectors/PHG4OuterHcalSubsystem.h
+++ b/simulation/g4simulation/g4detectors/PHG4OuterHcalSubsystem.h
@@ -29,7 +29,7 @@ class PHG4OuterHcalSubsystem: public PHG4Subsystem
   reates the stepping action and place it on the node tree, under "ACTIONS" node
   creates relevant hit nodes that will be populated by the stepping action and stored in the output DST
   */
-  int Init(PHCompositeNode *);
+  int InitRun(PHCompositeNode *);
 
   //! event processing
   /*!
--- a/simulation/g4simulation/g4eval/CaloTruthEval.C
+++ b/simulation/g4simulation/g4eval/CaloTruthEval.C
@@ -12,6 +12,7 @@
 #include <set>
 #include <map>
 #include <float.h>
+#include <iostream>
 
 using namespace std;
 
@@ -136,7 +137,10 @@ std::set<PHG4Hit*> CaloTruthEval::get_shower_from_primary(PHG4Particle* primary)
   if (!is_primary(primary)) return std::set<PHG4Hit*>();
 
   primary = get_primary_particle(primary);
-  if (!primary) return std::set<PHG4Hit*>();
+  if (!primary){
+      cout <<__PRETTY_FUNCTION__<<" - Error - encountered invalid primary as input. "<<endl;
+      return std::set<PHG4Hit*>();
+  }
 
   if (_do_cache) {
     std::map<PHG4Particle*,std::set<PHG4Hit*> >::iterator iter =
@@ -155,7 +159,11 @@ std::set<PHG4Hit*> CaloTruthEval::get_shower_from_primary(PHG4Particle* primary)
 
     PHG4Hit* g4hit = g4iter->second;
     PHG4Particle* candidate = get_primary_particle(g4hit);
-    if (!candidate) continue;
+    if (!candidate) {
+        cout <<__PRETTY_FUNCTION__<<" - Error - can not find the primary particle for g4hit: "<<endl;
+        g4hit->identify();
+        continue;
+    }
     if (candidate->get_track_id() != primary->get_track_id()) continue;
     truth_hits.insert(g4hit);
   }
--- a/simulation/g4simulation/g4eval/CaloTruthEval.C
+++ b/simulation/g4simulation/g4eval/CaloTruthEval.C
@@ -136,7 +136,8 @@ std::set<PHG4Hit*> CaloTruthEval::get_shower_from_primary(PHG4Particle* primary)
   if (!is_primary(primary)) return std::set<PHG4Hit*>();
 
   primary = get_primary_particle(primary);
-  
+  if (!primary) return std::set<PHG4Hit*>();
+
   if (_do_cache) {
     std::map<PHG4Particle*,std::set<PHG4Hit*> >::iterator iter =
       _cache_get_shower_from_primary.find(primary);
@@ -154,6 +155,7 @@ std::set<PHG4Hit*> CaloTruthEval::get_shower_from_primary(PHG4Particle* primary)
 
     PHG4Hit* g4hit = g4iter->second;
     PHG4Particle* candidate = get_primary_particle(g4hit);
+    if (!candidate) continue;
     if (candidate->get_track_id() != primary->get_track_id()) continue;
     truth_hits.insert(g4hit);
   }
--- a/simulation/g4simulation/g4eval/PHG4DSTReader.cc
+++ b/simulation/g4simulation/g4eval/PHG4DSTReader.cc
@@ -531,12 +531,12 @@ PHG4DSTReader::process_event(PHCompositeNode* topNode)
 //                  add_particle(rec, part);
 //                }
 
-              for (particle_iter = truthInfoList->GetPrimaryMap().begin();
-                  particle_iter != truthInfoList->GetPrimaryMap().end();
+              for (particle_iter = truthInfoList->GetMap().begin();
+                  particle_iter != truthInfoList->GetMap().end();
                   particle_iter++)
                 {
-
-                  _particle_set.insert(particle_iter->first);
+                  if (particle_iter->second->get_parent_id()<=0)
+                    _particle_set.insert(particle_iter->first);
 
 //                  PHG4Particle * part = particle_iter->second;
 //
--- a/simulation/g4simulation/g4main/PHG4Reco.cc
+++ b/simulation/g4simulation/g4main/PHG4Reco.cc
@@ -244,150 +244,165 @@ int PHG4Reco::Init( PHCompositeNode* topNode )
       reco->Init( topNode );
     }
 
-  // initialize registered subsystems
-  BOOST_FOREACH(SubsysReco * reco, subsystems_)
-    {
-      reco->InitRun( topNode );
-    }
-
-  // create phenix detector, add subsystems, and register to GEANT
-  if (verbosity > 1) cout << "PHG4Reco::Init - create detector" << endl;
-  detector_ = new PHG4PhenixDetector();
-  detector_->Verbosity(verbosity);
-  detector_->SetWorldSizeX(WorldSize[0]*cm);
-  detector_->SetWorldSizeY(WorldSize[1]*cm);
-  detector_->SetWorldSizeZ(WorldSize[2]*cm);
-  detector_->SetWorldShape(worldshape);
-  detector_->SetWorldMaterial(worldmaterial);
-
-  rc->set_FloatFlag("WorldSizex", WorldSize[0]);
-  rc->set_FloatFlag("WorldSizey", WorldSize[1]);
-  rc->set_FloatFlag("WorldSizez", WorldSize[2]);
-
-  BOOST_FOREACH( PHG4Subsystem * g4sub, subsystems_)
-    {
-      detector_->AddDetector( g4sub->GetDetector() );
-    }
-  runManager_->SetUserInitialization( detector_ );
-
-  setupInputEventNodeReader(topNode);
-  // create main event action, add subsystemts and register to GEANT
-  eventAction_ = new PHG4PhenixEventAction();
-
-  BOOST_FOREACH( PHG4Subsystem * g4sub, subsystems_)
-    {
-      PHG4EventAction *evtact = g4sub->GetEventAction();
-      if (evtact)
-	{
-	  eventAction_->AddAction(evtact);
-	}
-    }
-  runManager_->SetUserAction(eventAction_ );
-
-  // create main stepping action, add subsystems and register to GEANT
-  steppingAction_ = new PHG4PhenixSteppingAction();
-  BOOST_FOREACH( PHG4Subsystem * g4sub, subsystems_)
-    {
-      PHG4SteppingAction *action = g4sub->GetSteppingAction();
-      if (action)
-	{
-	  if (verbosity > 1)
-	    {
-	      cout << "Adding steppingaction for " << g4sub->Name() << endl;
-	    }
-	  steppingAction_->AddAction( g4sub->GetSteppingAction() );
-	}
-    }
-  runManager_->SetUserAction(steppingAction_ );
-
-  // create main tracking action, add subsystems and register to GEANT
-  trackingAction_ = new PHG4PhenixTrackingAction();
-  BOOST_FOREACH( PHG4Subsystem * g4sub, subsystems_)
-    {
-      trackingAction_->AddAction( g4sub->GetTrackingAction() );
-    }
-  runManager_->SetUserAction(trackingAction_ );
-
-  // initialize
-  runManager_->Initialize();
-
-  // add cerenkov and optical photon processes
-  // cout << endl << "Ignore the next message - we implemented this correctly" << endl;
-  G4Cerenkov* theCerenkovProcess = new G4Cerenkov("Cerenkov");
-  // cout << "End of bogus warning message" << endl << endl;
-  // G4Scintillation* theScintillationProcess      = new G4Scintillation("Scintillation");
-
-  /*
-  if (verbosity > 0)
-    {
-      // This segfaults  
-      theCerenkovProcess->DumpPhysicsTable();
-    }
-  */
-  theCerenkovProcess->SetMaxNumPhotonsPerStep(100);
-  theCerenkovProcess->SetMaxBetaChangePerStep(10.0);
-  theCerenkovProcess->SetTrackSecondariesFirst(true);
-
-  // theScintillationProcess->SetScintillationYieldFactor(1.);
-  // theScintillationProcess->SetTrackSecondariesFirst(true);
-
-  // Use Birks Correction in the Scintillation process
-  
-  // G4EmSaturation* emSaturation = G4LossTableManager::Instance()->EmSaturation();
-  // theScintillationProcess->AddSaturation(emSaturation);
-
-  G4ParticleTable* theParticleTable = G4ParticleTable::GetParticleTable();
-  G4ParticleTable::G4PTblDicIterator* _theParticleIterator;
-  _theParticleIterator = theParticleTable->GetIterator();
-  _theParticleIterator->reset();
-  while( (*_theParticleIterator)() )
-    {
-      G4ParticleDefinition* particle = _theParticleIterator->value();
-      G4String particleName = particle->GetParticleName();
-      G4ProcessManager* pmanager = particle->GetProcessManager();
-      if (theCerenkovProcess->IsApplicable(*particle))
-	{
-	  pmanager->AddProcess(theCerenkovProcess);
-	  pmanager->SetProcessOrdering(theCerenkovProcess, idxPostStep);
-	}
-      // if (theScintillationProcess->IsApplicable(*particle))
-      // {
-      //   pmanager->AddProcess(theScintillationProcess);
-      //   pmanager->SetProcessOrderingToLast(theScintillationProcess, idxAtRest);
-      //   pmanager->SetProcessOrderingToLast(theScintillationProcess, idxPostStep);
-      // }
-    }
-  G4ProcessManager* pmanager = G4OpticalPhoton::OpticalPhoton()->GetProcessManager();
-  // G4cout << " AddDiscreteProcess to OpticalPhoton " << G4endl;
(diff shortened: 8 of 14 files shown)


mccumbermike:
Hello Jin,

Thanks for taking this up.

There shouldn't be an implicit assumption that the key would be the same
between maps. I had to treat this for multiple collisions and direct
embedding when reading in a HepMC file.

However, adding pointer protection is pretty dangerous when it should not
be needed (as is the case for noise injections). We really need to know
when the tracing is broken and adding too much fault tolerance will allow
these bugs to persist and not be fixed. Can you explain a bit more why this
might be needed in this case? At the very least the eval should squawk
loudly each time this happens.

Can you provide a pointer to a pair of DST files I can run with the example
macro?

Mike

--
Michael P. McCumber, PhD
505-709-8161

blackcathj:
Hi, @mccumbermike 

Good questions.

> However, adding pointer protection is pretty dangerous when it should not
be needed (as is the case for noise injections). 
> We really need to know
when the tracing is broken and adding too much
> fault tolerance will allow
these bugs to persist and not be fixed. 

I agree. Usually I prefer to add a assert before any external pointer use. At least it crash with indication of line number. I guess in this case screaming error message would be fine.

> Can you explain a bit more why this might be needed in this case?
 
Probably not. I put it in before fixing the primary maps ( Otherwise it die silently with segment fault.). I did not test by putting an error message back. It is probably good to switch to squawk loudly. 

> Can you provide a pointer to a pair of DST files I can run with the example
macro?

This is for embeding new g4 simulation to DST hijing events. Therefore it only need one stream of production hijing DST. One example list is already loaded by default as in Fun4All_sPHENIX.C of this branch:
https://github.com/blackcathj/macros/tree/SinglePart_master_embed

mccumbermike:
Okay, I'll need a little bit more time to be sure I understand the new
TruthInfo internal structure is behaving as I expect in various scenarios.

One potential bug I have uncovered already is that the embedded track id
storage isn't following the new primary offsets. If I run one simple event
generator on top of the other, only the first embed flag is preserved. Can
you confirm that is the case on your side?

Mike

--
Michael P. McCumber, PhD
505-709-8161

blackcathj:
@mccumbermike , I tried embedding one simple event generator on to DST output of another simple event generator. The result looks fine:

Step1: get DST of one simple event generator: The embedded particle list is 

```
---particlemap--------------------------
hit key 1
PHG4Particlev2 name: e-, track id: 1, vtxid: 1, parent id: 0, primary id: 1, pid: 11, px: 17.7619, py: -16.1404, pz: 1.86228, e: 24.0721
....
---primaryparticlemap-------------------
primary_particle_ key 1
PHG4Particlev2 name: e-, track id: 1, vtxid: 1, parent id: 0, primary id: -1, pid: 11, px: 17.7619, py: -16.1404, pz: 1.86228, e: 24.0721
---list of embeded tracks-------------------
embeded track ID 1
```

Step2: Then embed another simple event generator on to the above DST, then I get two embedded tracks:  

```
---particlemap--------------------------
hit key 1
PHG4Particlev2 name: e-, track id: 1, vtxid: 1, parent id: 0, primary id: 1, pid: 11, px: 17.7619, py: -16.1404, pz: 1.86228, e: 24.0721
....
hit key 36627
PHG4Particlev2 name: e-, track id: 36627, vtxid: 34055, parent id: 0, primary id: 2, pid: 11, px: -8.77258, py: 19.4344, pz: 11.0156, e: 24
....
---primaryparticlemap-------------------
primary_particle_ key 1
PHG4Particlev2 name: e-, track id: 1, vtxid: 1, parent id: 0, primary id: -1, pid: 11, px: 17.7619, py: -16.1404, pz: 1.86228, e: 24.0721
primary_particle_ key 2
PHG4Particlev2 name: e-, track id: 2, vtxid: 1, parent id: 0, primary id: -1, pid: 11, px: -8.77258, py: 19.4344, pz: 11.0156, e: 24
---list of embeded tracks-------------------
embeded track ID 1
embeded track ID 36627
```

As you can see, the embedded electron of the second cycle is appended to the G4particle list as key36627, primary particle list as key2. It is also updated in the embedded list properly as G4particle key36627.

## Discussion

Therefore, it seems working properly to me. I guess your question may be that why we use G4particle key, rather than primary particle key in the embedded list. 

I guess it is a free choice of ours. And if we want to choose again, I would prefer to choose: save primary particle key in the embedded list. 

This can be switched with a simple change in PHG4TruthEventAction::EndOfEventAction(). Let me know if you want to me update it too.  


## PS: You may find useful: 

The above output comes from PHG4TruthInfo.identify(). There is a simple macro to print it out from any sPHENIX DST: 
```c++
void
DumpTruthContainer(TString DSTFile = "G4sPHENIXCells.root", int Event = 0)
{
  gSystem->Load("libg4hough.so");
  gSystem->Load("libcemc.so");
  gSystem->Load("libg4eval.so");

  TFile *_file0 = TFile::Open(DSTFile);
  T->Scan("DST.G4TruthInfo.identify()",Form("Entry$==%d",Event));
}
```
You might need one commit a81350cfef728ad5bccc67fb15c03d7d9fca5451 , which add the print out the embedded list in G4TruthInfo.identify().

mccumbermike:
Hello Jin,

Yes I'd also prefer storing primary id. Tracing back to the primary map is one more logn hop whereas moving forward is a linear search in a bigger map.

On the broken traces to NULLs, I think I also prefer the assert test to squawking. Let's do that in these few cases where you had problems.

Mike

--
Michael P. McCumber, PhD
505-709-8161

simulation/g4simulation/g4main/PHG4TruthEventAction.cc
```diff
@@ -202,7 +207,8 @@ void PHG4TruthEventAction::EndOfEventAction(const G4Event* evt)
       {
 	if (userdata->get_embed())
 	  {
-            truthInfoList_->AddEmbededTrkId(part->GetTrackID()+ trackidoffset);
+//      truthInfoList_->AddEmbededTrkId(part->GetTrackID()+ trackidoffset); // use G4 particle list ID for the embedded list
+      truthInfoList_->AddEmbededTrkId(part->GetTrackID()+ parimarytrackidoffset); // use primary ID for the embedded list
```
blackcathj: As Mike also agreed, I switched the ID used in the embedded particle list from the G4 particle list to the primary particle list. The option to switch back to the G4 particle list still gets preserved here via a comment line. 

A test of previous example (embed a new electron into DST of an embedded electron) runs well with following content in the truth container:
```
---primaryparticlemap-------------------
primary_particle_ key 1
PHG4Particlev2 name: e-, track id: 1, vtxid: 1, parent id: 0, primary id: -1, pid: 11, px: 17.7619, py: -16.1404, pz: 1.86228, e: 24.0721
primary_particle_ key 2
PHG4Particlev2 name: e-, track id: 2, vtxid: 1, parent id: 0, primary id: -1, pid: 11, px: -20.5186, py: 12.197, pz: 2.49396, e: 24
---primary vertex emap-------------------
vtx id: 1
vtx position: x: 8.18976, y: -4.01256, z: -6.19807, t: 0
---list of embeded tracks-------------------
embeded track ID 1
embeded track ID 2     <<<<<<<<<<<< now the second embedded electron use the primary ID #2. 
```

pinkenburg: Init() vs InitRun() in PHG4Reco subsystems

Fun4All always creates the DST node - it is not overwritten by an input manager (the input manager only creates one if it doesn't exist).

There is a real difference between our usual reco modules and those detector subsystems which run under PHG4Reco. In normal SubsysReco modules which are called by the Fun4AllServer, its Init() method is called as part of the registration with Fun4All. At that time the default node tree (TOP, DST) exists but the run number is not known. The InitRun() is called after the first event is read in. It was originally the place to load calibrations but when reading DSTs this is the time where the full node tree recovered from the DST is available. So if your module checks for object which come from DST, that code has to go into the InitRun method. The other important case is when a module creates a node which exists on DST. During reading from the DST the node created in the Init() method gets overwritten.

For the modules running under PHG4Reco the situation is substantially different. All of them need G4 to be initialized, which happens in PHG4Reco::Init() but PHG4Reco is only registered with Fun4All after the detector macros have been executed. This prevents us from calling their Init() method during the registration with PHG4Reco. That's why PHG4Reco loops over the Init() methods of its subsystems. That some modules have an InitRun() is likely some confusion because it's meaningless for detector simulations as they are right now.

But there will be a future need for a more elaborate scheme for embedding. What you really want is to load the same detector setup which was used in the original DST. Right now we rely on the user to execute the same macro, as soon as we have a lot of them this will fail. That configuration could be identified by a run number or better a separate identifier. This one would only be known after the DST was opened and detectors can therefore only be configured after this (that would involve calling an equivalent of the G4Setup.C macro at that time)

blackcathj:
Hi, @pinkenburg 

Thank you for the insight. A few more question and discussion as following:

## Embedding with detector configuration from input DST

The workflow as you described is the right way to go (after pre-CDR rush).  

One prototype how we can achieve this has already been implemented in two detectors: PHG4SpacalSubsystem and PHG4SectorDetector. In both cases, the configuration of the whole detector was described by a macro-configurable and DST&database-stream-able object. 

For example, in the case of nominal Geant4 simulation, the configuration flow for PHG4SpacalSubsystem is 
1. User macro configures PHG4SpacalSubsystem::get_geom(), which is a macro-editable object of PHG4CylinderGeom_Spacalv3
2. Then  PHG4SpacalSubsystem::_geom is passed down to both the DST tree (under TOP/RUN/CYLINDERGEOM_CEMC) and the detector construction module (PHG4SpacalDetector)
3. Then  PHG4SpacalDetector construct a SPACAL in Geant4 world based on this object. Identical geometry information was passed down to reconstruction code via TOP/RUN/CYLINDERGEOM_CEMC to form towers in analysis. 

Then we can imaging a proper embed flow would be
1. PHG4SpacalSubsystem read the geometry under TOP/RUN/CYLINDERGEOM_CEMC, if it is available from the embedding DST input (require few additional line of code in PHG4SpacalSubsystem).
2. PHG4SpacalSubsystem pass the geometry down to PHG4SpacalDetector for G4 simulation
Later we could also let PHG4SpacalSubsystem read database to fetch the geometry too, based on the run number/timestamp. 

In this work model, we can conveniently ensure the detector configuration of PHG4SpacalSubsystem is identical to the upstream DST (as well as for analysis). Also it eliminated the multiple property copying/initialization between PHG4SpacalSubsystem and PHG4SpacalDetector . 


## Merge new simulation with the embedding DST stream

Your description on the DST node handling make sense. However, the main reason I tried to split out InitRun() for PHG4Reco subsystems was that, if subsystems make TOP/DST/G4HIT_* nodes in PHG4Reco::Init(), then Fun4AllNoSyncDstInputManager refused to load G4HIT from the embedding DST stream. A bunch of errors popped up at the stage of loading input DST as appended. And the output DST only contain the newly simulated hit, without hit from the input embedding DST. 

Here is a macro branch for illustrating this problem based on the current master build (before split out InitRun):
https://github.com/blackcathj/macros/blob/SinglePart_master_embed_master_demo/macros/g4simulations/Fun4All_G4_sPHENIX.C

Here is the errors messages:
```
Fun4AllServer: Runnumber forced to 0 by RUNNUMBER IntFlag
executing run for input master DefaultSyncManager
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node CYLINDERGEOM_PIPE already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node CYLINDERGEOM_SVTX already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node CYLINDERGEOM_SVTXSUPPORT already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node CYLINDERGEOM_CEMC_ELECTRONICS already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node CYLINDERGEOM_CEMC_SPT already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node CYLINDERGEOM_HCALIN_SPT already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node CYLINDERGEOM_MAGNET already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node CYLINDERGEOM_BH_1 already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node CYLINDERGEOM_BH_FORWARD_PLUS already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node CYLINDERGEOM_BH_FORWARD_NEG already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node CYLINDERGEOM_CEMC already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node CYLINDERGEOM_ABSORBER_CEMC already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node CYLINDERGEOM_HCALIN already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node CYLINDERGEOM_HCALOUT already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node G4HIT_ABSORBER_HCALIN already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node G4HIT_HCALIN already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node G4HIT_ABSORBER_HCALOUT already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node G4HIT_HCALOUT already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node G4HIT_PIPE already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node G4HIT_SVTX already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node G4HIT_SVTXSUPPORT already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node G4HIT_CEMC_ELECTRONICS already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node G4HIT_CEMC already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node G4HIT_ABSORBER_CEMC already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node G4HIT_CEMC_SPT already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node G4HIT_HCALIN_SPT already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node G4HIT_MAGNET already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node G4HIT_BH_1 already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node G4HIT_BH_FORWARD_PLUS already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node G4HIT_BH_FORWARD_NEG already exists
/home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: Node G4TruthInfo already exists
```


## InitRun() for PHG4Reco subsystems

I think there is value to keep InitRun() for PHG4Reco subsystems, and call it at the stage of Fun4All::InitRun():
* Allow G4 hit appended to the input DST stream as described in the second bullet above
* Allow fetch geometry from input DST as described in the first bullet above
* Allow database reading in the future based on run number/timestamp.

pinkenburg:
That was an odd one - not sure why exactly this went wrong. It has been 
working in PHENIX forever.

When reading DST's this check serves to see if some node on the DST 
already exists:

       PHIODataNode<TObject> *newIODataNode =
     dynamic_cast<PHIODataNode<TObject> *> 
(nodeIter.findFirst("PHIODataNode", (*splitvec.rbegin()).c_str()));

The lookup is done by name and if it is found the pointer is returned. 
The objects of the nodes on the node tree inherit from PHObject which 
inherits from TObject so this should return the node if it exists. For 
some reason this results in a NULL pointer and then the object is 
created and put on the node tree which then results in this error 
message. Since we know that a TObject is returned I replaced this by a 
static cast which works.

On 09/29/2015 11:58 AM, Jin Huang wrote:
>
> Hi, @pinkenburg <https://github.com/pinkenburg>
>
> Thank you for the insight. A few more question and discussion as 
> following:
>
>
>     Embedding with detector configuration from input DST
>
> The workflow as you described is the right way to go (after pre-CDR 
> rush).
>
> One prototype how we can achieve this has already been implemented in 
> two detectors: PHG4SpacalSubsystem and PHG4SectorDetector. In both 
> cases, the configuration of the whole detector was described by a 
> macro-configurable and DST&database-stream-able object.
>
> For example, in the case of nominal Geant4 simulation, the 
> configuration flow for PHG4SpacalSubsystem is
> 1. User macro configures PHG4SpacalSubsystem::get_geom(), which is a 
> macro-editable object of PHG4CylinderGeom_Spacalv3
> 2. Then PHG4SpacalSubsystem::_geom is passed down to both the DST tree 
> (under TOP/RUN/CYLINDERGEOM_CEMC) and the detector construction module 
> (PHG4SpacalDetector)
> 3. Then PHG4SpacalDetector construct a SPACAL in Geant4 world based on 
> this object. Identical geometry information was passed down to 
> reconstruction code via TOP/RUN/CYLINDERGEOM_CEMC to form towers in 
> analysis.
>
> Then we can imaging a proper embed flow would be
> 1. PHG4SpacalSubsystem read the geometry under 
> TOP/RUN/CYLINDERGEOM_CEMC, if it is available from the embedding DST 
> input (require few additional line of code in PHG4SpacalSubsystem).
> 2. PHG4SpacalSubsystem pass the geometry down to PHG4SpacalDetector 
> for G4 simulation
> Later we could also let PHG4SpacalSubsystem read database to fetch the 
> geometry too, based on the run number/timestamp.
>
> In this work model, we can conveniently ensure the detector 
> configuration of PHG4SpacalSubsystem is identical to the upstream DST 
> (as well as for analysis). Also it eliminated the multiple property 
> copying/initialization between PHG4SpacalSubsystem and 
> PHG4SpacalDetector .
>
>
>     Merge new simulation with the embedding DST stream
>
> Your description on the DST node handling make sense. However, the 
> main reason I tried to split out InitRun() for PHG4Reco subsystems was 
> that, if subsystems make TOP/DST/G4HIT_* nodes in PHG4Reco::Init(), 
> then Fun4AllNoSyncDstInputManager refused to load G4HIT from the 
> embedding DST stream. A bunch of errors popped up at the stage of 
> loading input DST as appended. And the output DST only contain the 
> newly simulated hit, without hit from the input embedding DST.
>
> Here is a macro branch for illustrating this problem based on the 
> current master build (before split out InitRun):
> https://github.com/blackcathj/macros/blob/SinglePart_master_embed_master_demo/macros/g4simulations/Fun4All_G4_sPHENIX.C
>
> Here is the errors messages:
>
> |Fun4AllServer: Runnumber forced to 0 by RUNNUMBER IntFlag executing 
> run for input master DefaultSyncManager 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node CYLINDERGEOM_PIPE already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node CYLINDERGEOM_SVTX already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node CYLINDERGEOM_SVTXSUPPORT already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node CYLINDERGEOM_CEMC_ELECTRONICS already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node CYLINDERGEOM_CEMC_SPT already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node CYLINDERGEOM_HCALIN_SPT already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node CYLINDERGEOM_MAGNET already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node CYLINDERGEOM_BH_1 already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node CYLINDERGEOM_BH_FORWARD_PLUS already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node CYLINDERGEOM_BH_FORWARD_NEG already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node CYLINDERGEOM_CEMC already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node CYLINDERGEOM_ABSORBER_CEMC already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node CYLINDERGEOM_HCALIN already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node CYLINDERGEOM_HCALOUT already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node G4HIT_ABSORBER_HCALIN already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node G4HIT_HCALIN already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node G4HIT_ABSORBER_HCALOUT already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node G4HIT_HCALOUT already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node G4HIT_PIPE already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node G4HIT_SVTX already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node G4HIT_SVTXSUPPORT already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node G4HIT_CEMC_ELECTRONICS already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node G4HIT_CEMC already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node G4HIT_ABSORBER_CEMC already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node G4HIT_CEMC_SPT already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node G4HIT_HCALIN_SPT already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node G4HIT_MAGNET already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node G4HIT_BH_1 already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node G4HIT_BH_FORWARD_PLUS already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node G4HIT_BH_FORWARD_NEG already exists 
> /home/phnxbld/sPHENIX/new/source/offline/framework/phool/PHCompositeNode.cc:51: 
> Node G4TruthInfo already exists |
>
>
>     InitRun() for PHG4Reco subsystems
>
> I think there is value to keep InitRun() for PHG4Reco subsystems, and 
> call it at the stage of Fun4All::InitRun():
>
>   * Allow G4 hit appended to the input DST stream as described in the
>     second bullet above
>   * Allow fetch geometry from input DST as described in the first
>     bullet above
>   * Allow database reading in the future based on run number/timestamp.
>
> —
> <https://github.com/sPHENIX-Collaboration/coresoftware/pull/65#issuecomment-144102830>.
>


-- 

*********************************************************

Christopher H. Pinkenburg      ; pinkenburg@bnl.gov

Brookhaven National Laboratory ; phone: (631) 344-5692
Physics Department Bldg 510 C  ; fax:   (631) 344-3253
Upton, NY 11973-5000

http://www.phenix.bnl.gov/~pinkenbu/

*********************************************************

blackcathj:
Thanks for all the comments and discussions. As we agreed on this approach, I am merging this pull request into master. 

Next problem is that when running embedding production, I found the output file contain ~250MB/event from the Hijing related truth table. This would be too large for embedding production which require ~100k particles. And it is also not quite useful to trace truth particle for the hijing background. 

Therefore, I plan to skim and dropout non-embedded particle from the truth table, with an optional flag in the PHG4TruthSubsystem. It will be default to be not activated for nominal productions, but set to true for embedding productions. ATTN @mccumbermike , it will also require updates in the evaluator to tolerate Hijing PHG4Hit without parent particle in the truth table. They just become part of the background hit, treated like the noise in the calorimeter tower during the analysis.

Will propose that in the next pull request.

PR #66, opened 2015-09-30
Option to remove non-embedded particle and vertex from the truth table

blackcathj:
As advertised in pull request #65 , one remaining problem to kick off the embedding production is that the output DST file size is large, even after excluding the absorber G4Hits (~250MB/event). 

One solution to this problem is proposed in this pull request: an option (PHG4TruthSubsystem::SetSaveOnlyEmbeded() ) to drop off the truth G4 particle and vertex which is not associated with embedded particle. This option is by-default off, and can be activated via macro during the embedding production. This is driven by the analysis need for embedding production, for which hijing background are noise-like hits and the only embedded particle need to be traced via truth table. Meanwhile, the evaluator was updated in order to handle the missing truth info for the higjing particles. 

An example production macro using this feature is on this branch: 
https://github.com/blackcathj/macros/blob/SinglePart_master_embed_prod/macros/g4simulations/Fun4All_G4_sPHENIX.C 

With this option activated, the output size is reduced by ~x10, about 30MB/event (w/24GeV EM Shower embedded). The leading space consumers are as following, half of which is in hits of EMCal fibers:
```
******************************************************************************
*Tree    :T         : titled by PHOOL                                        *
*Entries :        2 : Total =       150750366 bytes  File  Size =   63266626 *
*        :          : Tree compression factor =   2.25                       *
******************************************************************************
*Br    2 :DST.G4HIT_HCALIN.hitmap : map<unsigned long long,PHG4Hit*>         *
*Entries :        2 : Total  Size=   30116732 bytes  File Size  =   12173537 *
*Baskets :        2 : Basket Size=      32000 bytes  Compression=   2.47     *
*............................................................................*
*Br    6 :DST.G4HIT_HCALOUT.hitmap : map<unsigned long long,PHG4Hit*>        *
*Entries :        2 : Total  Size=   16851408 bytes  File Size  =    7039598 *
*Baskets :        2 : Basket Size=      32000 bytes  Compression=   2.39     *
*............................................................................*
*Br   10 :DST.G4HIT_PIPE.hitmap : map<unsigned long long,PHG4Hit*>           *
*Entries :        2 : Total  Size=    4438069 bytes  File Size  =    2398833 *
*Baskets :        2 : Basket Size=      32000 bytes  Compression=   1.85     *
*............................................................................*
*Br   14 :DST.G4HIT_SVTX.hitmap : map<unsigned long long,PHG4Hit*>           *
*Entries :        2 : Total  Size=    5805603 bytes  File Size  =    3235109 *
*Baskets :        2 : Basket Size=      32000 bytes  Compression=   1.79     *
*............................................................................*
*Br   18 :DST.G4HIT_CEMC.hitmap : map<unsigned long long,PHG4Hit*>           *
*Entries :        2 : Total  Size=   66139416 bytes  File Size  =   30633112 *
*Baskets :        2 : Basket Size=      32000 bytes  Compression=   2.16     *
*............................................................................*
*Br   22 :DST.G4TruthInfo.hitmap : map<int,PHG4Particle*>                    *
*Entries :        2 : Total  Size=   12411081 bytes  File Size  =    3393135 *
*Baskets :        2 : Basket Size=      32000 bytes  Compression=   2.41     *
*............................................................................*
*Br   23 :DST.G4TruthInfo.vtxmap : map<int,PHG4VtxPoint*>                    *
*Entries :        2 : Total  Size=    8222009 bytes  File Size  =    2645347 *
*Baskets :        2 : Basket Size=      32000 bytes  Compression=   2.04     *
*............................................................................*
*Br   24 :DST.G4TruthInfo.primary_particle_map : map<int,PHG4Particle*>      *
*Entries :        2 : Total  Size=    2896191 bytes  File Size  =     625294 *
*Baskets :        2 : Basket Size=      32000 bytes  Compression=   3.03     *
*Br   29 :DST.PHHepMCGenEvent._theEvt : HepMC::GenEvent*                     *
*Entries :        2 : Total  Size=    3836883 bytes  File Size  =    1113638 *
*Baskets :        2 : Basket Size=      32000 bytes  Compression=   3.44     *
*............................................................................*
```

### DIFF
--- a/simulation/g4simulation/g4eval/SvtxEvaluator.C
+++ b/simulation/g4simulation/g4eval/SvtxEvaluator.C
@@ -31,6 +31,7 @@
 #include <iostream>
 #include <set>
 #include <cmath>
+#include <cassert>
 
 using namespace std;
 
@@ -595,7 +596,10 @@ void SvtxEvaluator::fillOutputNtuples(PHCompositeNode *topNode) {
     SvtxVertexMap* vertexmap = findNode::getClass<SvtxVertexMap>(topNode,"SvtxVertexMap");
     PHG4TruthInfoContainer* truthinfo = findNode::getClass<PHG4TruthInfoContainer>(topNode,"G4TruthInfo");
     if (vertexmap && truthinfo) {
-      PHG4VtxPoint* point = truthinfo->GetPrimaryVtx(truthinfo->GetVtxRange().first->second->get_id());
+//        PHG4VtxPoint* point = truthinfo->GetPrimaryVtx(truthinfo->GetVtxRange().first->second->get_id());
+        // to get the primary vertex? Then we probably should use the interface function in the truth container
+        PHG4VtxPoint* point =  truthinfo->GetPrimaryVtx(truthinfo->GetPrimaryVertexIndex());
+        assert(point);
       SvtxVertex* vertex = vertexeval->best_vertex_from(point);
     
       float gvx        = point->get_x();
@@ -653,15 +657,19 @@ void SvtxEvaluator::fillOutputNtuples(PHCompositeNode *topNode) {
       float glayer    = g4hit->get_layer();
   
       float gtrackID  = g4hit->get_trkid();
-      float gflavor   = g4particle->get_pid();
-      float gpx       = g4particle->get_px();
-      float gpy       = g4particle->get_py();
-      float gpz       = g4particle->get_pz();
 
-      PHG4VtxPoint* vtx = trutheval->get_vertex(g4particle);	
-      float gvx       = vtx->get_x();
-      float gvy       = vtx->get_y();
-      float gvz       = vtx->get_z();
+
+      float gflavor   = NAN;
+      float gpx       = NAN;
+      float gpy       = NAN;
+      float gpz       = NAN;
+
+      float gvx       = NAN;
+      float gvy       = NAN;
+      float gvz       = NAN;
+
+      float gembed  = NAN;
+      float gprimary = NAN;
 
       float gfpx      = NULL;
       float gfpy      = NULL;
@@ -669,6 +677,24 @@ void SvtxEvaluator::fillOutputNtuples(PHCompositeNode *topNode) {
       float gfx       = NULL;
       float gfy       = NULL;
       float gfz       = NULL;
+
+      if (g4particle)
+        {
+
+       gflavor   = g4particle->get_pid();
+       gpx       = g4particle->get_px();
+       gpy       = g4particle->get_py();
+       gpz       = g4particle->get_pz();
+
+      PHG4VtxPoint* vtx = trutheval->get_vertex(g4particle);	
+      assert(vtx); // vtx should always be available. Delete this is it is suppose not
+
+      if (vtx)
+        {
+       gvx       = vtx->get_x();
+       gvy       = vtx->get_y();
+       gvz       = vtx->get_z();
+        }
     
       PHG4Hit* outerhit = trutheval->get_outermost_truth_hit(g4particle);	
       if (outerhit) {
@@ -679,8 +705,9 @@ void SvtxEvaluator::fillOutputNtuples(PHCompositeNode *topNode) {
 	gfy       = outerhit->get_y(1);
 	gfz       = outerhit->get_z(1);
       }
-      float gembed    = trutheval->get_embed(g4particle);
-      float gprimary  = trutheval->is_primary(g4particle);
+       gembed    = trutheval->get_embed(g4particle);
+       gprimary  = trutheval->is_primary(g4particle);
+        } //       if (g4particle)
 
       std::set<SvtxCluster*> clusters = clustereval->all_clusters_from(g4hit);  
       float nclusters = clusters.size();
@@ -711,7 +738,10 @@ void SvtxEvaluator::fillOutputNtuples(PHCompositeNode *topNode) {
 	size       = cluster->size_hits();
 	phisize    = cluster->get_phi_size();
 	zsize      = cluster->get_z_size();
+  if (g4particle)
+    {
 	efromtruth = clustereval->get_energy_contribution(cluster,g4particle);
+    }
       }
 
       float g4hit_data[35] = {_ievent,
@@ -815,6 +845,9 @@ void SvtxEvaluator::fillOutputNtuples(PHCompositeNode *topNode) {
 	  gx       = g4hit->get_x(0);
 	  gy       = g4hit->get_y(0);
 	  gz       = g4hit->get_z(0);
+
+	  if (g4particle){
+
 	  gtrackID = g4particle->get_track_id();
 	  gflavor  = g4particle->get_pid();
 	  gpx      = g4particle->get_px();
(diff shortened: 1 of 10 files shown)


simulation/g4simulation/g4eval/SvtxEvaluator.C
```diff
@@ -595,7 +596,10 @@ void SvtxEvaluator::fillOutputNtuples(PHCompositeNode *topNode) {
     SvtxVertexMap* vertexmap = findNode::getClass<SvtxVertexMap>(topNode,"SvtxVertexMap");
     PHG4TruthInfoContainer* truthinfo = findNode::getClass<PHG4TruthInfoContainer>(topNode,"G4TruthInfo");
     if (vertexmap && truthinfo) {
-      PHG4VtxPoint* point = truthinfo->GetPrimaryVtx(truthinfo->GetVtxRange().first->second->get_id());
+//        PHG4VtxPoint* point = truthinfo->GetPrimaryVtx(truthinfo->GetVtxRange().first->second->get_id());
+        // to get the primary vertex? Then we probably should use the interface function in the truth container
+        PHG4VtxPoint* point =  truthinfo->GetPrimaryVtx(truthinfo->GetPrimaryVertexIndex());
```
blackcathj: @mccumbermike  What is the best way to get the primary vertex? We probably should use the interface function in the truth container, truthinfo->GetPrimaryVertexIndex(). The original line would crash after the hijing G4 vertex get removed.

blackcathj:
Merging, in order to start the embedding production. Example macro is here:
https://github.com/blackcathj/macros/blob/SinglePart_master_embed_prod/macros/g4simulations/Fun4All_G4_sPHENIX.C

mccumbermike:
Hello Jin,

It's difficult to follow the sequence of changes from QM. But it seems like this one will change the evaluation results even for embedded particles. For example a cluster that gets a minority of its energy from the embedded particle can trace to a particle in the underlying event. Once that particle is removed the code will have only the embedded particles to choose from leaving the association will change.

Chris had suggested at the last meeting removing particles that only leave absorber hits with the idea we should reclaim a factor twenty in space savings without impacting the evaluation. Was this tried first? Would that degree of savings be acceptable?

On the evaluation itself, I'm considering a strict mode that requires the guarantees  of the truth table to be kept and a relaxed mode that gracefully handles these broken promises.

Mike

--
Michael P. McCumber, PhD
505-709-8161

blackcathj:
Hi, @mccumbermike 

Thank you for your comments during the busy week of QM. That is an important point. I was planning another module which handle truth reduction, which can take a list of G4 DST node name to keep the associated truth. 

I gave a quick test by choosing to save G4track that is related to hits in the active volume of the four subsystem, then the disk size of the truth table would be ~10x of the one shown above and become the dominant factor for the overall output size beside EMCal. The result overall DST size is about twice of the above proposal too. 

I think we can handle a factor of two larger DST output. Therefore, both choices are viable (save only embedded truth or save embedded+hijing active truth). Meanwhile, from analysis point of view of the coming embedding production, hijing truth tracing is not very useful. It is true the max truth association will be altered as hijing particle being removed. However from the energy contributions, we can figure out which cluster/tower/jet is mainly contributed from the emended particle. That is all I needed for calorimeter analysis without knowing the hijing associations. 

Therefore, I prefer not to not to double the DST size by x2 by saving hijing truth (order of 10 TB total). 

By fully strip off the hijing truth, I could be missing need for future embedding. Therefore, I will check in the truth reduction module soon. So if people want embedding study with hijing truth in DST can use it.

blackcathj:
BTW @mccumbermike , I forgot to mention another point that, after hijing truth being removed from the embedding DST, they are fully recoverable in the analysis stage via merging the original hijing DST's truth table: we can read-in hijing DST via another TOPNODE, then use a simple module to merge hijing truth into the main truthcontainer (w/o overwritting the key). Since the same 1000 Hijing events are reused many times over the embedding production, this would also make it efficient in term of truth storage.

mccumbermike:
I'm getting caught up and I thought I should reply that this merge ended up doing exactly the thing I asked us not to do by slapping arbitrary NULL protections into the code for specific out-of-spec usage scenarios and blinding us to severe production bugs that will creep in later on---don't say it won't happen because it already has. 

I'll go through and clean this up now with a new request to enable a non-default mode for less reliance on the production code actually producing a proper set of traceable output. While I don't expect the minor gap in coverage to be a real problem, the full extent of these changes weren't clear to me prior to the merge (only 24 hours to respond were given and I spent much of QM without an IP address). What is clear to me is that we just went too fast with this merge and in this instance ignored our prior commitments to code reviews by library maintainers---if I hadn't looked in on this after the fact, I could have completely missed the extent to which these important correctness checks were removed.

blackcathj:
> this merge ended up doing exactly the thing 
> I asked us not to do by slapping arbitrary NULL protections into the code for 
> specific out-of-spec usage scenarios and blinding us to severe production bugs 
> that will creep in later on---don't say it won't happen because it already has.

For the purpose of consistency check, should not we use a purposely built module to carry out the cheeks systematically? One existing checker for truth container is PHG4ConsistencyCheck. We could enforce running such modules in the test production or in another round check condor job post-production, in order to systematically check all the link throughout the containers. It sounds better than using evaluator to spot check a subset of it depending on the usage. 

Meanwhile, I do agree it is useful to have the evaluator to have a default strict-check mode to ensure the output is fit to the quality insurance. A tolerate-mode flag as you mentioned before to allow the evaluator run in out-of-spec usage scenarios will be useful in case of specific studies. 


> (only 24 hours to respond were given and I spent much of QM without an IP address)
> What is clear to me is that we just went too fast with this merge 
> and in this instance ignored our prior commitments to code reviews by library maintainers

That is my fault. I was in a hurry to get the embedding production started for pre-CDR plots. It turns out the production still take longer with other issues...

mccumbermike:
A separate consistency check would be one way to go. I'm not in favor of
that solution because as our modes of operation multiply, someone would
have to maintain those modules and ensure that they are run regularly on
each way we can feed in data to the simulation. I don't think that is
maintainable by our small force and that code base if created would quickly
degrade into obsolescence. The fact that the evaluator stack is useful will
ensure it is run widely and often---a consistency check isn't immediately
useful and is unlikely be run widely or often.

I'll have a new eval library for you to test shortly with the relaxed
evaluation mode that should run right over inconsistencies or incomplete
records. I'm not a fan of making that the default mode for exactly the
reasons stated above, but it should be perfectly fine to enable this mode
during development or for specific use cases where we know we will be
breaking the evaluation paths and deep inspection of the simulation is not
possible.

We will need to know when the code breaks---that sounds obvious as I type
it out, but I'm not sure everyone agrees with this notion in practice when
other priorities are intruding. Anyway, the new relaxed mode should offer
some defensive depth that will allow a bug report to be generated, a toggle
thrown, and development to move on while a solution is generated in
parallel.

An "inspection mode" where we are always checking will find those things
much faster and much more completely than a "testing mode" where we only
look for specific issues in specific cases we think we should test in
advance. It will occasionally produce bug reports for the evaluation and a
diversion for me from time-to-time, but I'm also not convinced that is the
worse thing---those bugs would be present anyway, we would just stay
ignorant of them.


--
Michael P. McCumber, PhD
505-709-8161

PR #69, opened 2015-10-06
Problem with projection of track to calorimeter

blackcathj:
This is for discussion rather than merging immediately. 

I have quite some problem on projecting the track to the calorimeter using the default tool: PHG4HoughTransform::projectToRadius. First of all, it use a mix of track state and vertex state to start the projection. Then I couldn't figure out how it use the charge sign to determine the bending direction of the track. Most importantly, it do not seem to work at the current stage. 

Here is projection of one 8 GeV track to 10, 20, ... 80 cm radius. Note the coordinate jump sign around R=20 cm:
```
void EMCalAna::eval_trk(PHG4Particle*, EMCalTrk&, PHCompositeNode*) - info - handling track track p= (3.91465, 6.79369, 0.218141) GeV/c, size_states = 1
void EMCalAna::eval_trk(PHG4Particle*, EMCalTrk&, PHCompositeNode*) - info - track projection : v=1.2296e-05, -7.08519e-06 p = 3.91465, 6.79369
void EMCalAna::eval_trk(PHG4Particle*, EMCalTrk&, PHCompositeNode*) - info - handling track proj. (5.01748, 8.65014, 0.278212), eta 0.0278176, phi1.04518
void EMCalAna::eval_trk(PHG4Particle*, EMCalTrk&, PHCompositeNode*) - info - handling track proj. (-9.88573, -17.386, 0.556427), eta 0.0278178, phi-2.08781
void EMCalAna::eval_trk(PHG4Particle*, EMCalTrk&, PHCompositeNode*) - info - handling track proj. (-14.7538, -26.1214, 0.834646), eta 0.0278179, phi-2.08494
void EMCalAna::eval_trk(PHG4Particle*, EMCalTrk&, PHCompositeNode*) - info - handling track proj. (-19.5717, -34.8848, 1.11287), eta 0.0278182, phi-2.08208
void EMCalAna::eval_trk(PHG4Particle*, EMCalTrk&, PHCompositeNode*) - info - handling track proj. (-24.3395, -43.6759, 1.39111), eta 0.0278185, phi-2.07921
void EMCalAna::eval_trk(PHG4Particle*, EMCalTrk&, PHCompositeNode*) - info - handling track proj. (-29.057, -52.4947, 1.66935), eta 0.027819, phi-2.07634
void EMCalAna::eval_trk(PHG4Particle*, EMCalTrk&, PHCompositeNode*) - info - handling track proj. (-33.7241, -61.3407, 1.94761), eta 0.0278195, phi-2.07347
void EMCalAna::eval_trk(PHG4Particle*, EMCalTrk&, PHCompositeNode*) - info - handling track proj. (-38.3406, -70.2139, 2.22589), eta 0.02782, phi-2.0706
void EMCalAna::eval_trk(PHG4Particle*, EMCalTrk&, PHCompositeNode*) - info - handling track proj. (-47.4215, -88.0409, 2.7825), eta 0.0278214, phi-2.06487
void EMCalAna::eval_trk(PHG4Particle*, EMCalTrk&, PHCompositeNode*) - info - handling track proj. (1547.89, 872.793, 51.8743), eta 0.0291879, phi0.513422
```

## Quick fix

As a quick fix, this pull request carry one suggested algorithm. It is simple in the sense that it assumes the track come near x,y = 0 cm. Then does a helical projection through a homogeneous Bz field towards a cylinder of given radius.  For calorimeter analysis, where only process of cm is needed, this looks sufficient. But this is not a long term solution. 

Here is one projection check towards EMCal and inner HCal (inner HCal seems have a +0.5 phi bin offset which require additional fix in calorimeter code):

g4hits_sphenix_e-_eta0_8gev-all root_ana root_drawecal_pdst_trackprojection_checks_cut_all_event

### DIFF
--- a/simulation/g4simulation/g4hough/PHG4HoughTransform.C
+++ b/simulation/g4simulation/g4hough/PHG4HoughTransform.C
@@ -45,10 +45,12 @@
 
 // ROOT includes
 #include <TH1D.h>
+#include <TVector3.h>
 
 // standard includes
 #include <cmath>
 #include <iostream>
+#include <cassert>
 
 using findNode::getClass;
 using namespace std;
@@ -99,12 +101,20 @@ static inline double sign(double x)
 
 void PHG4HoughTransform::projectToRadius(const SvtxTrack* track, double B, double radius, vector<double>& intersection)
 {
+  assert(radius>0);
 
-  float phi  = atan2(track->get_y(),track->get_x());
+  float phi  = atan2(track->get_py(),track->get_px());
   float d    = track->get_dca2d();
-  float k    = B / 333.6 / track->get_pt();
+  float k    = - B / 333.6 / track->get_pt() * track->get_charge();
   float z0   = track->get_z();
-  float dzdl = track->get_pz()/track->get_p();
+//  float dzdl = track->get_pz()/track->get_p();
+  float dzdS2D = track->get_pz()/track->get_pt();
+
+  if (k>2./radius) return; //no intersection
+  if (fabs(d)>2)
+    {
+      cout <<__PRETTY_FUNCTION__<<" - Error - "<<"DCA ("<<d<<" cm) too large for assumption of this algorithm "<<endl;
+    }
 
   intersection.clear();intersection.assign(3,0.);
   double& x = intersection[0];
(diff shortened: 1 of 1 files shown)


simulation/g4simulation/g4hough/PHG4HoughTransform.C
```diff
@@ -99,12 +101,20 @@ static inline double sign(double x)
...
 void PHG4HoughTransform::projectToRadius(const SvtxTrack* track, double B, double radius, vector<double>& intersection)
 {
+  assert(radius>0);
 
-  float phi  = atan2(track->get_y(),track->get_x());
+  float phi  = atan2(track->get_py(),track->get_px());
   float d    = track->get_dca2d();
-  float k    = B / 333.6 / track->get_pt();
+  float k    = - B / 333.6 / track->get_pt() * track->get_charge();
   float z0   = track->get_z();
```
blackcathj: This seems to be DCA_z rather than abs. z location of the track? Then the track projection need to be further offset by vertex z in the user code.

mccumbermike: Yes, this is a long standing convention bug on the filling of the track.
All coordinates are supposed to be absolute, but it hasn't shown up so far
because most users are throwing events at (0,0,0). I have it on my todo
list, but hadn't fixed it yet because I needed to do a search to make sure
I can swap in the absolute coordinate without affecting downstream code. I
guess in this instance that swap would have fixed the (part of) this issue.

I will have to take a closer look to grok the full extent of the issues.
Give me a day or two to get caught up and I'll come back with some better
understanding. But I'm not opposed to a fast fix right now to aid the
calorimeter studies.

Aside: We also need to fill an outer state vector to make the projection
length to the calorimeter sorter. And we should pick a reference radius for
the hadronic calorimeters that minimizes the offset due to the tilt angles.

Mike

--
Michael P. McCumber, PhD
505-709-8161

mccumbermike:
Okay, I think I have a solution for you. The old code made an assumption that we want to project from the DCA position and with the changes to the state vector we'd have to recreate the exact same conventions for the perigee coordinate system---so I made a new version that is longer, but explicitly handles the different options and is well commented so that others can follow/support the algorithm. The solution in this pull request also makes assumptions that won't always be the case when using this function call---I've tried to remove as many of these assumptions as is reasonably possible in the new version.

The new code projects from the default state vector (DCA point) to the forward-in-time solution where the vector leaves the cylinder defined by the radius input argument. It will then look for a state vector closer to the solution and redo the projection from there. So if/when outer state vectors are stored we can utilize a smaller projection length which will minimize issues with assuming a uniform field and no energy loss in material. There is also a handle to explicitly project as a line when the field is zeroed.

Eventually we should move to a Kalman based swim for these kinds of tasks, but I'm certain that approach would take me longer to develop than this solution.

Jin, could you pull the code from mccumbermike/coresoftware/calo_proj (these commits not yet part of this merge request) and give it a try? It works for projecting electron/positrons into the calorimeter at both small pt and large pt, but there may be some edge cases I overlooked. If it works for you we can either create a new pull request and merge from there or force this branch to follow that sequence of commits before merging.

FYI: The calc_proj branch also reverts the state vector positions to the global coordinate system.

blackcathj:
This problem is fixed with a permanent solution from @mccumbermike  in pull request #80 . Closing this issue.

PR #70, opened 2015-10-07
Evaluator set_strict(true/false) option

mccumbermike:
@blackcathj Here is the option for adding defensive depth to the evaluation. Can you try this on the gutted embedding macro? You will need to call eval->set_strict(false) in the detector macros. It adds lookup protection pretty much everywhere and specifically in the spots you were modifying previously, but it is possible I let something slip through.

### DIFF
--- a/simulation/g4simulation/g4eval/CaloEvalStack.h
+++ b/simulation/g4simulation/g4eval/CaloEvalStack.h
@@ -24,7 +24,9 @@ public:
   virtual ~CaloEvalStack() {}
 
   void next_event(PHCompositeNode *topNode);
-
+  void do_caching(bool do_cache) {_clustereval.do_caching(do_cache);}
+  void set_strict(bool strict) {_clustereval.set_strict(strict);}
+  
   CaloRawClusterEval* get_rawcluster_eval() {return &_clustereval;}
   CaloRawTowerEval*   get_rawtower_eval() {return _clustereval.get_rawtower_eval();}
   CaloTruthEval*      get_truth_eval() {return _clustereval.get_truth_eval();}
--- a/simulation/g4simulation/g4eval/CaloEvaluator.C
+++ b/simulation/g4simulation/g4eval/CaloEvaluator.C
@@ -34,6 +34,7 @@ CaloEvaluator::CaloEvaluator(const string &name, const string &caloname, const s
     _truth_trace_embed_flags(),
     _truth_e_threshold(0.0), // 0 GeV before reco is traced
     _reco_e_threshold(0.0), // 0 GeV before reco is traced
+    _strict(true),
     _do_gpoint_eval(true),
     _do_gshower_eval(true),
     _do_tower_eval(true),
@@ -167,6 +168,8 @@ void CaloEvaluator::printOutputInfo(PHCompositeNode *topNode) {
   if (verbosity > 2) cout << "CaloEvaluator::printOutputInfo() entered" << endl;
 
   CaloEvalStack caloevalstack(topNode,_caloname); 
+  caloevalstack.set_strict(_strict);
+
   CaloRawClusterEval* clustereval = caloevalstack.get_rawcluster_eval();
   CaloTruthEval*        trutheval = caloevalstack.get_truth_eval();
   
@@ -303,6 +306,8 @@ void CaloEvaluator::fillOutputNtuples(PHCompositeNode *topNode) {
   if (verbosity > 2) cout << "CaloEvaluator::fillOutputNtuples() entered" << endl;
 
   CaloEvalStack caloevalstack(topNode,_caloname); 
+  caloevalstack.set_strict(_strict);
+
   CaloRawClusterEval* clustereval = caloevalstack.get_rawcluster_eval();
   CaloRawTowerEval*     towereval = caloevalstack.get_rawtower_eval();
   CaloTruthEval*        trutheval = caloevalstack.get_truth_eval();
--- a/simulation/g4simulation/g4eval/CaloEvaluator.h
+++ b/simulation/g4simulation/g4eval/CaloEvaluator.h
@@ -37,6 +37,8 @@ class CaloEvaluator : public SubsysReco {
   int process_event(PHCompositeNode *topNode);
   int End(PHCompositeNode *topNode);
 
+  void set_strict(bool b) {_strict = b;}
+  
   // funtions to limit the tracing to only part of the event ---------
   // and speed up the evaluation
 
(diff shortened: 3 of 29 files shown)


blackcathj:
Thank you for the quick follow up on the strict mode. 

Tested with the eID test embedded production, in which the truth G4 particle list for Hijing background was removed to save storage. Both evaluator ntuple in non-strict mode and a user code using tracking evaluator were tested (https://github.com/sPHENIX-Collaboration/analysis/blob/master/EMCal-analysis/EMCalAna/EMCalAna.C). 

The non-strict mode worked well on this embedding output. Please merge this pull request at your convenience.  

Meanwhile, in the process of the test, I noticed a few place in the evaluator need to be updated to make distinction between G4 particle or primary particle, when handling a pointer to PHG4Particle. I tried change the following lines which seems fixed the output:
https://github.com/mccumbermike/coresoftware/compare/strict_eval...blackcathj:TestTrackProjectionToRadius_strict_eval?expand=1

mccumbermike:
Hello Jin,

The evaluation of the tracks is intended to utilize particles leaving the
energy deposits and not trace back all the way to primaries when the two
differ---this was a conscience choice built into the tracking evaluation
design. The calorimeter secondaries deep in a shower are not what one would
try to energy match to a cluster Instead the tossed primary photon energy
is desired, so tracing all the way back to the primary is performed.
However the tracking evaluation has a different set of goals.

For instance a conversion in the beam pipe should trace to the secondary
electron that left the actual energy deposits, and not the tossed photon.
If a K+ leaves a hit in the inner layer and then decays and the secondaries
leave hits in the outer layers, that does not constitute a set of hits left
by a single kaon and does not represent a loss indicating a problem with
the tracking. So I had the evaluation setup to do these tasks differently,
which required a different design than the calorimeters. If I'm reading
your changes correctly you would have the tracking trace further back to
the primaries---please correct me if I am mistaken and you had a different
goal in mind.

We could add a handle to limit the evaluator ntuple generation to the
embedded particles (as we have made available for the calorimeter eval) and
recover space from these NULL entries during the embedding analysis with
the reduced truth storage.

Mike


--
Michael P. McCumber, PhD
505-709-8161

blackcathj:
Hi @mccumbermike 

I totally agree with the design concept on the tracking evaluation as you described. 

But that was not my point. My suggested commits were to address two different issues, which shows up when G4 list and primary list become independent during embedding: 
https://github.com/blackcathj/coresoftware/commits/TestTrackProjectionToRadius_strict_eval 

1) that some time the evaluator code used primary particle, which looks deviated from the design concept you mentioned here. 

One example is here:
simulation/g4simulation/g4eval/SvtxEvaluator.C
```c++
if (_ntp_gtrack) { 
PHG4TruthInfoContainer::Map map = truthinfo->GetPrimaryMap();
    for (PHG4TruthInfoContainer::ConstIterator iter = map.begin(); 
 	   iter != map.end(); 
	   ++iter) {
 	PHG4Particle* g4particle = iter->second;
....
 	SvtxTrack* track = trackeval->best_track_from(g4particle);
}
```
Here g4particle is primary particle. But to my understanding, trackeval->best_track_from() assumed an input parameter of G4 particle. The result would be the returned track is always NULL. 

I proposed a fix for that in a suggested commit fb9829e0260e7c31175f04fc594b2924bc987cea , which allows trackeval->best_track_from() handle both primary and G4 particles. I prefer this, because later I would like to know whether a primary particle get reconstructed reasonably in calorimeter track matching analysis. 

BTW, in a few place in evaluator this line was used:
```c++
PHG4TruthInfoContainer::Map map = truthinfo->GetPrimaryMap();
```
This is a complete copy of a container of ~10k primary paricle at each call. 
I assumed you meant to use:
```c++
PHG4TruthInfoContainer::Map  &  map = truthinfo->GetPrimaryMap(); // reference copy
```
which just take a reference and use the container. 


2) some code in evaluator assumed a pointer to a primary particle, but a G4 pointer pointer was passed instead. 

One example is suggested commit 975887ff152dd910a67a9cf36e5d54cf1e8e4bf8
```c++
....
// processing _ntp_cluster
 	  gembed   = trutheval->get_embed(g4particle);
	  gprimary = trutheval->is_primary(g4particle);
	 float cluster_data[33] = {_ievent,...gembed , gprimary ...  
```
Here we started with g4particle from the G4 particle list as the design. Then, gembed would always be false, since  trutheval->get_embed() assumed primary particle here. (I have another suggested commit f230dcf4fb1f4013c19fce56e3d04055c2589308 to allow trutheval->get_embed() take both G4 and primary PHG4Particles). Furthermore, gprimary will always be false, even the cluster is from the primary particle, because g4particle here is from the G4 particle list.

mccumbermike:
I hoped to have a full solution to the DST embedding tracing today, but I
think it will take me more time now. Here is an update of the basic problem
(documentation by email, yay).

There are two partial "copies" of the primary truth particle objects on the
node. We don't know a-priori which of the two pointers is being passed into
the evaluation. In the testing prior to the DST embedding running, the two
copies always shared the same track id since the secondaries were only
produced afterward processing the primaries and matching could be done with
just the track id regardless of the type of pointer passed. When embedding
in the DST secondaries are created in-between the primary processing so the
newly embedded particles get different ids when copied into the second map
breaking one of the basic assumptions made during the evaluator development.

For reasons previously stated, we cannot in the tracking evaluator simply
trace all the way through to the primary in the first map as has been done
in many calorimeter evaluation methods. A different approach is needed to
replace the track id matching assumption.

Thus the calls like:
particle1->get_track_id() == particle2->get_track_id()
and
g4hit->get_trk_id() == particle->get_track_id()
that appear scattered throughout the evaluation are now invalidated during
the embedded DST running for primaries run in the second pass. I've been
working at a solution that inspects the particles differently and
successfully replaced those lines in the tracking evaluation to utilize a
common >30 line function call that is needed to match between the two
different copies of the truth particles. This is being done with
mccumbermike/coresoftware/eval_devel branch if you wish to view my progress.

However, this problem also appears to a lesser extent for the calorimeter
and the jets... so I still have more work to do before I'll generate a new
pull request.

Now for your help if you've read this far, at the moment I'm relying very
heavily on the following undocumented rules for the PHG4TruthInfoContainer:
(1) both copies of the primary particles will have a parent id == 0
(2) primaries in the primary truth map will have primary id == -1
(3) primaries in the total truth map will have primary id == particle id in
the first map
I'm using these rules to both identify primaries and understand which map
there were drawn from. If anyone knows a mode of running that invalidates
these rules, I will probably need to know that sooner than later.

Mike


--
Michael P. McCumber, PhD
505-709-8161

PR #74, opened 2015-10-09
Tpc testing2

alandion:
This is take two on the pull request to get the TPC merged completely into master.  In this attempt, the changes necessary for the TPC tracking to run well in central HIJING have been disabled by default, so (fingers crossed) silicon tracking shouldn't be affected.


mccumbermike:
Validated. No problems with the memory consumption. 2000 of 2000 jobs finished. Performance same as before:
crosscheck

Merging...

PR #75, opened 2015-10-11
New truth matching conditions to handle truth table copies during embedding

mccumbermike:
Okay @blackcathj , I believe I've done it now. I wrote a new class BaseTruthEval to handle common tasks between the SvtxTruthEval and CaloTruthEval. This reduces the duplication of code between the two modules and ensures those portions are always doing the same things. Inside that new class, I have changed the truth id matching criteria to this: 

```
  if ((p1->get_primary_id() == -1) &&
      (p2->get_primary_id() == -1)) {
    // both particles are from the primary truth map, use track ids
    if (p1->get_track_id() == p2->get_track_id()) {
      are_same_particle = true;
    }
  } else if ((p1->get_primary_id() != -1) &&
	     (p2->get_primary_id() != -1)) {
    // both particles are from the total truth map, use track ids
    if (p1->get_track_id() == p2->get_track_id()) {
      are_same_particle = true;
    }
  } else if (is_primary(p1) && is_primary(p2)) {
    // both particles are primary
    if ((p1->get_primary_id() == -1) &&
	(p2->get_primary_id() != -1)) {
      // only the first particle is from the primary truth map, use one track id and one primary id
      if (p1->get_track_id() == p2->get_primary_id()) {
	are_same_particle = true;
      }
    } else if ((p1->get_primary_id() != -1) &&
	       (p2->get_primary_id() == -1)) {
      // only the second particle is from the primary truth map, use one track id and one primary id
      if (p1->get_primary_id() == p2->get_track_id()) {
	are_same_particle = true;
      }
    }
  }
```
This handles the special case where the two primary copies are from separate maps in the 3rd and 4th else-if blocks.

I see that the SVTX and CEMC evaluation is working properly for a handful of electrons thrown near a handful of positrons that were written out to a DST. I have also run the branch through my standard tracking examination (to see I haven't borked anything for the standard tracking eval:
crosscheck

Jin, could you grab the branch and see if it solves your issue with the DST embedded eval as intended?

### DIFF
--- a/simulation/g4simulation/g4eval/BaseTruthEval.C
+++ b/simulation/g4simulation/g4eval/BaseTruthEval.C
@@ -1,0 +1,226 @@
+
+#include "BaseTruthEval.h"
+
+#include <fun4all/getClass.h>
+#include <phool/PHCompositeNode.h>
+#include <g4main/PHG4TruthInfoContainer.h>
+#include <g4main/PHG4Particle.h>
+#include <g4main/PHG4Hit.h>
+
+#include <cstdlib>
+#include <float.h>
+#include <cassert>
+#include <iostream>
+
+using namespace std;
+
+BaseTruthEval::BaseTruthEval(PHCompositeNode* topNode)
+  : _truthinfo(NULL),
+    _strict(true) {
+  get_node_pointers(topNode);
+}
+
+void BaseTruthEval::next_event(PHCompositeNode* topNode) {
+  get_node_pointers(topNode);
+}
+      
+PHG4Particle* BaseTruthEval::get_particle(PHG4Hit* g4hit) {
+
+  if (_strict) assert(g4hit);
+  else if (!g4hit) return NULL;
+  
+  PHG4Particle* particle = _truthinfo->GetHit( g4hit->get_trkid() );
+  if (_strict) assert(particle); 
+
+  return particle;
+}
+
+int BaseTruthEval::get_embed(PHG4Particle* particle) {
+
+  if (_strict) assert(particle);
+  else if (!particle) return 0;
+
+  if (!is_primary(particle)) return 0;
+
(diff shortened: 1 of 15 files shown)


blackcathj:
Hi, @mccumbermike 

Thanks for the hard work sorting out the primary/all particle list associations. This branch has already became very useful for me when looking at the embedding data. I am running more checks via analysis too. I think if there is no more comments from simulation meeting today, we shall merge this pull request. 

### Minor suggestion: 

Passing of strict flags down to new base truth eval in PHG4TruthInfoContainer and SvtxTruthEval:
```c++
void set_strict(bool strict) {_strict = strict; _basetrutheval.set_strict(strict);}
```

### A discussion:

Following discussion in #70, regarding these rules that evaluation (any other truth analysis) require, we might consider formalize it in the truth container and PHG4Particle code + add these conventions to variable description via Doxygen. Possible implementations are

(1) both copies of the primary particles will have a parent id == 0, e.g.
```c++
PHG4Particle
{
...
bool is_primary() const {return get_parent_id()==0;} // primary particle should have track_id == 0
...
}
```

(2) primaries in the primary truth map will have primary id == -1 e.g.
```c++
PHG4TruthInfoContainer::ConstIterator
PHG4TruthInfoContainer::AddPrimaryParticle(PHG4Particle *newhit)
{
...
  boost::tie(it,added) = primary_particle_map.insert(std::make_pair(key,newhit));
  newhit -> set_primary_id(-1); // particle in the primary list should have primary_id == -1
...
}
```

(3) primaries in the total truth map will have primary id == particle id in
the first map : this seems already enforced by truth actions.

mccumbermike:
Good catch, thought I had those and did not. Add the passing down of the set_strict option to both SvtxTruthEval and CaloTruthEval.

mccumbermike:
Approved at the meeting. Merging...

PR #77, opened 2015-10-16
Fault tolerant evaluator defaults with destructor error reporting

mccumbermike:
As discussed as a possibility at this week's meeting, I switched the default in this merge to the fault tolerant strict=false setting. The evaluators will now tally the errors encountered and report at SubsysReco::End() the number of errors encountered if any. Setting the eval verbosity negative will quench even these messages. The location of the errors can be examined later by switching back to the strict evaluation setting and debugging from there so I have not added a complex location tracking. I changed the evalstack object scopes so that the destructor and associated error reporting is called only once per process in the g4eval SubsysRecos.

It is likely there will still be some spot in the eval I haven't considered, but when these are reported, I will add more assert/error_counts in those spots as well. But we should improve a bit with a more robust behavior and still find our errors.

The eval logic remains untouched.


blackcathj:
Thanks @mccumbermike ! 

Tried with the embedding production output, where truth structure was broken to save output size for hijing background. Apparently the report worked well (for 10 events):
```
SvtxEvaluator::End() - Error Count: 3207801426
SvtxVertexEval::~SvtxVertexEval() - Error Count: 1241763
SvtxClusterEval::~SvtxClusterEval() - Error Count: 57654
BaseTruthEval::~BaseTruthEval() - Error Count: 3206502009
```
As it is already discussed in the meeting and there has been no further comment, I will just merge this pull request then.

PR #80, opened 2015-10-21
New Projections of Tracks into Calorimeters

mccumbermike:
Jin reported trouble in #69 that for some radii values he was getting flip-flop phi values while projecting into the calorimeter. Jin and I misunderstood what the next step was on getting this code into master and I promised to take another look to make sure I'm confident with the new projection tool and generate this pull request.

One issue was that the SvtxTrackState was reporting position estimates only relative to the vertex, that is changed in this merge to absolute coordinates. Another issue was that the projection code was written to accept a perigee coordinate and some replication of these was likely done incorrectly when switching to the new interface, probably at minimum that conversion was missing some sense of the charge sign. In principle that code could have been patch as Jin had proposed.

However we also wish to have a code that will accept an arbitrary state vector and make the projection from there, not just from the perigee. This is useful for grabbing different state vectors when those are available. So I decided to rewrite the projection calls from top to bottom with a new set of manipulations. I've tried to make the projection process explicit in the new algorithm so others can follow along. I tested the code in a standalone macro to look for bugs, here is a look at that output:

screen shot 2015-10-21 at 2 06 53 pm

I've also run the evaluation for some electrons of different momentum to see that the located cluster positions are tightly correlated in both angle and energy:

ecluster

and for the projected 3x3 momentum:

e3x3

The module runs quickly in peripheral events, so I don't think this change will appear on anyone's profile list. This pull request would replace #69 and #69 should not be merged and closed if this path is selected.


blackcathj:
Two sets of tests using the EMCal analysis code and HCal energy sum. This new algorithm worked well: 

## EMCal and inner HCal projections

Projecting the track into the calorimeters to check the shower response in the neighboring towers. Centered shower in EMCal says the track projection is consistent between tracking and calorimetry. The inner HCal is off for known reason of starting bin counting problem. 

10 GeV negatively charged pion shower in EMCal and inner HCal, plotted as tower energy VS tower's distance from the track projection
g4hits_sphenix_pi-_eta0_10gev root_ana root_drawecal_pdst_trackprojection_checks_cut_all_event

## Three calorimeter total energy sums

3x3-tower sums of all three calorimeter layers for 10 and 24 GeV pi- shower is shown below. Track projections are used to define the location of 3x3 cluster. The sum covered majority of hadron energy but with significant invisible component and side leakages as expected for hadron showers. 
image

PR #86, opened 2015-11-21
Storage Revision for PHG4TruthInfoContainer

mccumbermike:
We should discuss this in a simulation meeting prior to merge, so I'm submitting this now for early review.

I've completed and tested the revision of the truth info container in this branch. The new code contains a single storage map for particles and a single storage map for vertexes (no more comparisons between primary copies needed---there can be only one). Primaries and secondaries are segregated to positive and negative key identifiers respectively. Thus the primary ids in the first round of G4 processing are not modified, preserving look up to the generator level. We can add a barcode based lookup in another round of revision to do that even better. There is a new set of range functions for iterating over the primaries and secondaries independently which have been used to replace parts of the core library where the deprecated primary map was returned. Additional runs of G4 append either to the top of the map for primaries or the bottom for secondaries. Primary vertexes can be identified by the sign of the id, something not possible previously without searching through the particle list. Information about primary ancestry is now passed down during processing instead of a tree-spanning trace at the end of the event, so the module should also consume less CPU now. Other revisions were made to pass the embed flag integer value all the way through the simulation and into the output. I envision this as a way to test the fidelity of multiple vertex algorithms (each pythia event gets a different embed flag), but there are lots of things that tool could be utilized to do. This merge as currently coded will break backwards DST compatibility---no versioning has been implemented on this object and the interface is still evolving anyway.

The evaluator chews happily on the new format as no assumptions on trackid values were made in the initial design. The new code doesn't interfere with the proper evaluation of the data as demonstrated here:

crosscheck

I haven't yet optimized the evaluation for the new format, but that will be a next step. That goal will alleviate some of the issue in the TPC evaluation where we were spending >50% of CPU trying to decide if two particle pointers were in fact the same particle under the different maps. Some of the speed issue might partially resolve already since this version of the truth info will send that code down the faster side of the evaluation, but the plan is to update the evaluator to strip those if-statements.

One big interface change is the removal of the GetPrimaryMap() function. Users should instead use the GetPrimaryParticleRange() which will result in the same basic functionality. I also rename the *Hit* methods to *Particle* methods to reflect the actual container contents as this was really very confusing when I first started using this object in the evaluation.

Jin, I've tried this code in an embedded event and I see it appending correctly, but if it is easy for you to try it out with a DST created with the branch in a real-world scenario, that would give me more confidence this revision is working as expected.

### DIFF
--- a/offline/packages/NodeDump/DumpPHG4TruthInfoContainer.cc
+++ b/offline/packages/NodeDump/DumpPHG4TruthInfoContainer.cc
@@ -40,7 +40,7 @@ int DumpPHG4TruthInfoContainer::process_Node(PHNode *myNode)
           (*vtxiter->second).identify(*fout);
 	}
       PHG4TruthInfoContainer::ConstIterator particle_iter;
-      PHG4TruthInfoContainer::ConstRange particlebegin_end = truthcontainer->GetHitRange();
+      PHG4TruthInfoContainer::ConstRange particlebegin_end = truthcontainer->GetParticleRange();
           for (particle_iter = particlebegin_end.first; particle_iter != particlebegin_end.second; particle_iter++)
             {
 	      *fout << "particle number: " << particle_iter->first << endl;
--- a/simulation/g4simulation/g4detectors/PHG4BlockSteppingAction.cc
+++ b/simulation/g4simulation/g4detectors/PHG4BlockSteppingAction.cc
@@ -85,15 +85,14 @@ bool PHG4BlockSteppingAction::UserSteppingAction( const G4Step* aStep, bool )
 	  hit->set_t( 0, prePoint->GetGlobalTime() / nanosecond );
 	  //set the track ID
 	  {
-            int trkoffset = 0;
+            hit->set_trkid(aTrack->GetTrackID());
             if ( G4VUserTrackInformation* p = aTrack->GetUserInformation() )
 	      {
 		if ( PHG4TrackUserInfoV1* pp = dynamic_cast<PHG4TrackUserInfoV1*>(p) )
 		  {
-		    trkoffset = pp->GetTrackIdOffset();
+		    hit->set_trkid(pp->GetUserTrackId());
 		  }
 	      }
-            hit->set_trkid(aTrack->GetTrackID() + trkoffset);
 	  }
 
 	  //set the initial energy deposit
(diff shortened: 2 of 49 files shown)


blackcathj:
Following Mike's request, I am checking this proposed branch with the DST based embedding (G4 simulation to produce background -> background DST file -> embed second simulation -> new embedded DST file). The result looks reasonable. 

## More details

To simplify the CPU intensity, the background used only one muon from one event. And the signal particle is one electron which is embedded on top of the muon DST. 

### Background DST file

Background DST file consist of one muon simulated in full sPHENIX. The new truth container is illustrated below via  T->Scan("DST.G4TruthInfo.identify()"):

```
---particlemap--------------------------
...Secondary particles with negative ID (as defined in the new rule)...
particle id -3
PHG4Particlev2 name: e-, track id: -3, vtxid: -3, parent id: 1, primary id: 1, pid: 11, px: -0.00054158, py: 0.000487923, pz: -0.000675439, e: 0.00
111746
...
particle id 1   <- input muon as primary particle
PHG4Particlev2 name: mu-, track id: 1, vtxid: 1, parent id: 0, primary id: 1, pid: 13, px: -3.29947, py: 23.7721, pz: -2.31893, e: 24.112 
...
---vtxmap-------------------------------
vtx id: 1
vtx position: x: -3.47657, y: 8.27957, z: -1.25928, t: 0
---list of embeded track flags-------------------
embeded track id: 1 flag: 1
---list of embeded vtx flags-------------------
embeded vertex id: 1 flag: 1
```

### Embedded DST file

Then an electron is embedded to the above DST file. Do another T->Scan("DST.G4TruthInfo.identify()"):

```
...
---particlemap--------------------------
...Secondary particles with negative ID not shown (as defined in the new rule)...
particle id 1
PHG4Particlev2 name: mu-, track id: 1, vtxid: 1, parent id: 0, primary id: 1, pid: 13, px: -3.29947, py: 23.7721, pz: -2.31893, e: 24.112
particle id 2
PHG4Particlev2 name: e-, track id: 2, vtxid: 2, parent id: 0, primary id: 2, pid: 11, px: -14.9892, py: -18.5365, pz: 2.77876, e: 24
---vtxmap-------------------------------
...
vtx id: 1
vtx position: x: -3.47657, y: 8.27957, z: -1.25928, t: 0
vtx id: 2
vtx position: x: -3.47657, y: 8.27957, z: -1.25928, t: 0
---list of embeded track flags-------------------
embeded track id: 1 flag: 1
embeded track id: 2 flag: 1
---list of embeded vtx flags-------------------
embeded vertex id: 1 flag: 1
embeded vertex id: 2 flag: 1
```

blackcathj:
Also went through the new software with a few more comments:

* I like the idea a lot in general. Besides fast primary look up, the new code also tag track ID via a user bar code (```aTrack->GetUserInformation()->GetUserTrackId()```), so we do not need to rely on the older implicit rule on ID shiftings. 
* We left with one implicit rule that sign of trackID note whether it is primary. We might want to make sure it is clear to users via documentation (note the comments for Map particlemap is not in Doxygen style) and via interface function (e.g. ```PHG4TruthInfoContainer::is_primary(PHG4Particle * p) {return p->get_track_id()>0;}```)
* A few place in the nightly build need a patch based on the new primary interface:
```
DumpPHG4TruthInfoContainer.cc	/sPHENIX/simulation/coresoftware/offline/packages/NodeDump	line 43	C/C++ Problem
 G4SnglNtuple.cc	/sPHENIX/simulation/coresoftware/simulation/g4simulation/g4histos	line 62	C/C++ Problem
 G4SnglTree.cc	/sPHENIX/simulation/coresoftware/simulation/g4simulation/g4histos	line 87	C/C++ Problem
mFillG4SnglHTCContainer.cc	/sPHENIX/simulation/coresoftware/simulation/g4simulation/g4picoDst	line 95	C/C++ Problem
```

mccumbermike:
Yes, good catch. I forgot to run a full build test and these modules were outside of my quick build script. I've patched the remaining spots in the new build. I've also added the methods and doxygen comments you wanted.

PR #87, opened 2015-11-21
Support timing cut in Cell reco

blackcathj:
As an action item from the last simulation meeting, this pull request introduces a timing cut in the cell reco to the mimic the effective ADC integration window for sPHENIX calorimeters.  

The idea is to apply this cut after PHG4Hit stage, since the timing span for each hit is small and it would allow us to tune the timing window post Geant production of hits. Also only a simple max-timing cut was applied, since at length scale of 100ns the start time variation is effectively negligible. The cut value can be changed via macro inputs. However, in near future, we plan to read it via the parameter node, after Chris finish building the parameter database infrastructure.

The default timing window is 0 to infinity, i.e. consistent with no timing cut as used in previous analysis. The timing cut can be activated for all calorimeters via macro such as this example: 
https://github.com/sPHENIX-Collaboration/macros/commit/1da9e30e9f21f63057b3ade72b401154645a3285 

Here is a comparison between three different timing window cut
1. No cut as default and consistent as previous analysis
2. 100ns timing cut to mimic 0-100ns ADC integration window we plan to use
3. 30ns timing cut to mimic fit for the amplitude the ADC-timing peak. 
For this fast test, 100 40 GeV pi- from the pre-CDR production was used. The change in the distribution for the total cluster energy is obvious between no-cut and timing cut applied:
image

### DIFF
--- a/simulation/g4simulation/g4detectors/PHG4BlockCellReco.cc
+++ b/simulation/g4simulation/g4detectors/PHG4BlockCellReco.cc
@@ -22,6 +22,7 @@
 #include <cstdlib>
 #include <iostream>
 #include <sstream>
+#include <limits>       // std::numeric_limits
 
 using namespace std;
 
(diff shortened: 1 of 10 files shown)


blackcathj:
Approved in the meeting. Merging 

Meanwhile, before the macro repository is updated, this timing cut is NOT activated.

PR #97, opened 2015-12-17
Forward cal eval

johnlajoie:
Updates to include the forward HCAL and EMC for fsPHENIX, including cell reco, towering and jet evaluation.


mccumbermike:
No problems with the barrel tracking or associated eval libraries:
crosscheck
I took a look by hand at 25 GeV/c jets from pythia in the barrel and didn't see issues, but we should really have automated performance testing in the future for those objects.

simulation/g4simulation/g4detectors/PHG4CylinderCell.h
```diff
@@ -54,6 +54,13 @@ class PHG4CylinderCell : public PHObject
   virtual int get_ladder_phi_index() const {return -9999;}
   virtual void set_ladder_z_index(const int i) {return;}
   virtual int get_ladder_z_index() const {return -9999;}
+
+  virtual int get_j_index() const {return -9999;}
+  virtual void set_j_index(const int i) const {return;}
```
blackcathj: @johnlajoie Came by these lines during cross check #97 , these three set functions should be 

```
virtual void set_j_index(const int i) {return;}
virtual void set_k_index(const int i) {return;}
virtual void set_l_index(const int i) {return;}
```
i.e. without ```const```

johnlajoie:
Commit 2a602a4 only affects forward calorimeter routines, so hopefully it doesn't require re-running test suites for the central barrel.  I should have had this included in the original pull request.

blackcathj:
Thanks @johnlajoie  again for the quick fixes. 

Then I further finished checking new code in g4cemc and g4detectors. Did some test run. 

To me, only one minor issue left: zero-energy PHG4hit was not cleared up, which could be not associated with any truth track (only truth track with none-zero energy was configured to save in Stepping Funtion of both FEMC and FHCAL). A fix is proposed via https://github.com/johnlajoie/coresoftware/pull/1

PR #98, opened 2015-12-23
Pythia8 Forward Jet Trigger

johnlajoie:
A simple, configurable forward anti-kT (truth) jet trigger for PHPythia8.

### DIFF
--- a/generators/PHPythia8/Makefile.am
+++ b/generators/PHPythia8/Makefile.am
@@ -16,7 +16,8 @@ lib_LTLIBRARIES = libPHPythia8.la
 include_HEADERS = \
   PHPythia8.h \
   PHPy8GenTrigger.h \
-  PHPy8ParticleTrigger.h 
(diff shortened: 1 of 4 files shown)


mccumbermike:
Looks good to me, I did only a compile check. I see this as being very useful for the barrel too. I suggest only that you consider s/Fwd/Truth/g'ing the name so new comers are aware of this. Merge at your discretion.

johnlajoie:
I would propose to just drop the "Fwd" from the name (PHPy8JetTrigger) and leave it at that.  It's a trigger at the Pythia8 level and just uses a generic (programmable) acceptance in eta, i don't see that it should directly reference fsPHENIX or sPHENIX, or maybe I misunderstand your comment?

johnlajoie:
The renaming has been committed to pull request #99, which overlaps this pull request, and I will close this request.

PR #99, opened 2015-12-27
Updated Forward EMC to match re-use of PHENIX EMCal, add PHPy8JetTrigger

johnlajoie:
Adjusted the number of layers and the scintillator and absorber thickness to match PHENIX EMCal

### DIFF
--- a/generators/PHPythia8/Makefile.am
+++ b/generators/PHPythia8/Makefile.am
@@ -17,7 +17,7 @@ include_HEADERS = \
   PHPythia8.h \
   PHPy8GenTrigger.h \
   PHPy8ParticleTrigger.h \
-  PHPy8FwdJetTrigger.h 
+  PHPy8JetTrigger.h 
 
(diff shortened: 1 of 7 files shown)


blackcathj:
This pull request overlaps with https://github.com/sPHENIX-Collaboration/coresoftware/pull/98 in many part. And for #98, I agree with Mike on the naming of the jet module to make is available for more people. Would you like to update the pull request with that? And since the commit in Pull#98 is also in #99, we just need to update #99 and merge it.

johnlajoie:
Sorry for the overlap - that was not my intent, in fact, I wanted to keep these two entirely separate. I should have rebase'd before commiting the EMCal stuff.  To go forward I have updated the trigger name in a commit to this branch, and I'll close the pull request for #98.

blackcathj:
Tried to trigger on pT>30GeV/c jet in the central acceptance. The new code works well for me. Thanks @johnlajoie ! 

Merging this pull request.

Issue #102, opened 2016-01-04
Inheritance of PHG4ProjCrystalCalorimeterDetector


pinkenburg:
cppcheck showed a lot of shadowed variables. I assume it started as a plain copy of the PHG4CrystalCalorimeterDetector base class. This is just the wrong way of going about inheriting and this needs to be fixed before this propagates into other code. To run cppcheck:

cppcheck --enable=all PHG4ProjCrystalCalorimeterDetector.cc

(this also prints out style warnings which are irrelevant). Another issue is the wrong order of includes. One ALWAYS has to put the includes which use quotes (#include "...") first. Otherwise the intended search path (using local inludes first) when using separate include install directories does not work! In implementations one often gets away with it but in header files this is fatal and is extremely hard to debug.

mccumbermike:
The issue here is the unnecessary protected declarations in the parent class. One should really use private for storage members by default and call the public interface in the daughter class.

blackcathj:
Closed via pull request #103

PR #103, opened 2016-01-11
Fix proj crystal calorimeter

nfeege:
Remove shadowed variables pointed out in issue https://github.com/sPHENIX-Collaboration/coresoftware/issues/102


blackcathj:
Hi @nfeege 

Thank you for looking into this problem. 

Meanwhile, I do have one comment: At this moment, 12 interface functions in PHG4ProjCrystalCalorimeterDetector still shadows those from PHG4CrystalCalorimeterDetector, e.g. PHG4ProjCrystalCalorimeterDetector::SetPlace() shadows PHG4CrystalCalorimeterDetector::SetPlace(). It is OK for get things running, but could be confusing in future because it is tricky for user to find whether SetPlace() is called in which of the two classes. 

A minor change to solve this confusion could be base PHG4ProjCrystalCalorimeterDetector on PHG4Detector, rather than on PHG4CrystalCalorimeterDetector. It looks one line change to me, but I could missed some additional connection between the two classes.

nfeege:
Hi @blackcathj ,

I mainly used the inheritance to make the selection of 'projective' or 'non-projective' version work in PHG4CrystalCalorimeterSubsystem.cc (lines 40 - 48, https://github.com/sPHENIX-Collaboration/coresoftware/blob/master/simulation/g4simulation/g4detectors/PHG4CrystalCalorimeterSubsystem.cc).

Do you have a better suggestion of how to do that without the inheritance? If yes, I'd be glad to remove it.

pinkenburg:
I would say to go with Mikes suggestion (which should also be 
implemented where I used protected variables like in our single particle 
generators). If you inherit from a class, make the data members of the 
parent private (not protected) so they are not shared with the derived 
classes. Then use the public methods of the parent class (if you want to 
make sure they are only accessible by the derived classes make those 
protected) to access those variables. As a concrete example in your 
projective implementation which inherits from the 
PHG4CrystalCalorimeterSubsystem class, just remove the SetPlace() method 
(and the associated variables). Then it will use the method from the 
parent. To access the variable which is now private in the parent class 
use the public/protected access method.

I thought of using something similar for the outer and inner hcal which 
do have a large overlap but then wanted the flexibility to change things 
without impacting the other hcal. I still might revisit this once things 
have settled down

Chris


On 1/12/2016 10:25 AM, Nils Feege wrote:
>
> Hi @blackcathj <https://github.com/blackcathj> ,
>
> I mainly used the inheritance to make the selection of 'projective' or 
> 'non-projective' version work in PHG4CrystalCalorimeterSubsystem.cc 
> (lines 40 - 48, 
> https://github.com/sPHENIX-Collaboration/coresoftware/blob/master/simulation/g4simulation/g4detectors/PHG4CrystalCalorimeterSubsystem.cc).
>
> Do you have a better suggestion of how to do that without the 
> inheritance? If yes, I'd be glad to remove it.
>
> —
> <https://github.com/sPHENIX-Collaboration/coresoftware/pull/103#issuecomment-170946324>.
>

-- 

Upton, NY 11973-5000

nfeege:
Thanks for the details, I'll implement this today for the CrystalCalorimeter.

nfeege:
@blackcathj I think I fixed the remaining issues.

blackcathj:
As discussed in the last meeting, we will merge once this problem is fixed. And the current version satisfied both GCC and CPPCHECK. And no more shadowing of function or variable. Merging.

PR #112, opened 2016-03-12
Outer Hcal prototype for test beam

pinkenburg:
This implementation uses the corner coordinates of the bottom steel plate to create the correct steel plate shape. Using the coverage angle of 360/320 it puts the top steel plate right at the corner coordinates of the top steel plate. The scintillators are implemented as extruded volumes which again creates the correct shape. The scintilators protrude out the back by about 5mm, just like the prototype in the high bay. It uses an assembly volume to position the steel and scintillators, so one can add easily the support structures without overlaps. Everything can be moved/rotated so the position of the outer hcal can be adjusted on the macro level if needed. The g4hit output should be usable by our cell reconstruction and thus feeds directly into our analysis chain.


pinkenburg:
This merge just adds new classes, no existing code was changed

PR #115, opened 2016-03-18
Pythia6 Generator

nfeege:
Adding Pythia6 as event generator. The code uses the Pythia6 wrapper provided by HepMC to generate events in HepMC format and stores them on the node tree. It uses a configuration file to configure Pythia settings in the same way PHPythia in the PHENIX software does.

The code runs, calls Pythia6, and generates some HepMC output which can be passed through the Geant4 simulation. Crosschecks of the output still need to be done. I just initiated the pull request so everyone who is interested in this can have a look.

This generator will allow us to do Pythia6 / Pyhtia8 crosschecks in the sPHENIX framework and run e+p simulations for EIC.


mccumbermike:
Jin asked me to review the unit translation from PYTHIA into our output. So I ran a single event and looked for an example of off-vertex decay in the input. Here is the example I found:

screen shot 2016-03-24 at 10 26 33 am

As you can see there is particle 104 generated at vertex -44. The units from pythia are GeV and mm. I then found the same particle in our final output here:

screen shot 2016-03-24 at 10 28 03 am
 
which is in our canonical GeV and CM units. I see that the units are passed correctly from input to output.

I have one suggestion for the pull request. The smearing interface is still available on the header but is commented out inside. We can smear at the HepMC reader, so I recommend also commenting out the header functions or someone will be confused why they don't work.

nfeege:
Thanks a lot, Mike!

I renamed PHPythia files, class and library to PHPythia6.

I also got rid of the (non-functional) vertex smearing interface.

PR #126, opened 2016-03-27
updated to allow PbSc tower type

johnlajoie:
This is the promised update to all the PbSc tower type to be used in the forward ECal.  Rather than just replace the PbGl tower type I added a new tower type so that in principle all three could be used.  Note that there are corresponding pull requests in calibrations and macros. 

This is not yet complete, but I wanted this out there so people could have a look. What remains is to have a look at the towering - we now have a detector that allows multiple different tower types all of which potentially have a different sampling fraction and  calibration to true energy. I want to have a look and see if there is graceful way to handle this. 

Here's a screenshot with the new forward ECAL in place:

new


johnlajoie:
The latest commit is my suggestion for how to handle the unique problems that the forward ECAL with multiple tower types presents.   Let me recap a bit...

The fsPHENIX forward ECAL has been modified to allow multiple tower types - MPC crystals, PbGl crystals, or PbSc towers.  The geometry file allows these towers to be defined by a type variable in the file, and the current incarnation has MPC crystals at high rapidity surrounded by PbSc towers. At the workfest it was decided that it was best to keep the detector geometry implementation on one object for two reasons - right now both reside in the same mother volume, and this makes is "plug-and-play" compatible with a forward FEMC for the EIC detector, which may be different. 

Under this approach, we now need to be able to handle tower digitization and calibration differently for each tower type.  These changes seek to address this by propagating a "TowerType" with the RawGeom and RawTower objects. The digitization and calibration modules maintain a TowerType that can be used to trigger action only for a given tower type within the detector they are attached to; other types are alone.   The tower type is set to -1 for digitization and calibration by default meaning it will be ignored and all other applications will still work as advertised.  For the fsPHENIX there are calibration and digitization modules for both the MPC and PbSc towers, selected by TowerType. 

The advantage of this approach (I feel) is that the split is then only where it it necessary - in generating calibrated towers.  It does not generate a bifurcation further upstream or downstream - for example, clustering can be done on all calibrated towers for the FEMC, without having different tower types for the FEMC in different nodes. 

There are corresponding changes in the macros repository. Let me know what you think. I am running some performance evaluations and will post results (and the inevitable fixes) shortly.

sPHENIX-GitHub-l:
Hi John,

I would agree with that, if on the clustering level all types of towers 
are used and it just affects the calibration step this makes sense.

Chris

On 03/30/2016 03:36 PM, John Lajoie wrote:
>
> The latest commit is my suggestion for how to handle the unique 
> problems that the forward ECAL with multiple tower types presents. Let 
> me recap a bit...
>
> The fsPHENIX forward ECAL has been modified to allow multiple tower 
> types - MPC crystals, PbGl crystals, or PbSc towers. The geometry file 
> allows these towers to be defined by a type variable in the file, and 
> the current incarnation has MPC crystals at high rapidity surrounded 
> by PbSc towers. At the workfest it was decided that it was best to 
> keep the detector geometry implementation on one object for two 
> reasons - right now both reside in the same mother volume, and this 
> makes is "plug-and-play" compatible with a forward FEMC for the EIC 
> detector, which may be different.
>
> Under this approach, we now need to be able to handle tower 
> digitization and calibration differently for each tower type. These 
> changes seek to address this by propagating a "TowerType" with the 
> RawGeom and RawTower objects. The digitization and calibration modules 
> maintain a TowerType that can be used to trigger action only for a 
> given tower type within the detector they are attached to; other types 
> are alone. The tower type is set to -1 for digitization and 
> calibration by default meaning it will be ignored and all other 
> applications will still work as advertised. For the fsPHENIX there are 
> calibration and digitization modules for both the MPC and PbSc towers, 
> selected by TowerType.
>
> The advantage of this approach (I feel) is that the split is then only 
> where it it necessary - in generating calibrated towers. It does not 
> generate a bifurcation further upstream or downstream - for example, 
> clustering can be done on all calibrated towers for the FEMC, without 
> having different tower types for the FEMC in different nodes.
>
> There are corresponding changes in the macros repository. Let me know 
> what you think. I am running some performance evaluations and will 
> post results (and the inevitable fixes) shortly.
>
> —
> You are receiving this because you are subscribed to this thread.
> <https://github.com/sPHENIX-Collaboration/coresoftware/pull/126#issuecomment-203599903>
>
>
>
> _______________________________________________
> sPHENIX-GitHub-l mailing list
> sPHENIX-GitHub-l@lists.bnl.gov
> https://lists.bnl.gov/mailman/listinfo/sphenix-github-l


-- 

*********************************************************

Christopher H. Pinkenburg      ; pinkenburg@bnl.gov

Brookhaven National Laboratory ; phone: (631) 344-5692
Physics Department Bldg 510 C  ; fax:   (631) 344-3253
Upton, NY 11973-5000

http://www.phenix.bnl.gov/~pinkenbu/

*********************************************************

johnlajoie:
Just a few plots to demonstrate that the basic G4Hits from the different detector types is working properly.  I sent in 20GeV negative muons into the FEMC and separate events based off the energy weighted average of the pseudorapidity of the hits so I know what detector type they are going into. Up to a pseudorapidity of 3.0 you are in the PbSc, from 3.0 to 3.3 the events are mixed (the PbW is a square array so 3.0 is at the corner and 3.3 is in the middle of the side, or the inscribed circle) and up to 3.3 to 4.0 is all in the PbW crystals. 

The sampling fraction for the PbSc:
sf_pbsc
 is consistent with the sampling fraction we know and love from PHENIX.  In the PbW: 

sf_pbw

The sampling fraction is essentially one as we collect all the energy in the crystal material. 

To me this demonstrates that the detector structure is implemented correctly.  I'm working next to demonstrate that the towering generates a consistent set of calibrated tower energies given the two digitization and calibration modules running on the same detector type.

johnlajoie:
With all the latest changes I think this is ready for merge. I have confirmed that the tower digitization and calibration is proceeding correctly for both the PbSc and PbW tower types. As an example, here's some plots from the jet evaluator (R=0.5) working on single e- events at 20GeV compared to the G4Hits.  The jet evaluator works on the calibrated towers as opposed to the raw G4Hits. 

Here's the G4Hits output for showers in the PbSc (corrected by sampling fraction):
pbsc_g4hits

and here's the results for the jet evaluator energy 

pbsc

The x-axis scales are different, but the means and distributions are the same. That tells me that the digitization and calibration are faithfully reproducing the energy in the PbSc towers. (What I don't understand is why the sf-corrected energy isn't closer to the incident 20GeV energy? This does need a closer look.)

Here's the same for the PbW, without a sf correction (the crystal collects all energy).  The G4Hits: 
pbw_g4hits

and from the forward jet evaluation: 

pbw

blackcathj:
Hi, @johnlajoie 

Thank you for the update. Three suggestions with suggested new code proposed to your development branch via pull requests:

1. The tower type tagging in RawTower seems to be redundant. The tower type tag has already been done in the RawTowerGeom and remain stable througout one run. RawTower insteand store information changes event to event, like energy in tower. 
Therefore I suggest remove RawTowerv2 : https://github.com/johnlajoie/macros/pull/3

2.  Use the newly defined calibration ROOT env variable, so it will be eaiser to overwride this variable to use a local check out of calibration repo for testing

3. Add pythia6 generator as developed by @nfeege . 

The last two items are suggested via https://github.com/johnlajoie/macros/pull/2

johnlajoie:
Hi @blackcathj, thanks for teh comments and code review: 

1. I like these suggested changes a lot, much cleaner and makes more sense. Merged pull request. 

2 & 3 are great as well, also merged.

blackcathj:
This is to confirm this pull request passed the QA test (new simulation of 100 24GeV pi- VS latest production). Merging 


singlepart_master_qa root_qa rootqa_draw_calorimeter_sum_trackproj

singlepart_master_qa root_qa rootqa_draw_calorimeter_sum_cluster

PR #133, opened 2016-04-13
Better initial vertexing

mccumbermike:
Ok, here is the follow up pull request I discussed this morning. I've decided to initialize the vertex with the simulated BBC vertex, catch this with an initial patterning reco code, fit that quick vertex, and pass it into the final tracking and vertexing code. The first two stages are setup to work with a BBC resolution in p+p and the tracker populated by a HIJING event---so it should satisfy both ends of the spectrum. Anyhow, the speed in central events is about the same as before so I'm not so far off on the CPU optimization.

Here is a look at getting the vertex guessing problem fixed for vertexes smeared throughout the beam diamond in the same way as shown this morning, solid points are this pull request.
crosscheck
You can see I've largely solved the 20% failure in the initial vertex guessing code.

I will toy around with replacing the BBC input with SVTX only code which I think should be workable for most of the trackers and satisfy my philosophical desire for a tracking code base that doesn't critically rely on other detectors---but I'm a little worried about designs with lots of fakes, confusion could run rampant in those cases. But practical demands to get us moving forward won the day today.

To build watchers: The BBC code previously resided in the g4vertex library with the composite global vertex. Since the g4hough needed to link against that without a circular dependency, I split g4vertex into g4bcc and g4vertex libraries. So I will need to make a pull request on the build utility and on the macros to prepare those things for these changes and the merges will need to happen at roughly the same times to avoid the worst aspects of the disruption.

I haven't ported any of these things into the TPC code base...  who is running support on that these days?


mccumbermike:
I updated this branch to vertex using only the SVTX. The performance fix is similar to using the BBC as a seed:
crosscheck

PR #136, opened 2016-04-19
Add preliminary geometry object to prototype towers for digitizer, fix coverity problems

pinkenburg:
Using the empty geometry object (which basically only contains the number of towers) the output of the prototype tower builder can now be fed into the digitizer. Coverity found issues in many stepping actions where the new shower objects wasn't filled properly for absorber hits (cut and paste error). Fixed some other benign warnings (like possible use of negative indices) which probably never happened.

### DIFF
--- a/simulation/g4simulation/g4cemc/RawTowerDigitizer.cc
+++ b/simulation/g4simulation/g4cemc/RawTowerDigitizer.cc
@@ -108,7 +108,6 @@ RawTowerDigitizer::process_event(PHCompositeNode *topNode)
       it != all_towers.second; ++it)
     {
       const RawTowerDefs::keytype key = it->second->get_id();
-     
       if(_tower_type>=0){
 	// Skip towers that don't match the type we are supposed to digitize
 	if(_tower_type != it->second->get_tower_type()) continue; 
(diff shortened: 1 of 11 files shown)


simulation/g4simulation/g4cemc/RawTowerDigitizer.cc
```diff
@@ -121,20 +120,21 @@ RawTowerDigitizer::process_event(PHCompositeNode *topNode)
...
       else if (_digi_algorithm == kSimple_photon_digitization)
+	{
         digi_tower = simple_photon_digitization(sim_tower);
+	}
       else
         {
 
           std::cout << Name() << "::" << detector << "::" << __PRETTY_FUNCTION__
               << " invalid digitization algorithm #" << _digi_algorithm
               << std::endl;
```
blackcathj: wouldn't a static code checker report memory leakage here?

pinkenburg:
no - the full code is here:

      RawTower *digi_tower = NULL;

      if (_digi_algorithm == kNo_digitization)
        {
          if (sim_tower)
	    {
            digi_tower = new RawTowerv1(*sim_tower);
	    }
        }
      else if (_digi_algorithm == kSimple_photon_digitization)
	{
        digi_tower = simple_photon_digitization(sim_tower);
	}
      else
        {

          std::cout << Name() << "::" << detector << "::" << __PRETTY_FUNCTION__
              << " invalid digitization algorithm #" << _digi_algorithm
              << std::endl;

          return Fun4AllReturnCodes::ABORTRUN;
        }

A new digi_tower is only created inside the if or else if, then else is not executed. If the code makes it to else it never made a new digi_tower

RawTower *digi_tower = NULL;

static code checkers should be able to figure this out

PR #141, opened 2016-05-02
mlp - added the class to maintain the temperatures for a given module.

mpurschke:
This is a class to store the temperature readings for the SiPMs on a tower by tower basis. This allows to structure the container in the same coordinate space as the real SiPM signals, and get the right temperature.
Since we are reading the temperature only every minute or so, we interpolate between readings for event times in between two measurements linearly in time.
One can also ask for the reading from a given entry if one wants to make more sophisticated interpolations.


blackcathj:
Thanks! Martin. It compile well for me. I am looking forward to the unpacking module that load these temperature readings to the RUN-node too.

PR #143, opened 2016-05-04
test faster post-hough cleaning for HIJING

alandion:
This is faster algorithm for the after-hough track-cleaning for the TPC (removing spurious hits).  The efficiency drops with pT faster than desired but hopefully is a good starting point for future central collision studies.


mccumbermike:
I ran a test on the MIE design to see we aren't disturbing that set of results in the merge. For some reason the CPU use exploded and the jobs don't finish, so I can comment on the actual performance. I still need to check the TPC code base (I'm looking into something else this weekend for Tony). But if the TPC is happy with it, I'll need to fork some of these objects/changes and update the branch prior to the merge (after double checking my first test wasn't somehow done incorrectly---which is possible as I flip back and forth between these different configurations).

mccumbermike:
We decided at the tracking meeting to come back with a different pull request to merge in a forked sPHENIXTracker object to bring the TPC tune into the build.

PR #146, opened 2016-05-09
Module to support ganging EMCal readout

blackcathj:
## Introduction

Adding a new tower analysis module to enable analysis of sPHENIX simulation to use multi-tower ganging of EMCal readout. For example 2x2-ganging as proposed for some CEMC de-scoping options. 

Technically, a new module ```RawTowerCombiner``` is introduced to combine MxN towers into one readout channel on an eta-phi tower grid exiting on the DST tree. Therefore, it needs to be called after ```RawTowerBuilder``` and before ```RawTowerDigitizer```. During the merging, truth structure (Cells and showers) is maintained. In the default mode, both RawTower and RawTowerGeometry are edited on the DST tree, rather than introducing a new node for combined towers. I found this is least intrusive to our analysis code base. 

If one need to use this module for ALD charge (e.g. photon position resolution, jet finding, etc.), please contact me directly, in order to speed up verification and feedback.

## Verification

The single particle simulation in the 2016-02-01 Geant4 production (```/sphenix/sim/sim01/production/2016-02-01```) was used for testing. 

One example is distance between best CEMC cluster from the trajectory projection of 24 GeV/c electrons. Green reference plot is default towering and blue curve is after 2x2 ganging. The ganged distribution is roughly twice wider while the central value remain zeroed.
2x2test

## Example macros

By default, this new module is NOT used in analysis and no ganging is applied. 

Example macro to enabling 2x2 ganging as proposed for some de-scoping otpions: https://github.com/blackcathj/macros/blob/EMCal2x2/macros/g4simulations/G4_CEmc_Spacal.C 
Or specifically these lines added:
```c++
void CEMC_Towers(int verbosity = 0) {
 ...
  // TowerBuilder
 ...

  // Make ganged output for CEMC
  if (combin_CEMC_tower_2x2)
  {
    // group CEMC RawTower to CEMC2x2
    RawTowerCombiner * TowerCombiner = new RawTowerCombiner("RawTowerCombiner_CEMC");
    TowerCombiner->Detector("CEMC");
    TowerCombiner->set_combine_eta(2);
    TowerCombiner->set_combine_phi(2);
//    TowerCombiner->Verbosity(RawTowerCombiner::VERBOSITY_SOME);
    se->registerSubsystem( TowerCombiner );
  }

 // TowerDigitizer
 ...
```

### DIFF
--- a/simulation/g4simulation/g4cemc/Makefile.am
+++ b/simulation/g4simulation/g4cemc/Makefile.am
@@ -57,6 +57,8 @@ libcemc_la_SOURCES = \
   RawClusterBuilderFwd_Dict.cc \
   RawTowerBuilder.cc \
   RawTowerBuilder_Dict.cc \
+  RawTowerCombiner.cc \
+  RawTowerCombiner_Dict.cc \
   RawTowerBuilderByHitIndex.cc \
   RawTowerBuilderByHitIndex_Dict.cc \
   RawTowerCalibration.cc \
--- a/simulation/g4simulation/g4cemc/RawTowerCalibration.cc
+++ b/simulation/g4simulation/g4cemc/RawTowerCalibration.cc
@@ -212,8 +212,7 @@ RawTowerCalibration::CreateNodes(PHCompositeNode *topNode)
       CaliTowerNodeName.c_str());
   if (!_calib_towers)
     {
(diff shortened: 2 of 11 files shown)


blackcathj:
Since this module is not used in default macro, merging it now to bring it to the nightly build.

PR #149, opened 2016-05-13
mlp - added the temperature-storing class and subsysreco class

mpurschke:
added the classes than stores the temperature readings on the DST.

### DIFF
--- a/offline/packages/Prototype2/Makefile.am
+++ b/offline/packages/Prototype2/Makefile.am
@@ -25,6 +25,8 @@ libPrototype2_la_SOURCES = \
   Prototype2DSTReaderDict.cc \
   CaloUnpackPRDF.C \
   CaloUnpackPRDFDict.C \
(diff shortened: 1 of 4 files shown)


blackcathj:
Hi, @mpurschke 

Thanks for the elegant solution for temperature decoding and storage. My test of the proposed branch works well for run 2609. Merging

I am reprocessing all test beam data with this temperature information with Mike Skoby set2 and set4 calibration. Output will replace the previous DST on the data disks: 
https://wiki.bnl.gov/sPHENIX/index.php/T-1044#Production_output 

Meanwhile, may I request you to write an example analysis code to read the time-interpolated temperature from DST in the analysis repository? It could be a new analysis module or a non-invasive update to Vera's module: 
https://github.com/sPHENIX-Collaboration/analysis/blob/master/Prototype2/EMCal/EMCalCalib_TestBeam/EMCalCalib_TestBeam.C 

Cheers

Jin

PR #150, opened 2016-05-14
mlp -- removed the "PROTOTYPE2_FEM.h" include from the header. 

mpurschke:
When you want to use it in a non-production environment, it fails since
we are not installing this header.
This will not affect the existing DSTs. After the next rebuild all will be fine.

### DIFF
--- a/offline/packages/Prototype2/RawTower_Temperature.h
+++ b/offline/packages/Prototype2/RawTower_Temperature.h
@@ -5,7 +5,6 @@
 #include <g4cemc/RawTowerDefs.h>
 #include <vector>
 #include <stdint.h>
-#include "PROTOTYPE2_FEM.h"
 
 class RawTower_Temperature : public RawTower {
  public:


blackcathj:
Good point. I need to fix that for RawTower_Prototype2 too.

pinkenburg:
please in the future put local  include files in quotes at the beginning 
of the list of includes, not at the end. We install the includes in 
package wise subdirs, this only works if we put local includes in quotes 
and add them before any with angle brackets


On 5/14/2016 3:49 PM, mpurschke wrote:
>
> When you want to use it in a non-production environment, it fails since
> we are not installing this header.
> This will not affect the existing DSTs. After the next rebuild all 
> will be fine.
>
> ------------------------------------------------------------------------
>
>
>         You can view, comment on, or merge this pull request online at:
>
> https://github.com/sPHENIX-Collaboration/coresoftware/pull/150
>
>
>         Commit Summary
>
>   * mlp -- removed the "PROTOTYPE2_FEM.h" include from the header. When
>
>
>         File Changes
>
>   * *M* offline/packages/Prototype2/RawTower_Temperature.h
>     <https://github.com/sPHENIX-Collaboration/coresoftware/pull/150/files#diff-0>
>     (1)
>
>
>         Patch Links:
>
>   * https://github.com/sPHENIX-Collaboration/coresoftware/pull/150.patch
>   * https://github.com/sPHENIX-Collaboration/coresoftware/pull/150.diff
>
> —
> You are receiving this because you are subscribed to this thread.
> <https://github.com/sPHENIX-Collaboration/coresoftware/pull/150>
>

-- 

Upton, NY 11973-5000

mpurschke:
Jin,

one more problem - we need to make the temperature correction in time
space but as far as I can tell, the event time ( the e->getTime() value)
isn't stored anywhere on the DST, unless I miss something. Similarly,
where is the event number?

I can get at the temp curve just fine, but it's useless if I don't know
the event time...

Best,
	Martin


On 5/14/16 15:55, Jin Huang wrote:
> Good point. I need to fix that for RawTower_Prototype2 too.
> 
> —
> You are receiving this because you authored the thread.
> <https://github.com/sPHENIX-Collaboration/coresoftware/pull/150#issuecomment-219248016>
> 


-- 
Martin L. Purschke, Ph.D.        ;   purschke@bnl.gov
                                 ;   http://www.phenix.bnl.gov/~purschke
                                 ;
Brookhaven National Laboratory   ;   phone: +1-631-344-5244
Physics Department Bldg 510 C    ;   fax:   +1-631-344-3253
Upton, NY 11973-5000             ;   skype: mpurschke
-----------------------------------------------------------------------

blackcathj:
Hmm. Both event number and time was not stored yet. Temperature correction probably is the first use of it. We probably need some object like DST/[EventHeader](https://www.phenix.bnl.gov/WWW/offline/doxygen/html/d8/db2/classEventHeader.html) to store the information. Both event header and its unpacker could be more widely used than just prototype 2 analysis. PHENIX used to have its unpacker at offline/framework/preco/HeadReco, we probably can consider place it at offline/packages/preco/HeadReco

PR #153, opened 2016-05-18
Tpc hijing take2

alandion:
Second attempt at running TPC+maps in high multiplicity

I committed changes to G4_Svtx_maps+tpc.C in the master branch.  This pull request was tested using that macro with 400 pions and it took 20 seconds/event.  Some more tweaks can speed it up further but this should get plots for the review with a big condor job.

Anyone else who wants to contribute to speeding things up : most of the time is in sPHENIXTracker::findTracks(...) .


mccumbermike:
We decided at the tracking meeting to come back with a different pull request to merge in a forked sPHENIXTracker object to bring the TPC tune into the build.

PR #165, opened 2016-06-21
Ph pythia6 update

johnlajoie:
This branch updates the PHPythia6 package by @nfeege with the following:

-> Added the ability to handle "mdme" settings in the config file (set particle decays)
-> Added triggers and the ability to handle a list of triggers in the event generation
-> Fixed passing the process ID to HepMC event record

All these are needed for the planned fsPHENIX DY simulations. This is an initial package to allow @nfeege and others to have a look and comment. 

As a check, setting the MDME flags in the config file as: 

mdme	174 1 0
mdme	175 1 0
mdme	176 1 0
mdme	177 1 0
mdme	178 1 0
mdme	179 1 0
mdme	180 1 0
mdme	181 1 0
mdme	182 1 1
mdme	183 1 0
mdme	184 1 0
mdme	185 1 0
mdme	186 1 0
mdme	187 1 0
mdme	188 1 0
mdme	189 1 0

results in the following output from a pylist(12):

       23     23    Z0                                  0    0    0     91.18800     2.47813    24.78129   0.00000E+00    1
           174    0   32    0.153995    d               dbar                                                            
           175    0   32    0.119420    u               ubar                                                            
           176    0   32    0.153984    s               sbar                                                            
           177    0   32    0.119259    c               cbar                                                            
           178    0   32    0.152272    b               bbar                                                            
           179    0   32    0.000000    t               tbar                                                            
           180   -1   32    0.000000    b'              b'bar                                                           
           181   -1   32    0.000000    t'              t'bar                                                           
           182    1    0    0.033576    e-              e+                                                              
           183    0    0    0.066806    nu_e            nu_ebar                                                         
           184    0    0    0.033576    mu-             mu+                                                             
           185    0    0    0.066806    nu_mu           nu_mubar                                                        
           186    0    0    0.033500    tau-            tau+                                                            
           187    0    0    0.066806    nu_tau          nu_taubar                                                       
           188   -1    0    0.000000    tau'-           tau'+                                                           
           189   -1    0    0.000000    nu'_tau         nu'_taubar                                                      

Which I think is correct - only decays to electrons will be generated. 

I have two things remaining that I want to do :

-> I need to write a quick module that will allow me to plot some distributions from the generated HepMC events. 

-> For every run, no matter the seed, I always get the message: 

vertex x/y/z6.22641/-2.54406/-4383.3 outside world volume radius/z (+-) 500/500, dropping it and its particles

which I tracked down as coming from HepMcNodeReader, I think on the first event. This seems to indicate some sort of problem with the HepMc node or how the initial event is added to the node.

### DIFF
--- a/generators/PHPythia6/Makefile.am
+++ b/generators/PHPythia6/Makefile.am
@@ -18,7 +18,11 @@ libPHPythia6_la_LIBADD = \
 
 libPHPythia6_la_SOURCES = \
   PHPythia6.C \
-  PHPythia6_Dict.C
+  PHPythia6_Dict.C \
+  PHPy6GenTrigger.C \
+  PHPy6GenTrigger.h \
+  PHPy6ForwardElectronTrig.C \
+  PHPy6ForwardElectronTrig.h 
 
 BUILT_SOURCES = \
   testexternals.C
@@ -38,6 +42,8 @@ testexternals.C:
 
 PHPythia6_Dict.C: \
   PHPythia6.h \
+  PHPy6GenTrigger.h \
+  PHPy6ForwardElectronTrig.h \
   PHPythia6LinkDef.h
 	rootcint -f $@ -c $(DEFAULT_INCLUDES) $(AM_CPPFLAGS) $^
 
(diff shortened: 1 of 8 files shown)


johnlajoie:
One more note - PHPythia6 also works properly when linked with a local pythia library that includes LHAPDF, another requirement for the fsPHENIX DY simulations.

PR #166, opened 2016-06-22
TPC Memory Reduction

mccumbermike:
The TPC was using fixed size arrays to cluster both the MAPS and the TPC layers. In the MAPS layers, due to the fine pitch pixels, this is a particularly wasteful approach. So I've rewritten both the MAPS and the TPC clusterizer to work over just those subsystems individually. The peak memory footprint is now 1.2 GB smaller and things will run on the normal condor queue for the TPC simulations.

I will need to push changes to the MAPS+TPC macro as well with this merge.

Here is the full MAPS performance cross check:
maps_memory

And here is the MAPS+TPC performance cross check:
tpc_memory

Some small changes in this one since the MAPS clusters are now made differently.


mccumbermike:
I'm going to merge this to get it into the new build ASAP for the TPC development crew.

PR #167, opened 2016-06-23
Adding PHG4DetectorSubsystem base class to handle parameters in one place

pinkenburg:
This new base class takes over the reading/writing/saving of PHG4Parameters. This way we can finally use a DB to store and retrieve configurations in the future. So far only the inner/outer hcal, and the block/cylinder subsystem has been ported in this pull request. It still needs an addition to the PHG4Parameter handling which is not able yet to store multiple detectors (e.g. more than one layer for the cylinders). This version stores those under nodes named <detector>_<layer> but this will be fixed in the next version. Running our sPHENIX macro gives identical results.
Also included some bugfixes: Reading back objects from files requires a default ctor for those objects which root uses to create it, those were missing in the raw tower geom classes. Also the identify method should write to the stream in the argument, not just cout.
Some new objects were added to the node dumper.


pinkenburg:
needs changes in our macros when merged

PR #168, opened 2016-06-23
Additional Pileup Changes

mccumbermike:
Some tools for pileup generation. I've added drift to the TPC for off-time particles. This should be the last pull request on pileup before I re-write the event generation workflow.


mccumbermike:
Chris, could you take a look at this merge conflict. I think the resolution will involve adding the blackhole times to the parameter set.

PR #176, opened 2016-07-13
added dielectron troigger

johnlajoie:
Added a dielectron trigger for fsPHENIX MB studies


johnlajoie:
Hi @pinkenburg - as far as the proliferation of trigger files, I was thinking about the same thing on the drive home and I plan to merge the the two triggers into one and make the number of candidates, the pT cut, and the eta cut configurable. I'll get that in tonight or tomorrow. 

As for the issue with the includes, mea clupa, I'm sloppy about this. I'll fix this as well and open a new pull request when it is all complete.

PR #177, opened 2016-07-13
Add event header for test beam analysis produciton

blackcathj:
From Test Beam workfest, adding event header to DST which store event number, type and timestampe. The immediate goals is to allow analysis code etrieve event time for use in temperature correction. 

Also the header installation location is updated to prototype2, e.g. inclusion of the EventHeader uses #include <prototype2/EventHeader_Prototype2.h>

## Example check

An example DST production from PRDF can be run using the default macro on RCF: 
https://github.com/sPHENIX-Collaboration/macros/blob/master/macros/prototype2/Fun4All_TestBeam.C 

From the DST output (./data/beam_00002609.root), the event header could be interactively printed via 
```
[jinhuang@rcas2073 macros]$ root
root [0]   gSystem->Load("libPrototype2.so");
root [2] TFile *_file0 = TFile::Open("data/beam_00002609.root")
root [3] T->Scan("DST.EventHeader.identify()")
************************
*    Row   * DST.Event *
************************
identify yourself: I am an EventHeader_Prototype2 Object
Event no: 2, Type: 1, RCDAQ arrival time: Wed Apr 27 19:21:33 2016
*        0 *         0 *
identify yourself: I am an EventHeader_Prototype2 Object
Event no: 3, Type: 1, RCDAQ arrival time: Wed Apr 27 19:21:33 2016
...
```

Or it can be plotted interactively:
```
T->Draw("DST.EventHeader.TimeStamp:DST.EventHeader.EvtSequence>>h2")
```

For formal analysis, a Fun4All analysis module is suggested. The event header can be fetched via DST node DST.EventHeader, like the standard PHENIX event headers.
```c++
#include <prototype2/EventHeader_Prototype2.h>
process_event(PHCompositeNode *topNode)
{
...
EventHeader_Prototype2 *eventheader = getClass<EventHeader_Prototype2> (topNode, "EventHeader");
...
}
```

### DIFF
--- a/offline/packages/Prototype2/EventHeader_Prototype2.C
+++ b/offline/packages/Prototype2/EventHeader_Prototype2.C
@@ -1,0 +1,38 @@
+#include "EventHeader_Prototype2.h"
+
+using namespace std;
+
+ClassImp(EventHeader_Prototype2)
+
+EventHeader_Prototype2::EventHeader_Prototype2()
+{
+  Reset();
+  return;
+}
+
+void 
+EventHeader_Prototype2::Reset()
+{
+  TimeStamp = 0;
+  EvtSequence = -999999;
+  EvtType = -999999;
+  return;
+}
+
+void 
+EventHeader_Prototype2::identify(ostream& out) const
+{
+  out << "identify yourself: I am an EventHeader_Prototype2 Object" << endl;
+  out << "Event no: " << EvtSequence 
+      << ", Type: " << EvtType 
+      << ", RCDAQ arrival time: " << ctime(&TimeStamp)
+      << endl;
+
+  return;
+}
+
+int 
+EventHeader_Prototype2::isValid() const
+{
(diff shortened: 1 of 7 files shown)


pinkenburg:
We should really use a base class for this. Given the extremely basic entries in this current implementation (event sequence, event type and timestamp) this shouldn't come as EventHeader_Prototype2 which inheritsdirectly from PHObject. Please change this to EventHeader.h with virtual methods to get those entries and then if you want to implement this as EventHeader_Prototype2. But the user code should use
EventHeader *eventheader = getClass<EventHeader> (topNode, "EventHeader");
which can be kept for the future even if we change our event header (which we most likely will)

blackcathj:
Hi @pinkenburg 

Thanks for the suggestion! 

I was trying to isolate the prototype code, to avoid letting this short-lifetime code become a long term burden in the main code base. But you have a good point to keep the analysis consistent. Since this is just a simple and generic event header, I am moving it to framework/ffaobjects  with addition of a base interface class following your suggestion.

PR #178, opened 2016-07-14
PHPythia6 merge electron triggers

johnlajoie:
Merge the single and di-electron trigger into one configurable trigger routine, fix includes.


blackcathj:
Hi @johnlajoie 

Thanks for the update! And thanks to @pinkenburg for his suggestions. 

I went through the patch and have few comments too. Via an another pull request to our development branch (https://github.com/johnlajoie/coresoftware/pull/4), I am suggesting following items:
1. use local include for PHPy6GenTrigger.h
2. install header to subfolder as in other sPHENIX packages
3. Include trigger class in the installed header, so later user could build their own test trigger class in separate libs without introducing many version of different trigger in core-software
Please consider merge or update them. 

Also please note that @mccumbermike is bringing a generic trigger mechanism to event generator as a base class. That would allow new generation of trigger classes used in many generators or their mixture. However I think we do not have to wait for that for the SpinFest simulation sample though. 

Thanks

Jin

PR #183, opened 2016-07-19
Geometry management from simulation to analysis

blackcathj:
## Introduction 

This is a preview for DST-based geometry management from simulation stage to analysis. It provide full geometry that is consistent with simulation for use in the analysis stage. Few example users are
* Digitization, to translate channel hit into space vectors 
* Kalman filter, to describe material distribution and measurement
* Alignment of detector
* Analysis-oriented event display

## Components

Basic component and data flow is illustrated here

image

### Run-time access of geometry

The run-time geometry can be found on DST tree on node ```RUN/GEOMETRY``` and presented by ```PHGeomTGeo```, which is a transient DST object that hold a pointer to current TGeoManager. TGeoManager contain the geometry and tools for use them. It already interfaces to GenFit2 Kalman filter and TEVE event display. And PHENIX FVTX detector use it to perform alignment too. 

TGeoManager currently is a global object in ROOT. Therefore PHGeomTGeo protects its pointer by locking gGeoManager and checks its validity before each operation. If ROOT decide to make TGeoManager non-global (they claim want to do so), these internal checks can be relaxed without change of interface. 

It is easier to use 
```c++
static  TGeoManager * PHGeomUtility::GetTGeoManager(PHCompositeNode *topNode) 
``` 
to fetch the TGeoManager from DST tree, which builds it if not yet initialized. 

### DST Storage of geometry

TGeoManager provides easy access of geometry, but it is very hard for DST storage due to manage due to many tentacles pointing and registering to global object. The solution is to introduce a storage object inspired by IO function supported by ROOT ```TGeoVolume::Import() and Export()```. 
1. During storage, the TOP volume in TGeoManager is streamed into a memory file. 
2. TFile was chosen for this operation to automatically add to storage of relevant TStreamInfo in order to support version evolution, something we learned in PHENIX FVTX geometry database storage. 
3. Then the binary data of ```TFile``` (via ```TMemFile```) is stored in persistent DST object PHGeomTGeo, which can be freely copied and streamed without worrying about global object.

The memory size of the geometry data of full sPHENIX detector is 2.1MB. The file size is compressed to 300kB, negligible compared with a nominal DST size. 

### Utilities

Utility class PHGeomUtility provides the functions illustrated the flow chart, without bothering user with the detail of operation on the DST objects. The functions are documented in code in doxygen style. 

Time for translation from Geant4 to DST for full sPHENIX detector is ~5s, with ROOT internal text output updating progress in GDML output and import into TGeoManager. And The DST file IO is very fast. 

### Example Fun4All macros

The G4 geometry translation and storage is enabled by default for sPHENIX simulation macro. One should find these new nodes on DST:
```
     RUN (PHCompositeNode)/
      ....
      GEOMETRY (PHDataNode)
      GEOMETRY_IO (PHIODataNode)
      ....
```

In addition, few example macro is provided to demonstrate import/export operations at 
coresoftware/offline/packages/PHGeometry/macros/

* Fun4All_ImportGeom.C imports a external geometry file (GDML, TGeoROOT file) onto DST
* Fun4All_ExportGeom.C export DST geometry into external geometry file  (GDML, TGeoROOT file, or .C macro)
* PHGeom_DSTInspection.C read in the DST file and illustrate the geometry in EVE event display. For example the default sPHENIX DST file will show up as
2016-07-18 23_56_01-nomachine - nx07
This macro is based on @sen1 's work.

## Plan

The plan is to do a joint test with Kalman filter to further validating these procedures with @HaiwangYu . Then we will request merge. An addition of ```coresoftware/offline/packages/PHGeometry```to the build package list is also required at the merge.


blackcathj:
From action item of last simulation meeting, this is to confirm only Geometry object are exported from Geant4 to DST via GDML tmp file. Geant tracks and hits are not exported. 

Few more minor commit to be appended to this pull request.

PR #184, opened 2016-07-21
Bug fix on local coordinate calculation

blackcathj:
This pull request introduce a bug fix for local coordinate calculation utility.

## What's fixed

Local coordinate calculation was introduced in #173 to the base class of ```PHG4SteppingAction```. Tony @adfrawley  reported the exit point calculation is wrong for the on-going MAPS detector development. This problem was traced down to the translation matrix for the exit point of PHG4Hit used the next logical volume passing the sensitive volume. Fix was straight forward: use the pre-step's translation matrix for glocal-to-local transformation of both pre- and post-points. 

## Verification

Currently two subsystem uses this local coordinates translation. Both are checked:

### Scintillation model in SPACAL fiber 

Local x-y coordinate of exit point is plotted here for PHG4Hits in SPACAL fibers in one electron shower. It shows a nice circular shell for at the radius of fiber core. This check did not reveal this bug before the fix because the next volume pass the fiber core is fiber cladding, which share the identical glocal-to-local transform matrix as the core.

2016-07-20 23_54_03-nomachine - nx07

### MAPS digitization in sensor

@adfrawley uses local coordinate in MAPS sensor to determine pixel hit. After this fix, the entry (black) and exit (red) points show up at the front and back surface of the sensor volume (y = +/- 10 um), except for very few secondary production in the sensor. The local x-y location nicely cover the surface of a sensor, which will be converted to pixel number. 

2016-07-21 10_40_44-nomachine - nx07

### DIFF
--- a/simulation/g4simulation/g4main/PHG4SteppingAction.cc
+++ b/simulation/g4simulation/g4main/PHG4SteppingAction.cc
@@ -122,11 +122,12 @@ PHG4SteppingAction::StoreLocalCoorindate(PHG4Hit * hit, const G4Step* aStep,
   assert(hit);
   assert(aStep);
 
+  G4StepPoint* preStepPoint = aStep->GetPreStepPoint();
+  G4TouchableHandle theTouchable = preStepPoint->GetTouchableHandle();
+
   if (do_prepoint)
     {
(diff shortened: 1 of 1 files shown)


blackcathj:
Merging for @adfrawley 's tests

PR #187, opened 2016-07-27
Constant cluster errors in HelixHough

mccumbermike:
I've replaced the variable cluster errors in the HelixHough with calculations from a constant covariance matrix. For the TPC the fixed errors are now generated in the clusterizer and not the pattern reco. I expect conflicts with Carlos's pull request. I will take a look at that now, and sort the conflicts out on my side. Here is a look at the full MAPS tracker:
maps

And a look at the MAPS+TPC tracker:
tpc

Some small rounding differences (covars take a sqrt()) and a patch here and there invalidate a pure diff on the dumps, but these look acceptably the same.


mccumbermike:
Merging... TPC performance after the conflict resolution in sPHENIXTrackerTPC:
tpc

PR #188, opened 2016-07-28
Patch overlapcheck

nfeege:
Passing parameter 'overlapcheck' down the detector construction chain for detector construction classes that are not derived from PHG4Detector.


blackcathj:
Thanks for patching it up!

PR #195, opened 2016-09-20
mlp - hopefully put the order of temperature sensors right.

mpurschke:
hopefully put the order of temperature sensors right. Module 0 has sensor 63, but on top of that position x,y = (0,0) is module 7. We changed the view to look downstream, not upstream as one
normally does, after the initial commit.


offline/packages/Prototype2/TempInfoUnpackPRDF.C
```diff
@@ -168,7 +168,9 @@ int  TempInfoUnpackPRDF::addPacketInfo(Packet *p, PHCompositeNode *topNode, cons
 		  tower = new RawTower_Temperature();
 		  emcal_temperature->AddTower(ibinz,ibinphi,tower);
 		}
-	      tower->add_entry( evtnr, etime, p->iValue(ibinz*PROTOTYPE2_FEM::NCH_EMCAL_COLUMNS + ibinphi)/1000. ); 
+	      // this takes care of the newly found "reverse" mapping. (0,0) is module 7, (0,7) is module 0, and 
+	      // the 63 - (...) takes care of the reversed vector.  
+	      tower->add_entry( evtnr, etime, p->iValue( 63- (ibinz*PROTOTYPE2_FEM::NCH_EMCAL_COLUMNS + (7-ibinphi) ) /1000. ) );
```
blackcathj: Hi, @mpurschke 

Thanks for the update! 

May I confirm the naming conversion is that
- Columns -> Eta bins = z bins
- Rows -> phi bins

I am asking because in earlier lines it iterates ```ibinz``` and ```ibinphi``` as 
```c++
 for(int ibinz=0; ibinz<PROTOTYPE2_FEM::NCH_EMCAL_ROWS; ibinz++)
 for(int ibinphi=0; ibinphi<PROTOTYPE2_FEM::NCH_EMCAL_COLUMNS; ibinphi++)
```

Jin

PR #202, opened 2016-10-24
add hit flag for shower studies in prototype

johnlajoie:
This patch adds a hit_type property to PHG4Hit and makes use of it in the inner and outer HCAL prototype stepping actions. This allows one to flag hits that have an EM (e+/e-) or hadronic (all others) origin very easily at the hist stage to facilitate studies of the HCAL signal shapes and the correlations with fluctuations in the EM fraction of the showers.

### DIFF
--- a/simulation/g4simulation/g4detectors/PHG4Prototype2InnerHcalSteppingAction.cc
+++ b/simulation/g4simulation/g4detectors/PHG4Prototype2InnerHcalSteppingAction.cc
@@ -190,6 +190,12 @@ bool PHG4Prototype2InnerHcalSteppingAction::UserSteppingAction( const G4Step* aS
 	  //set the initial energy deposit
 	  hit->set_edep(0);
 	  hit->set_eion(0); // only implemented for v5 otherwise empty
+
+          hit->set_hit_type(0);
+          if( (aTrack->GetParticleDefinition()->GetParticleName().find("e+") != string::npos) ||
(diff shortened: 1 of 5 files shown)


blackcathj:
Hi, John

Thanks for the update. Merging it. 

Meanwhile, to trace the e/h difference on ```PHG4Hit``` for all other detectors, we can always use the truth information container in an analysis module. One example is trace EM part of visible energy deposition in  [```QAG4SimulationCalorimeter```](https://www.phenix.bnl.gov/WWW/sPHENIX/doxygen/html/d9/d7b/classQAG4SimulationCalorimeter.html#a5b70cca03633b9708a6aea0b57e21272) . Specifically: 


```c++
int
QAG4SimulationCalorimeter::process_event_G4Hit(PHCompositeNode *topNode)
{
  double ev_calo_em = 0.0; // EM visible energy
      PHG4HitContainer::ConstRange calo_hit_range =
          _calo_hit_container->getHits();
      for (PHG4HitContainer::ConstIterator hit_iter = calo_hit_range.first;
          hit_iter != calo_hit_range.second; hit_iter++)
        {

          // EM visible energy that is only associated with electron energy deposition
          PHG4Particle* particle = _truth_container->GetParticle(
              this_hit->get_trkid());
          if (abs(particle->get_pid()) == 11)
            ev_calo_em += this_hit->get_light_yield();
        }

}
```

PR #203, opened 2016-10-24
Phhepmc event selector

SashaLebedev:
Particle selector class (and base class) for selecting particles for output 
in PHHepMC-based events.


blackcathj:
Hi, Sasha

Thanks for the pull request, which introduce an useful filtering functionality for fast G4 simulations. Few question or suggestions:

1.	It appears PHHepMCParticleSelectorBase is not doing much rather PHHepMCParticleSelectorB2Jpsi implements its functionalities. In that case, shall we rename PHHepMCParticleSelectorB2Jpsi as PHHepMCParticleSelectorDecayProduct. Then from macro, user could reconfig it to trigger on B->J/Psi using set functions. 
2.	It takes me many guess on what is ```_theParticle, _theDaughters, _theParents``` and ```_theTrigger```. Could you add a few comment lines in front of the set function and variable declaration to describe them? Please start the comment with 
```c++
/// comments
std::vector<int> _theDaughters;
```
Then it will be formatted in [the doxygen webpage](https://www.phenix.bnl.gov/WWW/sPHENIX/doxygen/html/index.html). More on example of comment documentation of code onto the doxygen page is [here](http://www.stack.nl/~dimitri/doxygen/manual/docblocks.html#docexamples).

If you could work on these updates, commit and upload to your _fork_ again, the new changes will be automatically appended to this pull request. 

Cheers

Jin

sPHENIX-GitHub-l:
Hi Jin,

PHHepMCParticleSelectorB2Jpsi has some B->J/psi specific
features, for example it takes into account that B can decay
to J/psi via B -> chi_c ->J/psi, so I would like to keep the name.

I can add another more general class named PHHepMCParticleSelectorDecayPr
oduct
if you want.

I will add doxygen comments and commit again.

Sasha.

SashaLebedev:
Now just one class does all the work, single particles, sets of particles, decay chains,
all I can imagine.
Use Fun4All return codes implemented

PR #207, opened 2016-11-04
Enable tracking for all tracks, not just primaries

johnlajoie:
Submitted for consideration, although I am sure this is not the last word.  Currently PHG4TrackFastSim only checks primary tracks for fast tracking. This modification leaves primary tracks as the default, but enables a switch to search all tracks  (set in G4_FGEM_fsPHENIX.C:FGEM_FastSim_Reco): 

  kalman->set_primary_tracking(0); // 0=all tracks, >0 = primaries only

I could use some advice here - it seems that PHG4TruthInfoContainer makes some distinction between tracks and showers, so this isn't horribly inefficient in the sense that it searches over all tracks in an EM shower, but I am uncertain how the truth tracking is done so this interpretation may not be correct. Another way to do this would be to just combine hits in the FGEM's from the same tracks, but that is more complicated and potentially biased if you are not careful. 

I ran tests with this using single e-, pi- as well as DY and MB events and although the time per event did not increase by huge factors, although I didn't time it closely.  The output is seeing tracks found that do not match up to primaries in my evaluator code. 

Finally, when running over all tracks there are more messages in the err files like: 

Error in <TDecompChol::Decompose()>: matrix not positive definite
genfit::Exception thrown with excString:
KalmanFitterInfo::calcAverageState: ill-conditioned covariance matrix.
in line: 149 in file: /phenix/u/phnxbld/workarea/sPHENIX/genfit/genfit/core/src/MeasuredStateOnPlane.cc
with fatal flag 0

or

genfit::Exception thrown with excString:
MaterialEffects::getMomGammaBeta - Energy <= mass
in line: 362 in file: /phenix/u/phnxbld/workarea/sPHENIX/genfit/genfit/trackReps/src/MaterialEffects.cc
with fatal flag 0

which I think makes perfect sense - we are now handing GenFit some pretty ragged tracks from secondaries, and things like the initial momenta and particle type choice may not be valid.

### DIFF
--- a/simulation/g4simulation/g4hough/PHG4TrackFastSim.C
+++ b/simulation/g4simulation/g4hough/PHG4TrackFastSim.C
@@ -66,7 +66,7 @@ PHG4TrackFastSim::PHG4TrackFastSim(const std::string &name) :
 				false), _use_vertex_in_fitting(true), _vertex_xy_resolution(
 				50E-4), _vertex_z_resolution(50E-4), _phi_resolution(50E-4), _r_resolution(
 				1.), _z_resolution(50E-4), _pat_rec_hit_finding_eff(1.), _pat_rec_noise_prob(0.), 
-		                _N_DETECTOR_LAYER(5), _N_STATES(0) {
+		                _N_DETECTOR_LAYER(5), _primary_tracking(1), _N_STATES(0) {
 
 	_event = -1;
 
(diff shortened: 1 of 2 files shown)


HaiwangYu:
Hi John, by default the PHG4TrackFastSim tracker uses a smeared primary vertex. I think those genfit::Exceptions are caused by we force a non-prompt track to fit with a primary vertex. When I run your PR code with the set_use_vertex_in_fitting(false) in the macro. The genfit::Exceptions are largely suppressed.

Also as mentioned in your test, the tracking time consumption seems to be still reasonable compared with other Geant simulation parts with this simple modification. So I think we could merge this now and improve the efficiency later.

blackcathj:
Thanks @HaiwangYu for the cross check. Merging both pull request and update GenFit2 error handling in future pull requests. 

https://github.com/sPHENIX-Collaboration/macros/pull/37

PR #208, opened 2016-11-04
Svtx evaluator

HaiwangYu:
Fix a typo(?) of the SvtxEvaluator:
Instead of using the incoming Geant hit position, use the average position for the PHG4Hits related evaluations.


mccumbermike:
At one point there was a discussion about removing the outgoing points from
the g4hits on both the tracking and the calorimetry. If that isn't the
plan---and I'm guessing it isn't---there shouldn't be any harm in using the
average if it improves the evaluation.

Mike

--
Michael P. McCumber, PhD
505-709-8161

blackcathj:
Thanks @mccumbermike for the confirmation. Merging.

PR #210, opened 2016-11-10
Keep track of number of hits for fast sim track

johnlajoie:
Submitted for your consideration - it would be very useful to keep track of *at least* the number of hits on a FastSim track, especially when looking at all tracks and not just primary tracks.  The evaluation software from Haiwang uses the "size_clusters()" method to extract the number of hits on the track, but in SvtxTrack this is meant to keep track of the number of calorimeter clusters attached to the track, and in any case PHG4TrackFastSim does nothing to fill it. 

The attached patch just adds the ability to store the number of measurements along the track in SvtxTrack_FastSim and adds the code to PHG4TrackFastSim to initialize it.  

It might be nice to store an id/pointer to the detector-specific hits instead, but this should be done in a more general way that will ultimately interface with the TPC, etc.  I thought about this but decided I don't really know enough to do this in a general way at this point.

### DIFF
--- a/generators/PHPythia6/PHPy6ForwardElectronTrig.C
+++ b/generators/PHPythia6/PHPy6ForwardElectronTrig.C
@@ -20,7 +20,7 @@ PHPy6ForwardElectronTrig::PHPy6ForwardElectronTrig(const std::string &name): PHP
   n_em_required = 1; 
   n_ep_required = 1; 
   n_comb_required = 1; 
-  ptot_required = 1.0; 
+  pt_required = 0.5; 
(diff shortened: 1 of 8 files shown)


blackcathj:
Thanks @johnlajoie . One minor comment: please init ```SvtxTrack_FastSim::_nmeas``` in the constructor ```SvtxTrack_FastSim::SvtxTrack_FastSim()```.

johnlajoie:
Thanks @blackcathj  - I'm embarrassed I neglected that...

PR #211, opened 2016-11-11
Simplification of hit saving in stepping actions

pinkenburg:
So far the steps were accumulated inside the stepping action for each tack but the hit was only saved when the next track was handled so zero energy hits could be rejected (those hits allocated too much memory previously). I finally read up on how to detect the last step inside a volume and hits are now saved during the last step if the energy deposition is not zero (geantinos are marked by negative energy deposition). This makes the code a lot simpler and it can easily be extended to other stepping actions where memory consumption due to saving of zero energy hits was bearable.
The G4Hit class has now a Reset() method which clears the hit, also the identify method now prints out all properties of a given hit, not just a selection.


pinkenburg:
One more change - the number of scintillators per tower for the hcals is now a parameter. This can be then propagated to the cell reconstruction instead of having to set this correctly in the reconstruction macro

pinkenburg:
Needed to change other steppingactions (the forward detectors bomb out when running hijing - this might actually help)

PR #216, opened 2016-11-15
carry HepMC barcode information through truth tracking

johnlajoie:
This set of modifications will carry the HepMC barcode through to the truth information for primary particles, allowing you to connect a PHG4Particle from PHG4TruthInfoContainer back to the HepMC event record, which then allows you to track back the true origin of a primary in the HepMC event record. For example: 

`         PHHepMCGenEvent *genevent = findNode::getClass<PHHepMCGenEvent>(topNode,"PHHepMCGenEvent");
          HepMC::GenEvent *evt = genevent->getEvent();

	  PHG4TruthInfoContainer::ConstRange range =
			_truth_container->GetParticleRange();

	  for (PHG4TruthInfoContainer::ConstIterator truth_itr = range.first;
			truth_itr != range.second; ++truth_itr) {

	    PHG4Particle* g4particle = truth_itr->second;

            if(_truth_container->is_primary(g4particle)){
	      HepMC::GenParticle *ptcle = evt->barcode_to_particle(g4particle->get_barcode()); 
              ....
	    }
            ...
          }
`

The cost of doing this is the addition of the barcode integer to PHG4Particle and the addition of PHG4UserPrimaryParticleInformation to every primary G4 track to allow the barcode information to carry forward; previously this was only added to embedded tracks.

### DIFF
--- a/simulation/g4simulation/g4main/HepMCNodeReader.cc
+++ b/simulation/g4simulation/g4main/HepMCNodeReader.cc
@@ -199,6 +199,7 @@ int HepMCNodeReader::process_event(PHCompositeNode *topNode) {
         particle->set_px((*fiter)->momentum().px() * mom_factor);
         particle->set_py((*fiter)->momentum().py() * mom_factor);
         particle->set_pz((*fiter)->momentum().pz() * mom_factor);
+	particle->set_barcode((*fiter)->barcode()); 
         ineve->AddParticle((*v)->barcode(), particle);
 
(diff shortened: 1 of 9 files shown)


mccumbermike:
Looks good. The overhead on passing things forward is a lot less than it was assigning things by trace at the end, so there shouldn't be an impact on our performance--didn't see any in my exercise of the branch. Merging...

Issue #232, opened 2017-01-04
PHG4CylinderGeom_MAPS class lacks public constructor


pinkenburg:
I got this message when running the new maps+itt+tpc macro:
Warning in <TBufferFile::WriteObjectAny>: since PHG4CylinderGeom_MAPS has no public constructor
	which can be called without argument, objects of this class
	can not be read with the current library. You will need to
	add a default constructor before attempting to read it.

root calls the default ctor without arguments when reading objects back from file. One option to fix this  would be to give the existing ctor default arguments.

mccumbermike:
I would also add to Chris's suggestion that I've found good practice in these cases to be hiding the default constructor within a protected block so that users are still directed to declare their arguments explicitly.

PR #235, opened 2017-01-06
Two fixes in silicon simulation code

blackcathj:
Two fixes in silicon simulation code
1. As raised by @pinkenburg and @mccumbermike in #232 , ```PHG4CylinderGeom_MAPS``` does not have a public accessible default constructor, which is required for ROOT file IO via the streamer utilities. Adding it. 
Note if the default constructor is put under ```protected``` flag, ROOT still complains on the missing default constructor.  
2. @HaiwangYu raised an issue that the INTT ladder header file ```PHG4CylinderGeom_Siladders.h``` , which is required for downstream analysis. Adding it.


blackcathj:
Have to merge it now as they are being used in the MAPS workfest on silicon ladder developments.

PR #240, opened 2017-01-19
Add PHG4BeamlineMagnetSubsystem

nfeege:
Adding a new class g4detectors/PHG4BeamlineMagnetSubsystem. This class represents beam line magnet for studies that need to consider the extended interaction region, for example exclusive processes in ep and eA collisions. The magnets are represented as cylinders with constant dipole or constant gradient quadrupole field within the cylinder volume. The magnet field, field type, position, aperture, outer radius, as well as position and rotation w.r.t. the nominal beam axis can be set via parameters.

This is a screenshot of five magnets in Fun4All (together with the BaBar coil from the default Fun4All macros):

screen shot 2017-01-18 at 9 14 58 pm

A rough crosscheck of the deflection of particles in 'x'-direction at z=30 m caused by these magnets agrees reasonably well with calculations by CAD/MagnetDivision for the same configuration of eRHIC magnets:

particle  p(GeV)          angle(mrad)       z(mm)           x_CAD(mm)       x_Geant4(mm)
neutron   250        +4              30000           120.8                      120.0
proton    250        0               30000           279.2                      275.5
proton    250        +5              30000           336.8                      329.8
proton    250*0.8    0               30000           341.6                      333.7

Refining the simulation in the future (e.g. correcting the aperture which is actually not a perfect cylinder) are expected to improve the agreement.

### DIFF
--- a/simulation/g4simulation/g4detectors/Makefile.am
+++ b/simulation/g4simulation/g4detectors/Makefile.am
@@ -131,6 +131,9 @@ libg4detectors_io_la_SOURCES = \
   PHG4ScintillatorSlatContainer_Dict.cc
 
 libg4detectors_la_SOURCES = \
+  PHG4BeamlineMagnetDetector.cc \
+  PHG4BeamlineMagnetSubsystem.cc \
+  PHG4BeamlineMagnetSubsystem_Dict.cc \
   PHG4BlockCellReco.cc \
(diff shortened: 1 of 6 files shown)


blackcathj:
Thanks for the useful addition of beam line magnetic fields!

PR #244, opened 2017-02-23
add PDF info to to HepMC::GenEvent

johnlajoie:
Adds a PdfInfo structure to the HEPMC GenEvent so the x1, x2, Q2 and parton id's for the event are saved.


blackcathj:
Great to have that information. Thanks!

PR #247, opened 2017-02-27
Memory leak fix for PHGenFit and PHG4TrackKalmanFitter

HaiwangYu:
Fix memory leaks related to PHGenFit and PHG4TrackKalmanFitter.

After this fixing, with DST output turned off, there will be no memory leaks in `PHG4TrackKalmanFitter::process_event`

valgrind log file for one full pp event refitting with DST output turned off; cppcheck log file for PHGenFit and g4hough could be find here:
https://www.phenix.bnl.gov/WWW/p/draft/yuhw/PullRequest/2017-02-27/0/

There are still some one time memory leaks related with TGeo and loading RAVE libraries. But as they are one time leaks, we should be fine for now.

### DIFF
--- a/offline/packages/PHGenFitPkg/PHGenFit/Track.cc
+++ b/offline/packages/PHGenFitPkg/PHGenFit/Track.cc
@@ -45,7 +45,7 @@ Track::Track(genfit::AbsTrackRep *rep, TVector3 seed_pos, TVector3 seed_mom, TMa
 	//_track = NEW(genfit::Track)(rep, seedState, seedCov);
 }
 
-int Track::addMeasurements(std::vector<PHGenFit::Measurement*> measurements)
+int Track::addMeasurements(std::vector<PHGenFit::Measurement*> &measurements)
 {
 	BOOST_FOREACH(PHGenFit::Measurement* measurement, measurements)
 	{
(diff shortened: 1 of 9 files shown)


HaiwangYu:
As suggested by Jin. Switching to use exception safe smart pointers.
Updated valgrind log files:
https://www.phenix.bnl.gov/WWW/p/draft/yuhw/PullRequest/2017-02-27/1/

offline/packages/PHGenFitPkg/PHGenFit/Track.cc
```diff
@@ -211,6 +226,8 @@ double Track::extrapolateToPoint(genfit::MeasuredStateOnPlane& state, TVector3 P
 
 	state = *dynamic_cast<genfit::MeasuredStateOnPlane*> (kfsop);
 
+	delete kfsop;
```
blackcathj: Hi @HaiwangYu , Thanks for the update! 

Would you like to try std::unique_ptr for these four occurrence of ```kfsop``` too?

HaiwangYu: @blackcathj Thanks for the comments. I updated the PR branch accordingly.

PR #248, opened 2017-02-27
A few more memory leak fixes for PHG4TrackFastSim

johnlajoie:
Clean up the gf_state allocation when done creating the SVXTrack.

### DIFF
--- a/simulation/g4simulation/g4hough/PHG4TrackFastSim.C
+++ b/simulation/g4simulation/g4hough/PHG4TrackFastSim.C
@@ -493,7 +493,7 @@ SvtxTrack* PHG4TrackFastSim::MakeSvtxTrack(const PHGenFit::Track* phgf_track,
 
 	double pathlenth_from_first_meas = -999999;
 	double pathlenth_orig_from_first_meas = -999999;
-	genfit::MeasuredStateOnPlane* gf_state = new genfit::MeasuredStateOnPlane();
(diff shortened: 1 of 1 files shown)


simulation/g4simulation/g4hough/PHG4TrackFastSim.C
```diff
@@ -597,6 +599,8 @@ SvtxTrack* PHG4TrackFastSim::MakeSvtxTrack(const PHGenFit::Track* phgf_track,
 	  out_track->insert_state(state);
 	}
 
+	delete gf_state;
```
blackcathj: Hi, @johnlajoie 

Thanks for the follow up. 

One comment: it would be more convenient and exception-safe if we use 
```c++
unique_ptr<genfit::MeasuredStateOnPlane> gf_state(new genfit::MeasuredStateOnPlane());
```
instead of 
```c++
genfit::MeasuredStateOnPlane* gf_state = new genfit::MeasuredStateOnPlane();
//....
delete gf_state; 
```
Then, ```unique_ptr``` will delete ```gf_state``` automatically at exit of the function.

johnlajoie:
Thanks @blackcathj !  It's a great Monday when you learn something right at the start of the day. I was unfamiliar with the unique_ptr construction - I'm going to make use of this a lot in the future.

blackcathj:
Thanks for the quick response! Merged

PR #251, opened 2017-03-08
HDI implementation is updated according to the recent design. strip_z…

gmitsuka:
…_index and strip_y_index are assigned with more precise calculation.


blackcathj:
Thanks for the fix!

PR #252, opened 2017-03-09
bug fixed on the INTT ladder z position; segment_z_bin<=1 means negat…

gmitsuka:
…ive rapidity, thus the |z| must be multiplied by signz=-1.
I forgot to commit this code yesterday.


HaiwangYu:
Hi Gaku,

Thanks a lot for the update.
I tested it with single pion simulation.
The cluster rphi and z pulls are showing below:
image
image

It seems the z pull looks much better now (maybe update the cluster uncertainties a little bit).
While the rphi pull is still off 0.

We could discuss this more tomorrow.

blackcathj:
Hi, @gmitsuka Thanks for contributing the z-fix. Looking forward to pinging down the problem and fixing in the phi-dimension bias too.

PR #256, opened 2017-03-17
PHG4CylinderCell replacement

pinkenburg:
this class has outlived its usefulness. The replacement PHG4Cell is formed like the PHG4Hits with properties rather than hardcoded variables which lead to an explosion of versions. The key size increased from 32 to 64 bits to encode layers/detector ids, the binning scheme and 32 bits for bins (eta/phi, size, slats,...). The maps still need a more reasonable lookup since getting phi/eta from the properties is more expensive than decoding the key. Overall the code is slightly faster (dealing with the maps will give another speed boost) and the dsts are smaller. The resulting tracks/clusters did not change.
This change is for the central barrel only - fs/ePHENIX are still to do. If this is merged the macros also need to be merged

I made a play build with this code for testing, just fetch the updated G4_CEmc_Spacal.C (or comment out the tmin/tmax setting - the defaults in the new code are set to 0/60)


pinkenburg:
Getting this in before proceeding with actual stepping action bugfixes

PR #260, opened 2017-03-24
Intt cluster fix

adfrawley:
Modified PHG4SiliconTrackerSteppingAction to work around a bug in the reporting of the volume for the first pre-step point in a multi-strip hit.

Fixed a bug in PHG4SvtxClusterizer that was preventing clustering for the INTT strips.

Cluster errors after the fix:

[intt_cluster_errors.pdf](https://github.com/sPHENIX-Collaboration/coresoftware/files/866737/intt_cluster_errors.pdf)

More details:

[INTT_clustering_fix_march23_2017.pdf](https://github.com/sPHENIX-Collaboration/coresoftware/files/866736/INTT_clustering_fix_march23_2017.pdf)

### DIFF
--- a/simulation/g4simulation/g4detectors/PHG4SiliconTrackerCellReco.cc
+++ b/simulation/g4simulation/g4detectors/PHG4SiliconTrackerCellReco.cc
@@ -195,6 +195,8 @@ int PHG4SiliconTrackerCellReco::process_event(PHCompositeNode *topNode)
       if (verbosity > 0)
         {
           std::cout //<< "Adding cell for sensor_index: " << mapiter->second->get_sensor_index()
+          << ", ladder z index: " << mapiter->second->get_ladder_z_index()
+          << ", ladder phi index: " << mapiter->second->get_ladder_phi_index()
           << ", srip z index: " << mapiter->second->get_zbin()
           << ", strip y index: " << mapiter->second->get_phibin()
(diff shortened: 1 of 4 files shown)


pinkenburg:
I have concerns about the use of the IsFirstStepInVolume() and IsLastStepInVolume()  methods:
if( aStep->IsFirstStepInVolume() == 1 &&  aStep->IsLastStepInVolume() == 1)
the current G4Step.hh include file marks them as being unused:
  // Get/Set/Clear flag for initial/last step
   // NOTE:  following flags are not used 
   //        will be ready in later release
   G4bool IsFirstStepInVolume() const;
   G4bool IsLastStepInVolume() const;

   void SetFirstStepFlag();
   void ClearFirstStepFlag();
   void SetLastStepFlag();
   void ClearLastStepFlag();

My tests show that the IsFirstStepInVolume() seems to work but is redundant and set whenever the prestep status is fGeomBoundary or fUndefined which is exactly what the current stepping actions check for when creating a new hit while the  IsLastStepInVolume() only works when the partticle leaves the volume - equivalent to post step status fGeomBoundary. When a track stops in a volume  IsLastStepInVolume() is false which is certainly wrong. Until this is better understood (or G4 removes the comment about this being used only in the future we should not use  IsFirstStepInVolume() or IsLastStepInVolume() but stick to the pre/post step status and the track status (fStopAndKill indicates the last step for a given track)

PR #261, opened 2017-03-29
change PHG4TrackKalmanFitter to use PHG4Cells

pinkenburg:
it compiles but I cannot test it. But since the new PHG4Cells and the old PHG4CylinderCells have the same interfaces it should work
The change in the HepMCNodeReader makes the random seed accessible from the outside so crashes where the seed is known can be reproduced


HaiwangYu:
Thanks Chris, we will test it.

PR #263, opened 2017-03-31
improving TPC reco and clusterizer

carlosperezlara:
TPCreco now fixed to use accurate formulation for diffusion, better use of CPU time too
TPCclusterizer compute errors of clusters now, better use of CPU time too


carlosperezlara:
call from macros a bit different too: more parameters are passed. New loading macro is being prepared  to handle this plus new configurations for MAPS, INTT and tracker.

vcanoa:
[compare_old_vs_newcluster.pdf](https://github.com/sPHENIX-Collaboration/coresoftware/files/909054/compare_old_vs_newcluster.pdf)

PR #274, opened 2017-04-24
Initial commit of calo trigger emulator and info node.

dvperepelitsa:
Initial commit of calo trigger related code. 

There are two major components: (1) a calo trigger simulator module meant to emulate the performance of the Level-1 trigger to be used in sPHENIX data-taking, and (2) a lightweight node which stores calo-trigger information and is to be written to the node tree in simulations.

At the moment, only a photon trigger (or single-electron trigger for Upsilon triggering) is implemented which consists of a 4x4 sliding window (composed of four 2x2 windows) algorithm on the EMCal towers. The simulator is agnostic as to the EMCal geometry (e.g. 1D- or 2D-projective, which differ in their eta structure) and will set up the internal tower map according to what the geometry container object tells it.

This is intended to be the first commit just to get some basic functionality into the framework -- additional functionality (N-bit ADC truncation, jet patch trigger, UE subtraction at Level-1 for Au+Au triggering, potentially di-electron trigger with mass hypothesis, etc.) will follow. I will present a short overview of the module and some first results in the next Simulations meeting.

### DIFF
--- a/offline/packages/trigger/CaloTriggerInfo.C
+++ b/offline/packages/trigger/CaloTriggerInfo.C
@@ -1,0 +1,21 @@
+#include "CaloTriggerInfo.h"
+
+ClassImp(CaloTriggerInfo)
+
+CaloTriggerInfo::CaloTriggerInfo()
+{
+  _EMCAL_4x4_BEST_E = 0;
+  _EMCAL_4x4_BEST_ETA = 0;
+  _EMCAL_4x4_BEST_PHI = 0;
+}
+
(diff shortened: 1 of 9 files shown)


blackcathj:
Made minor clean up of include file in ```CaloTriggerInfo.h```. Passed cppcheck and compile. Merging.

pinkenburg:
We should really add a virtual base class. It's not much work but will keep things readable if things are added/removed. There will be changes and given that our productions do take time and stay around for a while this should be done before this is merged

PR #276, opened 2017-04-24
PHG4TrackFastSim leaks like a sieve!

johnlajoie:
PHG4TrackFastSim was leaking like crazy, making it impossible to run forward jet simulations.  Several of these were simple, just neglecting to delete things that were unused or that a subsequent object makes a copy of.  The main problem, however, has to do with the PHGenFit interface to GenFit.  PHGenFit creates a number of interface objects that themselves create underlying GenFit objects. The top level of this is the PHEGenFit::Track object that owns its measurements in a parallel fashion to the way the GenFit::Track object owns its measurements.  However, the code was keeping track of the underlying GenFit::Track objects and deleting them.  This left all the PHGenFit wrapper objects lying around consuming memory like crazy.

Valgrind is happy now, and doesn't show PHG4FastTrackSim as responsible for leaking anything.  There are a lot of jumps on uninitialized variables in GenFit itself; however. That will have to be an investigation for another day

### DIFF
--- a/simulation/g4simulation/g4hough/PHG4TrackFastSim.C
+++ b/simulation/g4simulation/g4hough/PHG4TrackFastSim.C
@@ -318,9 +318,9 @@ int PHG4TrackFastSim::process_event(PHCompositeNode *topNode) {
 	  _fitter->getEventDisplay()->addEvent(rf_gf_tracks);
 	}
 	else{
-	  //for (std::vector<PHGenFit::Track*>::iterator it = rf_tracks.begin() ; it != rf_tracks.end(); ++it)
-	  //  delete (*it);
-	  //rf_tracks.clear();
+	  for (std::vector<PHGenFit::Track*>::iterator it = rf_tracks.begin() ; it != rf_tracks.end(); ++it)
+	    delete (*it);
+	  rf_tracks.clear();
 	}
 
 //	if(_trackmap_out->get(0)) {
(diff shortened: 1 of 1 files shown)


HaiwangYu:
Hi John, thanks for the fix.

I fixed memory leaks in `PHGenFit` and `PHG4TrackKalmanFitter`
in [PR247](https://github.com/sPHENIX-Collaboration/coresoftware/pull/247)
And didn't get a chance to fix memory leaks in this `PHG4TrackFastSim`.
It is great you did it.

May I ask is there memory leaks in `PHGenFit` related classes?
Or could you provide a valgrind log file?

johnlajoie:
I didn't see any memory leaks in PHGenFit once I made the modifications in this set of patches. I was seeing leaks before related to the wrapper class PHGenFit::Track not being deleted properly.  

Have you looked into the uninitialized variable complaints in genfit?  There were quite a few of those.

HaiwangYu:
Great, it is good to see the PHGenFit is OK.

Here is the a valgrind log file of running `PHG4TrackKalmanFitter`
https://www.phenix.bnl.gov/WWW/p/draft/yuhw/PullRequest/2017-02-27/1/valgrind_smart_ptr_hijing.log
I didn't find uninitialized variable complaints in genfit.

HaiwangYu:
Great, it is good to see the `PHGenFit` is OK now.

I noticed I only kept the memory leak related valgrind log files. I will re-run some valgrind test for the uninitialized variable complaints in genfit.

johnlajoie:
Great - if you become aware of issues like this in the future, but you don't have time to fix PHG4TrackFastSim, let me know. It took me a bit to track down the problems in PHG4TrackFastSim, if I would have had something pointing me in the right direction it would have gone faster.

HaiwangYu:
Sure. Thank you. Sorry this time. It was my overlook.

PR #282, opened 2017-04-26
TPC simulation stage 1

carlosperezlara:
This commit includes first stage of developments towards new generation TPC microsimulation.
Also enables compatibility of framework to control stepsize in geant4 in TPC gas volume


pinkenburg:
This pull request is outdated. In the meantime we have an agreement how to change it, there is no overlap with this structure.

PR #286, opened 2017-05-03
Starting to use the parameter map, right now for the active and black hole setting

pinkenburg:
The output did not change, this is a major cleanup and a step towards using the parameter maps to store/retrieve all those magical numbers which are hardcoded. It does need a change in the macro to actually set each strip layer active


pinkenburg:
I changed the default to active so existing macros do not need to change. In the next version of this the detector will be able to set it's default active/not active depending on its needs so the default (without macro settings) is what you normally want to run

PR #291, opened 2017-05-12
sPHENIX's own Geant4->GDML export module

blackcathj:
Introducing sPHENIX's own Geant4->GDML export module, based on the Geant4 exporter, ```G4GDMLWriteStructure```. The goal is to allow selective exportation of volumes.

#183 introduced geometry management to export Geant4 geometry to reconstruction stage via the path of Geant4->GDML->TGeo. However, this process produce a large geometry file for fully projective SPACAL simulation, as it has 300k unique fiber physical volumes, which take large space (GB), and long time (20 min) to process. 

To help speed up the coming productions, this pull request introduce our own Geant4->GDML export module, with which we can selectively turn off volumes to be exported. Without enable this feature, the exported GDML files are identical with the nightly build. 

With this pull request, the geometry export is turned on by default and only the SPACAL fiber is not exported. The process time is negligible when compared with a nominal Geant4 simulation. 

In order to compile coresoftware with this pull request, package ``` simulation/g4simulation/g4gdml ``` should be compiled before ```coresoftware/simulation/g4simulation/g4main```. It does not require macro modifications. 

Request @HaiwangYu  to cross check this pull request to ensure compatibility with the downstream analysis code that use the exported geometry objects.


HaiwangYu:
Tested successfully without laddered INTT, which has a known different problem.

blackcathj:
Thanks @HaiwangYu  for the cross check. Merging and adding this package to nightly-build: https://github.com/sPHENIX-Collaboration/utilities/blob/master/utils/rebuild/packages.txt

PR #307, opened 2017-05-30
Tracking with vertex off (0,0,0)

HaiwangYu:
This is a temporary fix.
Reads in truth vertex and smear it with 100 micro to mimic a reconstructed vertex.
This enables tracking for events from vertex off (0,0,0) for now.


blackcathj:
Thanks Sourav and Tony for the cross check!

PR #312, opened 2017-06-01
PHG4KalmanPatRec patch: DCA resolution; configure number of MVTX, INTT layers

HaiwangYu:
# DCA resolution

After several discussions with Christof, Tony, Jin, Sanghoon, Sourav and other people, we figured out that the large DCA resolution is to blame the full fitting module `PHG4TrackKalmanFitter`. That the DCA was track extrapolation based on the cluster with lowest cluster ID which is wrong with the laddered geometry. The extrapolation should be based on the cluster in the most inner layer. This is fixed in this PR.

I also modified the [MVTX](https://github.com/HaiwangYu/coresoftware/blob/KalmanPatRec/simulation/g4simulation/g4hough/PHG4SvtxClusterizer.C#L1222), [INTT](https://github.com/HaiwangYu/coresoftware/blob/KalmanPatRec/simulation/g4simulation/g4hough/PHG4SvtxClusterizer.C#L893) cluster cov calculation regarding to the rotation angle. Adding one [interface function](https://github.com/HaiwangYu/coresoftware/blob/KalmanPatRec/simulation/g4simulation/g4detectors/PHG4CylinderGeom_Siladders.h#L100) to obtain the INTT strip phi tilt angle. It would be great if  @adfrawley and @gmitsuka could review those changes.

Single muon performance after these changes is attached.
[KalmanPatRec_2017-06-01_cluster_order.pdf](https://github.com/sPHENIX-Collaboration/coresoftware/files/1046045/KalmanPatRec_2017-06-01_cluster_order.pdf)

-----------------------------------------

# configure number of MVTX, INTT layers

As requested by @adfrawley and @gmitsuka , we want the tracking could also be run with various number of MVTX, INTT layers. This is also implemented in this PR. @adfrawley and Sourav is testing.
Example of using:
`PHG4KalmanPatRec* kalman_pat_rec = new PHG4KalmanPatRec("PHG4KalmanPatRec", n_maps_layer, n_intt_layer, n_gas_layer);`
or [here](https://github.com/HaiwangYu/macros/blob/KalmanPatRec/macros/g4simulations/G4_Svtx_maps_ladders%2Bintt_ladders%2Btpc_KalmanPatRec.C#L417) for a complete working macro.
Note [this part](https://github.com/HaiwangYu/macros/blob/KalmanPatRec/macros/g4simulations/G4_Svtx_maps_ladders%2Bintt_ladders%2Btpc_KalmanPatRec.C#L196) of the macro is also slightly modified from the current sPHENIX:master for running 0 MAPS configuration.

An event display without MVTX:
image

Sourav, Christof and Tony are making more thorough tests.


blackcathj:
Thanks for @HaiwangYu  for fixing the ladder geometry interface, and Thanks for Tony, Sourav and Christof for running the test over the weekend. Merging.

PR #314, opened 2017-06-02
Adding RawClusterRecalibrator for CEMC

jdosbo:
RawClusterRecalibrator implements position dependent recalibration on CEMC clusters. I also added into the g4cemc Makefile.am "RawClusterv1.h" since it was not there previously, so I couldn't globally include it in the header file for RawClusterRecalibrator.


pinkenburg:
Using versioned classes is a recipe for disaster. The versioned class should only be used at the time of the direct creation of a DST object. From then on always use the base class - even if you copy the object (the versioned classes should have a copy ctor which returns a base class pointer). Specifically this here: 
RawCluster *cluster = 
+      dynamic_cast<RawClusterv1*>(iter->second);
+
+    float clus_energy = cluster->get_energy()
will break if we introduce another version which might just add a parameter. The static cast will return a null pointer and the next line will segfault.
I can't find the code where the rawtowerv1 is being used and it might be a valid case but to prevent the above we normally do not install those version class headers

PR #319, opened 2017-06-09
Change ReadEICFiles to create HepMC node

nfeege:
The previous version of the ReadEICFiles module (which reads Monte Carlo events from trees created by the eic-smear package) directly created a PHG4INEVENT (PHDataNode) node and placed all final-state-particles from the input file in the Geant4 detector simulation.

This pull request changes the output so that ReadEICFiles creates a HepMC node instead of the PHG4INEVENT node. This makes this module work like other event generators (like PHPythia6) in Fun4All and helps streamline modules that use the output from either module.

Running the HepMCNodeReader module after ReadEICFiles puts the final state particles from the HepMC node in the Geant4 simulation, just like it works for the PHPythia6 and other event generator modules.


pinkenburg:
Just for the record - this ReadEICFiles module was a bad quick hack of mine to read eic files. In the long run this should become an input manager which would make this consistent with everything else.

PR #322, opened 2017-06-16
psTOF Added

mchiu-bnl:
New mRPC psTOF at radius between 80-90 cm.  First commit.


pinkenburg:
Sorry - that code needs some cleanup. The stepping action was copied from ancient code which leaves zero energy hits and the handling of parameters does not seem consistent. We also don't want the position parameters hardcoded in large arrays. It might be worthwhile to implement this with our generic phg4block detectors since these are just boxes of glass so far. Then we get fixes for all the above for free (for the price of going through the phg4block detectors making them work)

PR #325, opened 2017-06-24
Add PHSartre Diffractive/UPC Event Generator

johnlajoie:
PHSartre is a wrapper for the Sartre 1.20 event generator for diffractive and UPC events. It works in much the same way as the PHPythia event generators, and the event output is in HepMC format, suitable for the sPHENIX G4 simulations. The event generator includes a generator level trigger interface and a simple particle trigger. 

(Sartre itself and the root upgrades required have been previously installed by @pinkenburg  - thanks Chris!)


pinkenburg:
Also added it to packages.txt to get this into our build

PR #328, opened 2017-06-30
Fix for TPC cluster efficiency

carlosperezlara:
The patch below fix the problem of truncating neff in between simulation and clusterization.
The problem was there all along, but become manifest when increasing cell density (which reduced neff occupancy)

The update effectively adds a constant amplification term that avoid tripping of cloud shape by the digitizer. Mods in PHG4CylinderTPCReco.cc

The clusterizer added the factor back to estimate the number of effective electrons from the adc signal. Mods in PHG4TPCClusterizer.C

====

Adding also missing includes in PHG4Reco to deal with G4StepLimiter and makefile.


simulation/g4simulation/g4detectors/PHG4CylinderCellTPCReco.cc
```diff
@@ -403,7 +403,7 @@ int PHG4CylinderCellTPCReco::process_event(PHCompositeNode *topNode)
 	    double zLim1 = 0.5*M_SQRT2*( (iz+0.5)*zstepsize - zdisp )*cloud_sig_zz_inv;
 	    double zLim2 = 0.5*M_SQRT2*( (iz-0.5)*zstepsize - zdisp )*cloud_sig_zz_inv;
             double z_integral = 0.5*( erf(zLim1) - erf(zLim2) );
-            float neffelectrons = nelec*( phi_integral * z_integral );
+            float neffelectrons = 2000*nelec*( phi_integral * z_integral ); // adding constant electron avalanche (value chosen so that digitizer will not trip)
```
blackcathj: There is no fluctuation in charge amplification that may introduce additional pad to pad variation (only carry fluctuation from initial ionization-drift). Is that is intended?

carlosperezlara: Hi Jin, Yes, fully intentional. A constant gain was added to deal with the practical limitation at the moment. No fluctuations are added at this stage. The results were presented in today's tracking meeting. 

https://indico.bnl.gov/conferenceDisplay.py?confId=3284

sorry forgot to add this link in my original comments.

blackcathj: Thanks for the confirmation!

PR #340, opened 2017-07-27
Modified the parameter initialization so it does not hard-code four I…

adfrawley:
…NTT layers
The default parameter initialization assumed there were 4 INTT layers, so trying to drop the outer layer caused the code to quit. I modified PHG4SiliconTrackerSubsystem so that it counts the number of layers requested and only defines parameters for those layers.
At present, layers can be dropped only from the outside in (i.e. drop outer layer to get 3 layers, then second outer layer to get 2 layers, ....).
Tested, and it works.

### DIFF
--- a/simulation/g4simulation/g4detectors/PHG4SiliconTrackerSubsystem.cc
+++ b/simulation/g4simulation/g4detectors/PHG4SiliconTrackerSubsystem.cc
@@ -23,10 +23,14 @@ PHG4SiliconTrackerSubsystem::PHG4SiliconTrackerSubsystem(const std::string &dete
   , layerconfig_(layerconfig)
   , detector_type(detectorname)
 {
+  nlayers = 0;
   for (vector<pair<int, int>>::const_iterator piter = layerconfig.begin(); piter != layerconfig.end(); ++piter)
   {
+    if(verbosity > 1) cout << PHWHERE << " adding INTT layer " << (*piter).second << endl;
+    nlayers++;
     AddDetId((*piter).second);
   }
+
   InitializeParameters();
   // put the layer into the name so we get unique names
   // for multiple layers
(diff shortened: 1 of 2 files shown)


sPHENIX-GitHub-l:
Hi Tony, I am interested in knowing more about this exercise – simulating the effect of dropping layers of the INTT to check the tracking efficiency? From the MVTX standpoint the inner layer of the INTT is the hardest one to integrate – since it is so close to the outer cylinder of the MVTX and we do not know (yet) what the total profile for these ladders will be.
Cheers,
Walter

From: sPHENIX-GitHub-l [mailto:sphenix-github-l-bounces@lists.bnl.gov] On Behalf Of adfrawley
Sent: Wednesday, July 26, 2017 9:25 PM
To: sPHENIX-Collaboration/coresoftware <coresoftware@noreply.github.com>
Cc: Subscribed <subscribed@noreply.github.com>
Subject: [Sphenix-github-l] [sPHENIX-Collaboration/coresoftware] Modified the parameter initialization so it does not hard-code four I… (#340)


…NTT layers
The default parameter initialization assumed there were 4 INTT layers, so trying to drop the outer layer caused the code to quit. I modified PHG4SiliconTrackerSubsystem so that it counts the number of layers requested and only defines parameters for those layers.
At present, layers can be dropped only from the outside in (i.e. drop outer layer to get 3 layers, then second outer layer to get 2 layers, ....).
Tested, and it works.

You can view, comment on, or merge this pull request online at:

  https://github.com/sPHENIX-Collaboration/coresoftware/pull/340

Commit Summary

  *   Modified the parameter initialization so it does not hard-code four INTT layers

File Changes

  *   M simulation/g4simulation/g4detectors/PHG4SiliconTrackerSubsystem.cc<https://github.com/sPHENIX-Collaboration/coresoftware/pull/340/files#diff-0> (33)
  *   M simulation/g4simulation/g4detectors/PHG4SiliconTrackerSubsystem.h<https://github.com/sPHENIX-Collaboration/coresoftware/pull/340/files#diff-1> (3)

Patch Links:

  *   https://github.com/sPHENIX-Collaboration/coresoftware/pull/340.patch
  *   https://github.com/sPHENIX-Collaboration/coresoftware/pull/340.diff

adfrawley:
Hello Walter,


The simulations I am running now are with the outer layer of the INTT dropped.They are aimed at seeing if reducing the INTT from 4 layers to 3 layers has an effect on tracking performance. I was asked to do this by Ed because there have been questions about it in past reviews. I think that the questions really amounted to "if necessary, could you save money by dropping a layer without destroying the performance".


The reason that I chose the outer layer to drop is that I could do it without significantly modifying the G4 detector setup code - i.e. it was the easy one to drop. It would not be that big a deal to modify the code to drop any INTT layer, or layers. From a tracking point of view, I doubt that whether one dropped the inner or outer layer would make much difference.


We should keep in mind that the INTT is fast enough to resolve a beam crossing while the TPC, and to a smaller extent the MVTX, integrate over multiple beam crossings because of their long signal collection times.  We do not yet have this event pileup incorporated into the simulations, and so conclusions we draw about the effect of dropping INTT layers based on current simulations are tentative at best. This will be a very important function of the INTT in sPHENIX.


Cheers

Tony

PR #341, opened 2017-07-28
Initial commit of PHG4GenFitTrackProjection

HaiwangYu:
## `PHG4GenFitTrackProjection`

Project track to Calorimeters using GenFit projection.
GenFit projection used Runge-Kutta swimming and TGeo material effect interface.
It is supposed to be more accurate than ideal helix projection which is used in `PHG4SvtxTrackProjection`.

So this module could be used to substitute the `PHG4SvtxTrackProjection` module after thorough testing in the future.

Quickly tested with 10 single pions.
image

-------
Minor changes to PHG4KalmanPatRec:
- Add setter and getter of the pid guess
- change one debugging line


blackcathj:
Thanks for the update. Please submit a pull request on macros to switch to this track projection module by default. @HaiwangYu

PR #342, opened 2017-08-01
Disable geometry export for EMCal blocks and wall volumes

blackcathj:
Following discussion of today's simulation meeting on #341, disabling the geometry export for EMCal blocks and wall volumes. In the current Geant4->GDML->ROOT TGeo export/import chain, the SPACAL blocks was not properly translated and is placed overlapping active volumes of tracking detectors. 

@HaiwangYu : please cross check


blackcathj:
Thanks @HaiwangYu  for the cross check!

PR #345, opened 2017-08-22
Bug fix for PHG4GenFitTrackProjection

HaiwangYu:
Thanks for the feedback from Sasha L. and John. L.
Found a major bug in the projection module.

The `genfit::StateOnPlane` for the projection was initialized with wrong pz = 0 GeV
Fixed it in this pull request.

Tested with e-, pT = 1 GeV/c, |vtx_z| < 5cm, eta = 0.8.

matching deta:
image

matching dphi:
image

electron EMCal3X3/pT
image

### DIFF
--- a/simulation/g4simulation/g4hough/PHG4GenFitTrackProjection.C
+++ b/simulation/g4simulation/g4hough/PHG4GenFitTrackProjection.C
@@ -248,12 +248,9 @@ int PHG4GenFitTrackProjection::process_event(PHCompositeNode *topNode) {
 					}
 				}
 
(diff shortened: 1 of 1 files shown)


blackcathj:
Thanks for the update. Since you are here, please also fix a warning pointed out by cppcheck at line 105 of ```PHG4GenFitTrackProjection.C```:

```c++

	return Fun4AllReturnCodes::EVENT_OK;  // <------ code after this can not be reached. Remove it

	if (verbosity > 0) {
	.....

```

PR #349, opened 2017-08-29
Central management of magnetic field

blackcathj:
As we are working on more sophisticated embedding and analysis, there is more and more need for a centralized management of magnetic field to ensure a consistent field are used in all stages of simulation and reconstruction.  

This pull request previews a new lib perform such tasks and updates downstream modules to use it. Specifically: 
* New persistent DST node ```TOP/RUN/FIELD_CONFIG``` record configuration of magnetic field, such as field map file location and fiend strength scaling factor. For embedded simulation and for reconstruction modules, the field setting is automatically loaded from the DST configuration.
* New transient DST node ```TOP/PAR/FIELD_MAP```, which is the numerical table of the magnetic field map and performs field interpolation.
* Light weighted wrappers to interface the above field map in both Geant4 simulation and GenFit-2 reco. 
* Utility module to grab or construct field map

If done correctly, the swap to new field setup should be transparent to users and produce identical simulation results. The reconstruction quality should be slightly improved due to identical map are used in both sim and reco. 

Check is on going and results will be posted as comments. Suggest further check by @adfrawley and @HaiwangYu 

PS: #348 is absorbed into this pull request.


blackcathj:
Update on additional checks:

## Geant4 reproducibility 

Verifying calorimeter simulation output from Geant4 is identical, since the same field map is used. In this particular picture, 4 GeV/c muon sample of both charges covering |eta|<0.9 in outer HCal is plotted:

image


## Improvements in track reconstruction. 

This pull request apply the same bilinear interpolation of field map from Geant4 simulation to the GenFit-2 reco modules (previous used the nearest interpolation). This result in minor improvement in the core width of reconstructed muons. 4 GeV/c muons of both charges covering |eta|<0.9 are used for this check. 

image
image

The charge sign is also checked to be consistent with generated particles.

PR #364, opened 2017-10-03
HEPMC updates for event pile up simulations

blackcathj:
This is mainly work from Mike McCumber and Yorito Yamaguchi, which allow collision event pile up simulations using HEPMC generators and properly performing the time and space vertex displacement.

This is in preview stage. Some update may follow prior to merge.


blackcathj:
Current status:
Few more fixes to recover the barcode tagging as introduced in #216 and recover the event generators when the HEPMC nodes switch to ```PHHepMCGenEventMap```.

However, the embedding simulation of Pythia into pre-generated HIJING/pile up background simulation still has problem (duplicated HIJING simulation). Truth tracing into HEPMC records is still clunky (difficult to find which HEPMC event a track come from). 

Fixing these issues will require further work on this branch, which could be a continuation of this pull request, or merge this first and start a new one.

blackcathj:
Status to be reported in simulation meeting https://indico.bnl.gov/conferenceDisplay.py?confId=2715

blackcathj:
# Update since last meeting

*In tracking evaluator: Speed up evaluation of gpoint ntuple
*In tracking evaluator: tag secondary particles of embedded particle as embedded particle. The previous version tag all secondary particle with embedding ID = 0. Then for @sookhyun for pointing out this problem.
*In framework random generator seeding: generate a sequence of predetermined seeds for modules, if use fixed overall seed. Print out message of how to reproduce the random number sequences.

From the test below, the pile up mechanism appears working as expected. But the TPC timing response is problematic.

Since many package is involved by this pull request, it is hard to test without merging into nightly builds. Suggest merge then fix TPC drift simulation in a separated pull request.

# Note to users

Once this pull request is merged or during testing, the update touches many packages. One need to update build of the changed packages following the order of nightly build: https://github.com/sPHENIX-Collaboration/utilities/blob/master/utils/rebuild/packages.txt

Once this pull request is merged into nightly builds, suggest rebuild all local libs to refresh them.

# Check of the simulation results

## Number of pile up vertexes.

With const double pileup_collision_rate = 100e3 Hz collisions, within the default +/-17.5 us timing window, the presence of the pile up collision is plotted as below. We should expect it is the complementary cumulative distribution function of Poission distribution with u=3.5. :

image


Note: The pile up sub-event are placed on negative embedding IDs

## pile up vertex distribution

From truth pile up primary vertex from 100 Hz AuAu collision placed in +/-17.5 us window and 30 cm collision vertex width.

The z-position should be Gauss distribution with sigma of 30 cm as set by default (can be changed via macro input):
image


The time should be flat distribution with length of 2x 17.5 us. The bunch structure is simulated but not visible in this binning scale:
image


## Causality check of HepMC secondary vertex placement

Checking placement of secondary vertex from Pythia8 -> HepMC -> Geant4 simulation chain. It should be below the speed of light limit:

image


## Tracking response check

Checking hits in the tracking detectors and those with clusters reconstructed in 100kHz AuAu pile up simulations.

The cluster association appear to be correct for INTT ([-20ns, 80ns] timing window) and MVTX (currently [-2, 2] us timing window). But it is too narrow for TPC, which should have more than 10us response window:
image


Checking the response of TPC in the pile up hits. Due to the collision time difference, we should expect the reco cluster position cover all z, but it does not yet:
image


Checking more on the drift simulation of the pile up PHG4 hits hitting the positive-z half of the drift volume. The slope represents the +3 cm/us drift. But the timing window with successful cluster reco is too narrow, only show up within +/- 2 us.

image

PR #365, opened 2017-10-06
Allow inserting block dividers in EMCal active volume

blackcathj:
During the EMCal meeting ( https://indico.bnl.gov/conferenceDisplay.py?confId=3577 ) , we need to answer a design decision regarding whether to place block dividers in EMCal active volume to assist assembly. If so, what would be the optimal material: Stainless steel or carbon/glass fiber type of materials. 

This pull request allow us to perform simulation to reach a conclusion quantitatively.  The default behavior is no block divider which is the setting for the current reference design. 

Similar to the engineering design, 15 mil divider space is inserted into between EMCal block rows: 
* User can turn ON/OFF via macro by setting parameter ```divider_width``` to a positive value. Suggested width is 14.5 cm. Default is OFF with width of 0 cm, until the EMCal group finalize the decision on the divider.
* The material choice of the divider can be set in macro too: via parameter ```divider_mat```
 
An example macro of enabling simulating a FR4 glass fiber divider is as following:
```c++
    cemc = new PHG4SpacalSubsystem("CEMC", ilayer);

    //......

    cemc->UseCalibFiles(PHG4DetectorSubsystem::xml);
    cemc->SetCalibrationFileDir(string(getenv("CALIBRATIONROOT")) + string("/CEMC/Geometry_2017ProjTilted/"));
    cemc->set_double_param("radius", radius);            // overwrite minimal radius
    cemc->set_double_param("thickness", cemcthickness);  // overwrite thickness

//    cemc->set_double_param("divider_width", 0);  // turn off divider
    cemc->set_double_param("divider_width", 14.5);  // overwrite thickness
//    cemc->set_string_param("divider_mat", "G4_AIR");  // Effectively no divider
//    cemc->set_string_param("divider_mat", "SS310");  // Possible dense material
    cemc->set_string_param("divider_mat", "FR4");  // Possible light material
    
// ....
```

## QA checks

Since this only introduce 15mil extra material between 2x2 blocks in the azimuthal direction, the change in bulk response is negligible. Checked with electron with pT = 16 GeV in full acceptance:

### Steel divider vs air divider (effectively no divider)
image
image

### Nylon divider (gemerioc) vs air divider (effectively no divider)
image
image

## Request of follow up studies

Request @jdosbo to perform check of phi-boundary response with 5 GeV electron/positron and 20 GeV photons.  We can compare no divider, FR4 fiber glass divider and SS310 steel divider

### DIFF
--- a/simulation/g4simulation/g4detectors/PHG4CylinderGeom_Spacalv3.cc
+++ b/simulation/g4simulation/g4detectors/PHG4CylinderGeom_Spacalv3.cc
@@ -95,6 +95,8 @@ PHG4CylinderGeom_Spacalv3::SetDefault()
   sidewall_outer_torr = 0.030000;
   sidewall_mat = "SS310";
   max_phi_bin_in_sec = 8;
+  divider_mat = "G4_AIR";
+  divider_width = 13.5;
 }
 
 void
@@ -110,6 +112,10 @@ PHG4CylinderGeom_Spacalv3::ImportParameters(const PHG4Parameters & param)
     sidewall_mat = param.get_string_param("sidewall_mat");
   if (param.exist_int_param("max_phi_bin_in_sec"))
     max_phi_bin_in_sec = param.get_int_param("max_phi_bin_in_sec");
+  if (param.exist_string_param("divider_mat"))
+    divider_mat = param.get_string_param("divider_mat");
+  if (param.exist_double_param("divider_width"))
+    divider_width = param.get_double_param("divider_width");
 
   // load sector_tower_map
   if (param.exist_int_param("sector_tower_map_size"))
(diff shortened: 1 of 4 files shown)


blackcathj:
Forgot to attach the detector display, produced during check of this pull request. 

The block divider is the think light green sheet inserted between gray blocks. In this transverse projection, the sheet is show as a line. The divider is about 15 mil thick and a 1 um air gap is left between the divider and the blocks. Only the divider near the active block volume is simulated, which is the relevant part for this study. 

## End view of few EMCal sectors:

spacal_tilted43

## Zoom into the inner radius side of the blocks:

spacal_tilted45
spacal_tilted44

## Zoom into the outer radius side of the blocks:
spacal_tilted42

PR #368, opened 2017-10-18
Tower Energy Cut and Tower Energy Scale prior to Jet Finder

SongkyoLee:
Tower Energy Cut (default=0MeV) and Tower Energy Scale (default=1 for CEMC, 1.49 for IHCAL and OHCAL) are added. Values can be modified in macros/macros/g4simulations/G4_Jets.C. More study will be followed up, but currently, it is still useful to have constant values. (presentation at Jet Topical Meeting: https://indico.bnl.gov/getFile.py/access?contribId=3&resId=0&materialId=slides&confId=3749)

### DIFF
--- a/simulation/g4simulation/g4jets/TowerJetInput.C
+++ b/simulation/g4simulation/g4jets/TowerJetInput.C
@@ -25,9 +25,11 @@
 
 using namespace std;
 
-TowerJetInput::TowerJetInput(Jet::SRC input)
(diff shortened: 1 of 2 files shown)


johnlajoie:
This pull request will be superseded or updated by later work from Songkyo.  She is working on setting the scales for the inner and outer HCAL using single hadrons using the same techniques that were developed for the test beam.  I think she will be able to present an update at the simulations meeting on 11/7.

PR #373, opened 2017-10-24
Update tracking evaluator for TPC off-time hit analysis

blackcathj:
This is a follow up to #364 on pile up simulation. 

In #364 we reported that the TPC cluster -> truth hit associated seems to be constrained to very narrow timing window of about +/-2us in evaluator ntuple ```ntp_g4hit```, while the pile up simulation is expected to plant off-time collisions up to 17 us by default. Meanwhile, @adfrawley later confirmed the z-range of reconstructed TPC clusters appear to be correct in ```ntp_cluster```. 

This problem was traced to an speed optimization in building ```ntp_g4hit```, for which only clusters within +/- 5 cm from the truth hit position are considered for matching with g4hits. Therefore, this pull request edit the optimization criteria to allow evaluating matching of g4hit -> all cluster in the same z-half to accommodate the drift of off-time TPC hits.

This is expected to recover the full z and time range of TPC truth hit->cluster matching in evaluatiors. Test is on going. Meanwhile, it would slow down the building of ntuple ```ntp_g4hit```, if it is enabled in the macros.

### DIFF
--- a/generators/phhepmc/Fun4AllHepMCOutputManager.cc
+++ b/generators/phhepmc/Fun4AllHepMCOutputManager.cc
@@ -1,8 +1,10 @@
 #include "Fun4AllHepMCOutputManager.h"
 
 #include "PHHepMCGenEvent.h"
+#include "PHHepMCGenEventMap.h"
 
 #include <fun4all/Fun4AllServer.h>
+#include <fun4all/Fun4AllReturnCodes.h>
 #include <phool/getClass.h>
 
 #include <HepMC/IO_GenEvent.h>
@@ -16,6 +18,7 @@
 #include <TPRegexp.h>
 
 #include <ostream>
+#include <cassert>
 #include <fstream>
 
 #include <boost/iostreams/filtering_streambuf.hpp>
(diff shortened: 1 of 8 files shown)


blackcathj:
# Updates

Following the last simulation meeting, two updates on this pull request:
1. Following suggestion from Christof, restricting the g4hit->cluster matching space in order to keep CPU consumption low. The new matching is in the non-drifting azimuthal dimension. To speed up, an approximated phi location and matching window of 100 mrad is used. 
2. Tony showed clusters #364 does not show a drift boundary for pile up collisions produced after the primary collision in time. This problem trace down to the TPC digitization stage, where a timing boundary cut is missing around the central TPC membrane. This cut is added into ```PHG4CylinderCellTPCReco```

# Note to users: 

Without this pull request, the TPC pile up hit density will be too high for pile up collisions happened after the primary collisions. 

# Tests

Wrapping up the tests with 100 event of 100 kHz pile up simulation in Au+Au collisions: 

* Production time for the primary particles in the pile up collisions. Note the time window is now +/- 35 us as set in https://github.com/sPHENIX-Collaboration/macros/pull/88 

image

* Drift check for pile up hits. Comparing with #364, showing the full range of drifting in TPC: 

image
 
* Drift boundary check for pile up hits: truth hits in each drifting volume of TPC are only accepted for the time window corresponding to the hit position within the same volume in its digitization chain. Each diagonal stripe is corresponding to one pile up collision with distinct collision time offset to the primary collision

image

PR #385, opened 2017-11-14
Update the definition of primary vertex in the truth information

blackcathj:
Primary vertex in the truth information is used for the fast simulation of BBC and the current place-holder for the vertex seeding. After the implementation of the pile up tool chain, the primary vertex is defined as the first vertex with highest embedding ID as in #372. 

However, as Sanghoon find out, in the Pythia-8 event simulations, the collision vertex is saved to the last vertex in the truth container, and the first vertex is usually a decay vertex which leads to low track seeding efficiency. Therefore, this pull request change the primary vertex definition to the last vertex with the highest embedding ID. 

Tested with few Pythia-8 B-jet events. Looks working fine.


sPHENIX-GitHub-l:
Hi JIn,

I confirmed the consistency between generated vertex and seed vertex with 100 single pythia events. Thanks for the quick fix.

Best,
Sanghoon

PR #386, opened 2017-11-14
Standalone RAVE vertexing module

HaiwangYu:
Standalone vertexing module using RAVE.

## Usage:
```
PHRaveVertexing* rave = new PHRaveVertexing();
se->registerSubsystem(rave);
```

By default it create new `SvtxVertex` node 'SvtxVertexMapRefi'.
User could choose to overwrite the 'SvtxVertexMpa' node by:
`PHRaveVertexing::set_over_write_svtxvertexmap(true)`

Users could also set the vertexing alg. by:
`PHRaveVertexing::set__vertexing_method("$some-alg-name$")`
default alg. is "avf-smoothing:1"
more info [here](https://rave.hepforge.org/trac/wiki/RaveMethods)

## How it works:

Based on Kalman information saved in the `SvtxTrackState`.
Convert the Kalman state to `genfit::Track`s.

Flow chart:
image

Due to `genfit::Track` used 5D (q/p, u', v', u, v) state/cov-matrix internally.
Translating between 6D(x,y,z,pz,py,pz) causes some slight difference.

## Validation
Plot below shows the difference between standalone module (this PR) and the RAVE vertexing together with genfit track fitting. The gaussian width is about 0.4-0.5 micron. The test is based on pp jet events. 

image


blackcathj:
Since this module does not change the default behavior of the tracking reco, merging to include build and avoiding it to drift apart from the master branch

PR #388, opened 2017-11-15
Random generator cleanup

pinkenburg:
This PR maximizes entropy by always returning a value from rdev rather than a uniform distribution initialized by a single seed. With PHRandomSeed::LoadSeed(unsigned int) seeds can be preloaded to reproduce any sequence. If the RANDOMSEED flag is set, it still uses this seed to initialize a uniform random number distribution.
It prints out the random seeds so a script can pull the seeds from a log file to recover the sequence. It is not necessary anymore for consumers to print out the seed themselves

### DIFF
--- a/offline/framework/phool/PHRandomSeed.cc
+++ b/offline/framework/phool/PHRandomSeed.cc
@@ -1,29 +1,39 @@
 #include "PHRandomSeed.h"
 #include "recoConsts.h"
 
-#include <gsl/gsl_const.h>
-#include <gsl/gsl_randist.h>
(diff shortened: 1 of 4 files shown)


blackcathj:
The per-event get of random seed is traced to https://github.com/sPHENIX-Collaboration/coresoftware/blame/master/simulation/g4simulation/g4hough/PHG4KalmanPatRec.C . I think we should use per-run random seed here. I can make an update to it. 

Meanwhile, I think there is no need to change this pull request in suppressing the printing out of the random seed at each call.  @pinkenburg

PR #396, opened 2017-12-05
Modifications in PHENIX-style Clustering

shura95:
* RawCluster.h,cc:
    New filed "ecore" and corresponding get and set functions
* BEmcRec.cc:
    New shower profile parameters and other minor modifications (more clear error message if cluster is too large ... I'll remove this limitation later)
* BEmcCluster.h,cc:
    Change output data format in GetPeaks(): from fixed size array to vector
    Introduce GetECoreCorrected()
    Increase parameter fgMaxNofPeaks to make sure clustering will go on even for very large clusters
    Tune ECore for sPHENIX
    Other minor modifications
RawClusterBuilder.cc
    AbortEvent on clustering error with proper message
    Set default tower threshold of 20MeV - looks safe even for central HI collisions
    Some more modification related to changes in BEmcCluster.h(cc), RawCluster.h(cc)

### DIFF
--- a/simulation/g4simulation/g4cemc/BEmcCluster.cc
+++ b/simulation/g4simulation/g4cemc/BEmcCluster.cc
@@ -12,7 +12,7 @@ using namespace std;
 // Define and initialize static members
 
 // Max number of peaks in cluster; used in EmcCluster::GetPeaks(...)
-int const EmcCluster::fgMaxNofPeaks=10;
+int const EmcCluster::fgMaxNofPeaks=1000;
 
 // Used in EmcCluster::GetPeaks(...), it is the number of iterations
 // when fitting photons to peak areas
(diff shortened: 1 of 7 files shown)


blackcathj:
Hi, Martin

Thanks for bring this up! Since we do not have staged production with the clusters, we do not yet have to fix the v1 version of cluster yet. Keeping evolving it with software development helps to limit many versions of clusters before we do have first staged cluster production. Then since our tentative plan is to switch to Sasha's clusterizer for both EMCal and HCals, all clusters would carry the new E_Core variable. And therefore, I feel it is proper to include it into the default version of cluster data structure. I am happy to learn further suggestions too. 

Meanwhile, as you pointed out, the use of ROOT versioning is not intended for any serious purpose. The +1 for the ROOT versioning was just adding the convenience to avoid upset the DST read back from past tests. 

Cheers

Jin

PR #399, opened 2017-12-09
Tpc update

pinkenburg:
The tpc gas volume is now split into two volumes (north/south) with a kapton window in the center. This will give more realistic g4hits around the center since they won't cross z=0 anymore


pinkenburg:
Ignore the PHG4TPCElectronDrift module for the time being - this is just development

PR #397, opened 2017-12-06
adding ecore, chi2 for PHENIX clusterizer

jdosbo:

PR #403, opened 2017-12-19
Reorganizing clusters for calorimeters

blackcathj:
# Introduction

Following discussion in the last simulation meeting and offline chat with @jdosbo @shura95 and @dvperepelitsa , this is a preview of reorganizing the calorimeter cluster objects. The goal is reduce the ambiguity of cluster position recording and allow flexible tag of analysis quantity onto cluster object. Specific changes are discussed as following. 

# Changes highlight

## Cluster position reporting 

Previously, clusters are reported in the 2-D phase space of eta-phi. The reporting of eta rely on presumption of vertex position at z= 0 cm. A correction of eta based on a known vertex position, e.g. that reported from MVTX, is complex, which requires knowledge of the reference radius (for CEMC) or z-location (for FEMC/EEMC) of the calorimeters, recalculate 3-D cluster position and then recalculate eta with new z-vertex position. It is also confusing since it is not apparent to users whether the eta reported by clusters are reported with respect to vertex of z = 0 cm or a pre-selected reconstructed vertex. Also given in sPHENIX p+p events, we expect more than one collision per event/crossing, it is really up to analyzer to choose the likely vertex point for eta/eT analysis, not by the framework. 

Therefore, @jdosbo @shura95 and I think it is best that the clusters report its intrinsic information, i.e. 3-D cluster position as measured on the calorimeter. 

Considering both central and forward calorimeters, we found it is convenient to use cylindrical coordinate system for the internal storage:

```c++

class RawClusterv1 : public RawCluster
{
...
  //! cluster ID
  RawClusterDefs::keytype clusterid;
  //! total energy
  float _energy;
  //! Tower operations
  TowerMap towermap;
  //! location of cluster in cylindrical coordinate
  float _r;
  float _phi;
  float _z;
}
```

then we give interface functions to convert the cluster location to  Cartesian coordinate system or to eta and Et with an external vertex position input: 

```c++
class RawClusterv1 : public RawCluster
{
...
  //! access Cartesian coordinate system
  virtual float get_x() const { return get_r() * std::cos(get_phi()); }
  virtual float get_y() const { return get_r() * std::sin(get_phi()); }
  //! convert cluster location to psuedo-rapidity given a user chosen z-location
  virtual float get_eta(const float z) const;
  //! convert cluster E_T given a user chosen z-location
  virtual float get_et(const float z) const;
}
```

## Flexible tag of analysis quantities

As further analysis procedures get developed for clusters, more analysis quantity would be tagged onto clusters. For example, as introduced in #397, @shura95 showed e_core, chi2 and prob are useful in improved EM observable, and @dvperepelitsa suggested to add isolation energy to photon and pi0 clusters too. 

To accommodate new developments without introducing many versions of the clusters, we brought the property map feature from ```PHG4Hit``` to ```RawCluster```. They allow flexible tag of additional quantities on to clusters without new versioning, but perform at a reduced performance in both CPU and space usages. We intend to move many of the analysis quantities to direct variable access in new versions of ```RawCluster``` in the future, when the development of calorimeter analysis settles down on the selection of new quantities. The user interfaces to ```RawCluster``` would not be changed as these properties are accessed via virtual get/set functions. 

To get started, following quantities can be tagged to ```RawCluster``` with this pull request: 

```c++
std::pair<const std::string, RawCluster::PROPERTY_TYPE>
RawCluster::get_property_info(const PROPERTY prop_id)
{
  switch (prop_id)
  {
  case prop_ecore:
    return make_pair("EM cluster energy estimated with its core towers", RawCluster::type_float);
    break;
  case prop_prob:
    return make_pair("cluster template probability for EM shower", RawCluster::type_float);
    break;
  case prop_chi2:
    return make_pair("reduced chi2 for EM shower", RawCluster::type_float);
    break;

  case prop_et_iso:
    return make_pair("isolation ET", RawCluster::type_float);
    break;

  case prop_truth_track_ID:
    return make_pair("truth cluster's PHG4Particle ID", RawCluster::type_int);
    break;
  case prop_truth_flavor:
    return make_pair("truth cluster's PHG4Particle flavor", RawCluster::type_int);
    break;

  default:
    cout << "RawCluster::get_property_info - Fatal Error - unknown index " << prop_id << endl;
    exit(1);
  }
}

```

## Truth cluster objects

@dvperepelitsa and I also suggest building truth cluster nodes using the ```RawCluters``` base, for example for truth photon, electrons and pi-0 clusters. Analyzers can then iterate over truth and reco clusters in the same way by just switching the DST node name. To help recording the truth tracing information in the truth clusters, two optional properties were added as shown above : ```prop_truth_track_ID``` and ```prop_truth_flavor```

# Path forward

This pull request carries all changes proposed for g4cemc lib for discussion and preview. If OK, the changes need to be propagated to other libs in both coresoftware and analysis repositories via an updated to this pull request prior to merging. 

Further more, we need to separate the reconstruction and analysis of the  clusters away from the ```g4simulation``` area. One possible layout would be:

image


blackcathj:
# Update

Follow up to last meeting discussion, here are recent changes: 

* Remove placeholder for fast simulation of truth clusters in ```RawClusterv1```
* Move the cluster location -> vertex-dependent pseudorapidity calculation to a new utility class ```RawClusterUtility```
* Restructure of folders, RawClusterBuilder names, and libraries as suggested in the last post. The move is visualized here: 

test


# What's next

If consensors is made in simulation meeting, will move forward with these changes and update the downstream libs in both coresoftware and analysis repositories to use these new structures.

blackcathj:
# Update

This update completes this pull request by updating the rest of coresoftware to use the new structure of calorimeter analysis. Related pull request are 
* Compile utility : https://github.com/sPHENIX-Collaboration/utilities/pull/11
* Analysis modules : https://github.com/sPHENIX-Collaboration/analysis/pull/9 
* Macros: https://github.com/sPHENIX-Collaboration/macros/pull/95 , which will be merged after this pull request is incorporated into the nightly build. 

# Test

Tested using pT= 24 GeV pi- at eta between 0.3--0.4 and vertex of -5 < z < 5 cm, and the standard calorimeter QA tools. The macro is here:
https://github.com/blackcathj/macros/blob/SinglePart_master_qa_Clusterizer/macros/g4simulations/Fun4All_G4_sPHENIX.C 

This pull request (blue points) is checked against the same setup run with the current nightly-build (green shade). They show consistent performance. 

## Track projection to calorimeter clusters

image

## E/p of calorimeters 

image

## Single pion event analyzed with tower jet finder of anti-kT R=0.7 

image

PR #405, opened 2018-01-18
Relocating the parameter IO utilities for use by wider set of packages

blackcathj:
@pinkenburg  introduced a generic parameter utilities ```PHG4Parameters``` for Geant4 detectors for managing the detector setup parameters during Genat4 detector construction. Later this utility was also picked up for use in the reconstruction modules of calorimeter calibrations, analysis modules, and test beam reconstruction modules. 

Recently, as we are reorganizing calorimeter and tracking software, it is obvious this parameter utility does not belong to the ```g4detector``` libs, but to the framework. Therefore, this pull request propose one way to use to relocate ```PHG4Parameters*``` to ```offline/database/PHParameter/PHParameter*```. Downstream packages in coresoftware, analysis and calibration repositories those are using this utilities are also updated in pull requests. 

We expect the shower profile and tracker geometry objects are next user of this utility. Test and suggestions welcomed.


blackcathj:
From Jan 23 meeting, Chris' comment: take a two stage update strategy: 

1. add new ```PHParameter``` class and use for offline reconstruction, while leaving ```PHG4Parameter``` used for Geant4 simulation
2. Swap ```PHG4Parameter``` to ```PHParameter``` in a later stage. 

This pull request is updated to implement step 1 and leave ```PHG4Parameter``` in the repository.

PR #410, opened 2018-02-05
Calorimeter Clustering: switch between cylindrical and plane geometries; remove the cluster size limitation

shura95:
Modifications to address two goals: 
1. Introduce bCYL flag in BEmcRec to switch between cylindrical and planar geometries (needed e.g. for forward calorimeter). The default value is true (for cylindrical geometry); the switch to planar geometry use BEmcRec::SetPlaneGeometry()
2. Remove the maximal cluster size limitation (now it can deal with any cluster size); despite that is was large anyway (half of the calorimeter size in phi), I removed this limitation properly correcting the code

The modifications should not affect any existing codes in any way.


blackcathj:
Since there is no more comment after the simulation meeting. Merging for use in the next nightly build.

PR #423, opened 2018-03-22
update mapping of inner and outer hcal

sunxuhit:
inner hcal mapping is off by 1, low gain and high gain of outer hcal is reversed


blackcathj:
Thanks for identifying and fixing these problems! Merging for today's build

PR #425, opened 2018-03-28
mRICH

sunxuhit:
update mRICH detector and compatible with sPHENIX Fun4all


sunxuhit:
add module_id to G4HIT_mRICH.scint_id for each RICH modular
module_id = (sector_id-1)*100+mRICH_id 
sector_id (impr) and mRICH_id(pv) are from G4AssemblyVolume::MakeImprint

default module_id is -201

simulation/g4simulation/g4detectors/PHG4mRICHSteppingAction.cc
```diff
@@ -197,43 +281,62 @@ void PHG4mRICHSteppingAction::SetInterfacePointers( PHCompositeNode* topNode )
   }
 }
 //____________________________________________________________________________..
-/*
-int
-PHG4ForwardEcalSteppingAction::FindTowerIndex(G4TouchableHandle touch, int& j, int& k)
+int PHG4mRICHSteppingAction::GetModuleID(G4VPhysicalVolume* volume)
```
blackcathj: Instead of string parsing, is that possible to extract the ID number from copy numbers? One example is via SPACAL:
`bool PHG4SpacalSteppingAction::UserSteppingAction(const G4Step* aStep, bool)`. It will be much faster than string parsing.

simulation/g4simulation/g4detectors/PHG4mRICHDetector.cc
```diff
@@ -752,142 +778,135 @@ void PHG4mRICHDetector::build_lens(LensPar* par, G4LogicalVolume* motherLV)
...
-	rot.rotateY(phi*cos(theta)*180*deg/pi);
+	rot->rotateX(phi*(-1)*sin(theta)*180*deg/pi);
+	rot->rotateY(phi*cos(theta)*180*deg/pi);
       }
 
-      sprintf(name,"mRICH_%d_%d",i+1,j+1);
-      new G4PVPlacement(G4Transform3D(rot, G4ThreeVector(x, y, z)),
-			a_mRICH,name,space,
-                        0, 0, 1);    //last digit for checking overlapping                                                                                                                               
+      mRICHwall->AddPlacedVolume( a_mRICH,pos,rot);
```
blackcathj: Please introduce geometry overlap check in the volume placements. Example is `int PHG4InnerHcalDetector::ConstructInnerHcal(G4LogicalVolume *hcalenvelope)`. And run a geometry overlap check in the end.

simulation/g4simulation/g4detectors/PHG4mRICHSteppingAction.cc
```diff
@@ -47,122 +45,208 @@ using namespace CLHEP;
...
+
 	/* set intial energy deposit */
 	hit->set_edep( 0 );
-	hit->set_eion( 0 );
-	
+	if (isactive == PHG4mRICHDetector::SENSOR) 
+	{
+	  hit->set_eion( 0 );
+	  savehitcontainer = hits_;
+	  if(PID == 0)
```
blackcathj: I assume you are searching for the Cherenkov photon here. Then why their `PID` == 0? Should not be 22? 
Also higher energy X-ray and gammar ray photon will also fall into this category, and the RICH photo sensor does not absorb them. 
Suggest cut on photon PID of 22 and cut on the energy of photon based on the photo-absorption range of the PMT, e.g. cut off in the UV. Then let Geant4 to continue handle propagation of higher energy photons.

simulation/g4simulation/g4detectors/PHG4mRICHSteppingAction.cc
```diff
@@ -47,122 +45,208 @@ using namespace CLHEP;
...
+	savetrackid = aTrack->GetTrackID();
+
 	/* set intial energy deposit */
 	hit->set_edep( 0 );
-	hit->set_eion( 0 );
-	
+	if (isactive == PHG4mRICHDetector::SENSOR) 
+	{
+	  hit->set_eion( 0 );
+	  savehitcontainer = hits_;
```
blackcathj: Here, you probably would only need to record photon hit in the mRICH detector in the `hits_` container. Ionizing hits in the Sensor volume does not get read out and can go to `absorberhits_`

simulation/g4simulation/g4detectors/PHG4mRICHSteppingAction.cc
sunxuhit: I cannot find a easy way to use copy number to get the same module_id in previous code.
I am considering to use an external map to construct mRICH wall and use G4PVPlacement to replace G4AssemblyVolume. And implement a new set of module_id at that time.
I prefer to leave this as it is for right now and I will update it soon.

simulation/g4simulation/g4detectors/PHG4mRICHDetector.cc
sunxuhit: done.

simulation/g4simulation/g4detectors/PHG4mRICHSteppingAction.cc
sunxuhit: based on GEANT website, http://geant4.cern.ch/G4UsersDocuments/UsersGuides/ForApplicationDeveloper/html/Appendix/particleList.src/others/index.html
PID == 0 is for optical photon and PID == 22 is for gamma, therefore I think PID == 0 should be used in the code to select Cherenkov photons.

simulation/g4simulation/g4detectors/PHG4mRICHSteppingAction.cc
sunxuhit: Moved `hit->set_eion(0)` out of if block.

sunxuhit:
construct mRICH through an external map
update module_id with copy number

PR #430, opened 2018-04-17
add light version for vertex evaluator

sookhyun:
Add _do_eval_light option to reduce CPU time.
gntracksmaps will not be counted with _do_eval_light==true.


sookhyun:
close for now to resolve conflicts

PR #433, opened 2018-04-25
add do_eval_light option

sookhyun:
Counting gntracksmaps will be disabled when do_eval_light is turned on.


blackcathj:
Hi @sookhyun , after the last merge on https://github.com/sPHENIX-Collaboration/coresoftware/pull/432 , there is now a confliction between fbd71ac with the updated evaluator. Could you double check and resolve the new confliction? Thanks!

PR #436, opened 2018-04-28
Calorimeter dead tower simulation and recovery/masking in reconstruction

blackcathj:
# Introduction

We were instructed to demonstrate the impact of calorimeter dead towers in CD-1 review. The levels of dead tower are 5% and 10% in the current Objective Key Performance Parameter (Objective KPP) and Threshold Key Performance Parameter (Threshold KPP), respectively. The goal is to show even in the worse case of Threshold KPP, the experiment can continue with reduced physics performance. 

# Dead tower simulation and reconstruction

A few steps were introduced to handle dead towers

1. `RawTowerDeadMapLoader`: Dead map from calibration database is loaded on DST node in container `RawTowerDeadMap`. 
2. `RawTowerDigitizer`: dead tower signal is not readout. However, it can still carry noise hits, which emulate disconnected front end with a working digitizer. 
3. `RawTowerDeadTowerInterp`: in reconstruction stage after tower energy calibration, the known dead tower is recovered via interpolating energy from neighboring non-dead towers. The aim is to recover the background level and some shower information in the dead towers for jet reco and background determination. 
4. `RawClusterDeadAreaMask`: in reconstruction stage after clustering, cluster . 

_Note_: the last step is aimed for photon cluster analysis, and can lead to significantly increased loss of alive area since cluster usually require multiple towers under the core to be _not_ dead. 
* It does not impact track-based clustering `PHG4GenFitTrackProjection` for electron ID analysis as lower energy tail can be tolerated as loss of electron efficiency 
* It does not impact tower/track based jet reconstruction. Dead tower with energy interpolated from neighboring tower would leads to worsen of jet energy resolution. But we can not afford the loss of reconstructed jet due to a few dead towers. 

# Related updates

* Dead map in calibration repository: https://github.com/sPHENIX-Collaboration/calibrations/pull/33
* Example macros that simulates for dead towers, which is tagged as `CD-1 Objective/Threshold KPP`
** `CD-1 reference` with no dead towers : https://github.com/blackcathj/macros/tree/CD-1-reference-single-production-readback 
** `CD-1 reference with Objective KPP`: https://github.com/blackcathj/macros/tree/CD-1-reference-Objective-KPP-single-production-readback
** `CD-1 reference with Threshold KPP`: https://github.com/blackcathj/macros/tree/CD-1-reference-Threshold-KPP-single-production-readback
** `CD-1 reference with Threshold KPP` tagged branch for HIJING embeddings: https://github.com/blackcathj/macros/tree/CD-1-reference-Threshold-KPP-jet-production-readback-embed

_Note_: final macros will be placed on branches of https://github.com/sPHENIX-Collaboration/macros

# Next

Running tests to check the leading order effects. 

After test, I will request Jet Structure TG to test jet reconstruction and @jdosbo for test the EMCal resolution. 

# Acknowledgement

Thanks to Alexander Bazilevsky @shura95  and Gunther Roland for suggestions on the masking/recovery algorithm, and to Chris Pinkenburg @pinkenburg  who refreshed the single particle central production sample for testing.

### DIFF
--- a/offline/packages/CaloBase/Makefile.am
+++ b/offline/packages/CaloBase/Makefile.am
@@ -23,6 +23,10 @@ libcalo_io_la_SOURCES = \
   RawTowerv1_Dict.cc \
   RawTowerContainer.cc \
   RawTowerContainer_Dict.cc \
+  RawTowerDeadMap.cc \
+  RawTowerDeadMap_Dict.cc \
+  RawTowerDeadMapv1.cc \
+  RawTowerDeadMapv1_Dict.cc \
   RawTowerGeom.cc \
   RawTowerGeom_Dict.cc \
   RawTowerGeomv1.cc \
@@ -74,6 +78,8 @@ pkginclude_HEADERS = \
   RawTowerDefs.h \
   RawTowerv1.h \
   RawTowerContainer.h  \
+  RawTowerDeadMap.h  \
+  RawTowerDeadMapv1.h  \
   RawTowerGeom.h \
   RawTowerGeomv1.h \
   RawTowerGeomv2.h \
--- a/offline/packages/CaloBase/RawCluster.h
+++ b/offline/packages/CaloBase/RawCluster.h
@@ -60,7 +60,7 @@ class RawCluster : public PHObject
     PHOOL_VIRTUAL_WARNING;
     return 0;
   }
-  virtual TowerConstRange get_towers()
+  virtual TowerConstRange get_towers() const
   {
     PHOOL_VIRTUAL_WARN("get_towers()");
     static TowerMap dummy;
--- a/offline/packages/CaloBase/RawClusterv1.h
+++ b/offline/packages/CaloBase/RawClusterv1.h
@@ -32,7 +32,7 @@ class RawClusterv1 : public RawCluster
   float get_energy() const { return _energy; }
   //! Tower operations
   size_t getNTowers() const { return towermap.size(); }
-  RawCluster::TowerConstRange get_towers() { return make_pair(towermap.begin(), towermap.end()); }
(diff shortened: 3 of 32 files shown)


blackcathj:
# Test

Tested with 10k 16 GeV e- simulation in CD-1 reference configurations. Three reconstruction tags are compared: 

** `CD-1 reference` which is consistent with the default readback macros: https://github.com/blackcathj/macros/tree/CD-1-reference-single-production-readback 
** `CD-1 reference with Objective KPP`: which carry 5% random dead towers. https://github.com/blackcathj/macros/tree/CD-1-reference-Objective-KPP-single-production-readback
** `CD-1 reference with Threshold KPP`: which carry 10% dead towers https://github.com/blackcathj/macros/tree/CD-1-reference-Threshold-KPP-single-production-readback

Two variables are checked in the track evaluator output: 
* Left panel with `ntp_track/cemce3x3`, which is 3x3-tower energy sum around track projection. The tower is after dead tower recovery with an attempt to recover energy by averaging over neighboring non-dead tower (`RawTowerDeadTowerInterp`). This used for Upsilon electron ID studies. With dead tower recover, it carry best efficiency but has larger reconstructed energy spread due to the loss of dead tower energy information.
* Right panel with `ntp_track/cemce`, which is the closest EMCal cluster energy. The normalization is all good tracks. A cut removed cluster that is more than 1 tower width away from the track projection too (which is usually noise cluster). After the dead tower recovery, cluster set is further masked to reject the cluster whose center is within 1.4 tower width away from any dead tower. This leads to lower reco efficiency: with 10% dead towers, 65% clusters are rejected. But the good cluster passing that survive the cut retain the energy spectrum line-shape of the no-dead-tower scenario (blue). The cluster method are used for photon and photon-jet measurements. 

image

PR #446, opened 2018-06-07
electron side mRICH

sunxuhit:


blackcathj:
Hi, @sunxuhit Thanks for the update!

One request though: it that possible to configure the mRICH class via macro so that it can build e-going mRICH and h-going mRICH separately with two instances? 

During the tests, we might want to test these two mRICH subsystems separately just by adjusting the RICH setup macros. Also there should be two completely separated subsystems in digitization and reconstruction with different node names.

sunxuhit:
@blackcathj 
Thanks for the comments.
I will introduce an external parameter to select which mRICH people want to build.

sunxuhit:
update external parameter subsystemSetup to configure mRICH array:
-1: single module
0: build h-side sectors and e-side wall
1: build h-side sectors
2: build e-side wall
3: build h-side wall <= geometry under development
4: build h-side wall and e-side wall <= geometry under development

simulation/g4simulation/g4detectors/PHG4mRICHDetector.cc
```diff
@@ -90,13 +90,27 @@ int PHG4mRICHDetector::IsInmRICH(G4VPhysicalVolume * volume) const
 void PHG4mRICHDetector::Construct( G4LogicalVolume* logicWorld)
 {
   int subsystemSetup =params->get_int_param("subsystemSetup");
```
blackcathj: Could you use a `enum` to represent the configurations rather than numbers? Here is an example : `enum RawTowerDigitizer::enu_digi_algorithm` https://www.phenix.bnl.gov/WWW/sPHENIX/doxygen/html/d0/db9/classRawTowerDigitizer.html#a1b2848f27d01236afc6c11eb4e5e666c

sunxuhit:
use enum DetectorSetUp for mRICH configurations and tested.

PR #447, opened 2018-06-08
Tracker

damcglinchey:
This pull request fleshes out the new tracking object structure to bring it in line with the plan discussed at the tracking workfest June 3-4. The flow chart below shows the proposed interactions between the containers and the modules which fill and use them.

trackingreorgflow

Note that most of these objects are still untested, and this pull request is meant to update the objects so that the tracking team can begin to implement their use. I've also included skeleton examples of a digitizer for the MVTX (simulation/g4simulation/g4mvtx/PHG4MvtxDigitizer.h) which converts the PHG4Cell objects to the new MvtxHit objects. Again, this has not been fully tested but should compile. Likewise, I've included clustering code for the MVTX using the new objects (offline/packages/mvtx/MvtxClusterizer.h) as an example that needs further development.


pinkenburg:
I had a brief look and a few comments from that. (sorry, the list looks a lot worse than it is)
We should create a libmvtx_io and libmvtx library, where the io objects (like mvtxhit) go into the _io library. This prevents library dependencies on some reconstruction if you only want to read a DST.
include statements like
#include "mvtx/MvtxDefs.h"
are bad and likely screw you at some point. This is a local include, use #include "MvtxDefs.h", never use relative paths with "" (if you don't have a local subdir with the include files which you just shouldn't).
Please use double math, not float. The worst thing we can do is convert between double and float all over the place, with 64 bit systems I don't think we even save time using floats. For i/o objects one can save space using floats (or half precision) and one can consider this but for any math stuff, please use double.
The PHTimeServer is in Fun4All, if you want to measure the speed of your module, please use the functionality in Fun4All for this, not implement redundant time servers in modules.
<limits.h> -> <climits> (I know that is new - I just found it)

TrkrHitSetContainer::addHitSet probably creates a memory leak if it is called with a duplicate key since the pointer is just overwritten. Also doing 
m_hitmap[key] = newhit;
return m_hitmap.find(key);
spends time searching the map again for the key (I know this kind of code is all over the place dating back to the muons I guess - we need to fix this).
One can use map.insert for this which does not overwrite an entry, one can check the return if the insertion was successful and it returns the iterator as second argument (you can probably use 
auto ret =  timeseq.insert(make_pair(tbin,edepmap)); rather than the pair<map<int,EdepMap>::iterator, bool> nobdy cares about:
// If the returned boolean from the map.insert is
// true: the element was inserted
// false: the element existed 
// in any case an iterator to the element in the map is returned
 
    EdepMap edepmap;
    pair<map<int,EdepMap>::iterator, bool> ret = timeseq.insert(make_pair(tbin,edepmap));
     map<int,EdepMap>::iterator mapiter = ret.first;

damcglinchey:
Chris, the recent commits hopefully address the points you made. Thanks for catching what should have been some obvious issues. On <climits> vs <limits.h>, I believe CINT still has some trouble with <climits>. In TrkrDefs.h I previously had to wrap <climits> so CINT ignored it. The dependencies on <limits.h> in other places were removed. 

I additionally fixed the TrkrHit and MvtxHit objects so that they conform better with the other objects. The MvtxHit doesn't need any additional class members, and is really only a concrete wrapper to TrkrHit with some convenience interfaces. 

I've also tested PHG4MvtxDigitizer and MvtxClusterizer with a quick 10 particle simulation. The simulation runs, and appears to be doing what is expected, but more testing and debugging is really needed before they get used. Again, they were mainly included to help the tracking team with development including the new objects, and are not meant to be used in prime time yet.

pinkenburg:
That was a bit premature - the compiler just complained about using ULONG_MAX to set an uint32_t variable. It should be UINT_MAX. But this puts back dependencies on the bitness of ints or longs which we tried to avoid by using uint32_t in the first place. if you want all bits set, you can use ~0x0 which is independent of the number of bits of a variable

PR #457, opened 2018-06-30
Update new tracking objects

damcglinchey:
This pull request is meant to bring the new tracking object structures up to date and allow subsystem developers to begin to implement them. 

The hitkey has been removed from TrkrHit in order to limit the storing of redundant information. This change has been propagated to the MvtxHit object. Skeleton's have also been put in place for the INTT and TPC.


offline/packages/tpc/TpcDefs.cc
```diff
@@ -0,0 +1,87 @@
...
+TpcDefs::genHitKey(const uint16_t pad, const uint16_t tbin)
+{
+  TrkrDefs::hitkey key = (pad << TpcDefs::kBitShiftPad);
+  TrkrDefs::hitkey tmp = (tbin << TpcDefs::kBitShiftTBin);
+  key |= tmp;
+  return key;
+}
+
+TrkrDefs::hitsetkey
+TpcDefs::genHitSetKey(const char lyr, const uint8_t sector, const uint8_t chip)
```
blackcathj: I assume these parameters would be `(const char lyr, const uint8_t sector, const uint8_t side)`?

offline/packages/tpc/TpcDefs.cc
```diff
@@ -0,0 +1,87 @@
...
+  TrkrDefs::hitsetkey key = TrkrDefs::genHitSetKey(TrkrDefs::TrkrId::mvtxId, lyr);
+  TrkrDefs::hitsetkey tmp = sector;
+  key |= (tmp << TpcDefs::kBitShiftSectorId);
+  tmp = chip;
+  key |= (tmp << TpcDefs::kBitShiftSide);
+  return key;
+}
+
+TrkrDefs::cluskey
+TpcDefs::genClusKey(const char lyr, const uint8_t sector, const uint8_t chip, const uint32_t clusid)
```
blackcathj: I assume these parameters would be `(const char lyr, const uint8_t sector, const uint8_t side, const uint32_t clusid)`?

pinkenburg: rather than const char lyr which is currently a uint8_t, we should use const uint8_t for this as well

damcglinchey:
Good catch. I fixed the inconsistency and now use uint8_t for all lyr function parameters.

PR #471, opened 2018-08-26
Fixes on Cherenkov simulation

blackcathj:
The Cherenkov  physics list which is used in gas and Aerogel RICH detector simulation was configured to track the Cherenkov photon before completing track the charged track within Geant4. This conflicts the assumption within `PHG4TruthTrackingAction` which split a charged track into multiple duplicates. This significant disrupts truth tracking and truth tracking on EIC detector simulations. 

This pull request should fix that. In addition, it fixes a missing variable init for `PHG4RICHSteppingAction`


nfeege:
Jinlong tested the changes for fixing the particle duplication issue. The issue is fixed!

Before applying these changes:

screen shot 2018-08-29 at 6 30 10 pm

The exact same 5 events with changing those two lines from this pull request:

screen shot 2018-08-29 at 6 30 26 pm

PR #496, opened 2018-10-03
Fix prototype2/3

pinkenburg:
The numbering in terms of rows/columns of the scintillators in the hcals is inconsistent. This pull request (which is still being worked on - so do not merge) makes the numbers consistent between inner and outer hcal for the 3rd testbeam (2nd testbeam inner hcal still needs to be looked at). The rows are unchanged, the columns go from 0-3.
Sadly this is not just a number change, while the scintillators stay the same (except their column number), the resulting towers and their energies are very different. Maybe its the lookup of calibration constants which his probably keyed to the row/column numbers

### DIFF
--- a/simulation/g4simulation/g4detectors/PHG4BlockDetector.cc
+++ b/simulation/g4simulation/g4detectors/PHG4BlockDetector.cc
@@ -4,7 +4,6 @@
 
 #include <g4main/PHG4Utils.h>
 
-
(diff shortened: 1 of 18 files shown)


pinkenburg:
All testbeam setups have now consistent numbering (4 bins in row/column, all starting at zero). It uses a faster lookup for the row/column id than the previous parsing of the volume name.

PR #502, opened 2018-10-16
install include files needed for Root6 macros

pinkenburg:
root6 macros insist on having all include files available for objects one instantiates in the macros. These ones are needed to make Fun4All_G4_sPHENIX.C and the macros it loads functional


pinkenburg:
minor change, just adding a few files to the lists of include files to be installed

PR #514, opened 2018-11-06
Tpc time window fix

adfrawley:
This pull request contains two unrelated bug fixes:

1) Fix to a bug in the g4tpc code that caused the TPC valid time window check to effectively not be made in the electron drift module. 

2) Fix to a bug found by Takahito in the INTT code. The dimensioning of one of the arrays used to construct the INTT was not increased from 4 to 8 when we switched from layers to sublayers. This caused problems for setups containing more than 4 sublayers. Fixing this bug corrected a problem reported separately by Sanghoon today. He saw bad behavior in the sublayers beyond 4. 
Before the bug fix, I reproduced Sanghoon's problem using the 6 layer setup "INTTLADDER6" in G4_Tracking.C. After the bug fix, the distribution of clusters in phi-Z space looks normal. See the before and after plots below.
[INTTLADDER6_before.pdf](https://github.com/sPHENIX-Collaboration/coresoftware/files/2551611/INTTLADDER6_before.pdf)
[INTTLADDER6_after.pdf](https://github.com/sPHENIX-Collaboration/coresoftware/files/2551612/INTTLADDER6_after.pdf)

Also fixed some minor bugs in the dimensioning of edge regions pointed out by Takahito.

### DIFF
--- a/simulation/g4simulation/g4detectors/PHG4SiliconTrackerDetector.cc
+++ b/simulation/g4simulation/g4detectors/PHG4SiliconTrackerDetector.cc
@@ -94,13 +94,12 @@ void PHG4SiliconTrackerDetector::Construct(G4LogicalVolume *logicWorld)
 
 int PHG4SiliconTrackerDetector::ConstructSiliconTracker(G4LogicalVolume *trackerenvelope)
 {
-  // We have an arbitray number of layers (nlayer_)
+  // We have an arbitray number of layers (nlayer_) up to 8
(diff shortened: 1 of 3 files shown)


blackcathj:
Thanks for the update! Suggesting three minor changes in https://github.com/adfrawley/coresoftware/pull/2 :

- Use stl::array to enforce boundary check in the future
- Read the length of TPC via geometry DST node of TPC
- Move print statement under verbosity flag checks

PR #522, opened 2018-11-21
PHG4ParticleGeneratorBase::get_mass() returns in units of GeV now

mchiu-bnl:
get_mass() was returning the mass in MeV.  Modified to return in units of GeV.


pinkenburg:
The stepping action definitely has issues, we should probably provide a stripped down example which does the minimal thing (creating a G4Hit on entry and setting the exit coordinates when the track leaves the volume, update the truth info and do our current consistency checks)  rather than starting from a specialized one. This one will create a new hit twice leaking memory right there.
I think we also don't want printouts spewing on the screen, we have already too many of those.
I will close this pull request and apply the fix for the returned mass separately

PR #529, opened 2018-12-11
Tpc sims reorganization

adfrawley:
This is misnamed - it is a reorganization of ALL tracking simulations code in which:
- Subsystem modules for detector construction, stepping action, cell reconstruction and digitization are moved to g4simulation/g4mvtx(g4intt,g4tpc) and renamed.
- Clustering code is moved to offline/packages/mvtx(intt,tpc) and renamed
- The containers (SvtxHit, SvtxCluster, SvtxTrack) are moved to offline/packages/trackbase_historic
- The library build order is adjusted in utilities/utils/rebuild/packages.txt (separate PR)
- The tracking macro is modified to use the new paths and module names (separate PR)

I made a private build and debugged the new library build dependencies and fixed all include paths. 

There are no real code changes made here, just reorganization, breaking up of some modules to separate subsystems, and extensive renaming of modules so you can figure out what does what. I ran the code from my private build and it produces what I expected to see.

An overview of what has done was presented at the sims meeting on November 27:
[Simulations_reorganization_nov27_2018.pdf](https://github.com/sPHENIX-Collaboration/coresoftware/files/2668396/Simulations_reorganization_nov27_2018.pdf)

I will make a separate PR for the macro changes and packages.txt changes. This PR will break all tracking until those changes are merged too.


sh-lim:
Hi Tony,

Thanks for updating the default timing window for MVTX. One thing I experienced to play with the timing window before is that 'set_timing_window' function is not working because a new element with the existing key value (layer index) will not be inserted into the map.

PHG4MVTXCellReco::set_timing_window
https://github.com/sPHENIX-Collaboration/coresoftware/pull/529/commits/16da5b6b4f6c57994af8079c4da8f02686fcc45b

Best,
Sanghoon

PR #543, opened 2019-01-16
Root6 fun4all

pinkenburg:
split LinkDef files, build cmd line dictionaries only for root5, install pcm files, fix cppcheck warnings


pinkenburg:
wow, that cosmetic change had more repercussions than anticipated - please everyone, use forward declarations, do not include all include files you can think of

PR #579, opened 2019-02-06
Root6 trackbase

pinkenburg:
the utils lib was just the def namespace which is needed by the io objects, no need to keep that one around. cmd line dicts for root5, i/o discts for root5/6. Finally figured out what made cling spew warnings overrides a member function but is not marked 'override' [-Winconsistent-missing-override]. When using the override keyword one has to apply this consistently, not for some random subset of methods. Need to look into how to handle this in the future


pinkenburg:
The Jenkins test ran into a known infinite loop in the tracking. I made a build from this and tested with the standard macro with fixed seed and it succeeds to run. The results are identical to the previous builds

PR #580, opened 2019-02-07
Added noise electron when determining ADC value.

tktdrk:
### Introduction

This is a pull request to implement the contribution from noise electrons into the determination of INTT hit ADC values. The noise electron distribution for 320um thick silicon cell was obtained using INTT prototype modules, which is shown in the right panel of the below figure. The mean(457e-) and sigma(166e-) values of the noise electron distribution are available only in this plot right now. Once we have a new INTT prototype module with 320um thick cell, we will confirm these values again. For now we want to set up an infrastructure to handle the noise contributions.

intt_prototype_320um

### How to implement the noise contribution
Currently the ADC value of INTT hit is determined by directly comparing the true deposited energy in a cell (cell->get_edep()) to the energy threshold (vadcrange * mip_e). 

The code change strategy here is as follows:
1.  Convert the true deposited in cell and ADC threshold energies to the number of electrons
2. Add the noise electrons to the # of electrons from the true deposited energy in cell
3. Compare # of electrons from true deposited energy+noise to the # of electrons converted from the ADC threshold 

### Result

We ran simulations with an identical random number seed for both the with and without noise cases.
The histograms are normalized at ADC=4 to compare their shapes. One can find that the distribution  with noise becomes slightly wider than that without noise.

The second plot below shows the pointing resolution of INTT clusters in phi and z direction.
Because only phi sensitive layers are going to be installed in our future detector, so are they
in this simulation. The resolution in phi direction is very sharp in both cases. The poorer resolution
in z direction is because of the length of cell in this direction. The widths before/after look quite similar.

adcwnoise

inttclusterpointres

### DIFF
--- a/simulation/g4simulation/g4intt/PHG4INTTDigitizer.cc
+++ b/simulation/g4simulation/g4intt/PHG4INTTDigitizer.cc
@@ -18,7 +18,10 @@
 #include <phool/PHCompositeNode.h>
 #include <phool/PHIODataNode.h>
 #include <phool/PHNodeIterator.h>
+#include <phool/PHRandomSeed.h>
 #include <phool/getClass.h>
+
+#include <gsl/gsl_randist.h>
 
 #include <cmath>
 #include <iostream>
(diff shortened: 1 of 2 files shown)


sPHENIX-GitHub-l:
Dear Takahito:

As you note, you will revisit this when you have a new prototype module. At that time it would be good to determine if there is a two peak structure in the noise spectrum. I think there clearly is (pedestal? + ‘physics’ noise??). Your single Gaussian fit seems to underestimate the values in the second peak, which could lead to overly optimistic performance projections.

Best regards,

Bill 

—————————————
W.A. Zajc
I.I. Rabi Professor of Physics
Columbia University
New York, NY 10027
—————————————

PR #582, opened 2019-02-07
Root6 trigger

pinkenburg:
cmd line dict for root5 only, i/o dicts for root5/6, replace loop over 2d vector entries in reset by fill. That pasta code still needs some serious work to be readable and compact


pinkenburg:
The macro failure looks odd, I didn't see this when testing but my test included the changes from trackreco which I am not sure are part of this. Let me merge this and see where the pieces fall

PR #589, opened 2019-02-12
Tktdrk intt sim

pinkenburg:
This is Takahitos PR for the INTT noise with some fixes to make it work

### DIFF
--- a/simulation/g4simulation/g4intt/PHG4INTTDigitizer.cc
+++ b/simulation/g4simulation/g4intt/PHG4INTTDigitizer.cc
@@ -21,9 +21,12 @@
 #include <phool/PHRandomSeed.h>
 #include <phool/getClass.h>
 
+#include <TSystem.h>
+
 #include <gsl/gsl_randist.h>
 
 #include <cmath>
+#include <cfloat>
 #include <iostream>
 
 using namespace std;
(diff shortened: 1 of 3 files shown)


simulation/g4simulation/g4intt/PHG4INTTDigitizer.cc
```diff
@@ -287,3 +354,48 @@ void PHG4INTTDigitizer::PrintHits(PHCompositeNode *topNode)
...
+
+void PHG4INTTDigitizer::SetDefaultParameters()
+{
+  set_default_double_param("NoiseMean", 457.2);
+  set_default_double_param("NoiseSigma", 166.6);
+  set_default_double_param("EnergyPerPair", 3.62e-9);  // GeV/e-h
+  return;
+}
+
+float PHG4INTTDigitizer::added_noise()
```
blackcathj: @todoroki ,   the noise is bi-polar, i.e. can make ths signal fluctuate up and down. Much of the `mNoiseSigma` as extracted in https://github.com/sPHENIX-Collaboration/coresoftware/pull/580 is statistical fluctuation from the limited calibration data. They does not directly apply here in generating noises.

tktdrk: For the record, here are updated ADC and pointing resolution plots with the correct bi-polar Gaussian. Now we see stronger effects in the small ADC range as expected.

inttadc

inttpointres

PR #594, opened 2019-02-27
Eic physics list

pinkenburg:
This PR adds the proposed EIC G4 physics list. It covers the same energy range interesting for sPHENIX. It is experimental but we should have a close look at it


pinkenburg:
The build fails in online. The eic physics list is nowhere exercised but the new entry in package.txt will crash all  builds which do not contain this package. Merging

PR #598, opened 2019-03-05
Modified diffusion calculation for MVTX and set a single ADC threshol…

adfrawley:
…d value. Both diffusion and threshold can be set from macro. The default values in the code are 22 microns and 0.95 keV, respectively. Thers are optimized to give a pixel multiplicity per ccluster of 2.4. The MVTX cluster resolution looks good:
[cluster_resolution_z.pdf](https://github.com/sPHENIX-Collaboration/coresoftware/files/2933003/cluster_resolution_z.pdf)
[cluster_resolution_rphi.pdf](https://github.com/sPHENIX-Collaboration/coresoftware/files/2933004/cluster_resolution_rphi.pdf)


blackcathj:
Thanks Tony!
For future reference, see discussion in the simulation meeting talk: https://indico.bnl.gov/event/5454/

PR #601, opened 2019-03-15
INTT new endcap ring geometry using Carbon-PEEK

tktdrk:
This is a PR to implement the new INTT endcap ring geometry using Carbon-PEEK (CP) proposed by Rachid at the simulation meeting 3/12/2019. Defined the CP material in the "PHG4Reco.cc". Introduced a parameter in the "PHG4INTTSubsystem.cc" to select either the aluminum(Al) or CP endcap rings. The Al endcap ring option remains as default for now. The new geometry in "PHG4INTTDetector.cc"  successfully passed the overlap check.

Ring  EC CP


sPHENIX-GitHub-l:
Dear Takahito,

Could you make carbon peek as default setting in simulation and aluminum end rings as option. I am concern about next set of simulation,

Thanks a lot.
Rachid

PR #617, opened 2019-04-10
Fix g4dst

pinkenburg:
This PR updates the libraries which need loading when opening a DST. The g4mvtx package had the sources mixed, the mvtx_geo ended up in libg4mvtx rather than in libg4mvtx_io. Added test linking of both libs, not just libg4mvtx which would have picked this up


pinkenburg:
skipping the vagrind check result, this is just a lib dependency change. I want to re-enable the build cron jobs which bomb out with the change in packages.txt to put the building of libg4dst last

PR #624, opened 2019-04-20
TrkrHitSetContainer::Reset() did not properly free memory allocated t…

adfrawley:
…o TrkrHit's. Fixed by calling TrkrHitSet::Reset().
TrkrHitSet::Reset() frees the TrkrHit memory before erasing the entries.
I also added a destructor that calls the Reset() method.

### DIFF
--- a/offline/packages/trackbase/TrkrHitSet.cc
+++ b/offline/packages/trackbase/TrkrHitSet.cc
@@ -55,6 +55,27 @@ void TrkrHitSet::identify(std::ostream& os) const
   }
 }
 
+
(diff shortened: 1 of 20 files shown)


blackcathj:
Multiple additional fixes have been introduced to the tracking code:

- virtual destructor for `TrkrHitSet`
- remove hits cleaning from  `TrkrHitSet`
- memory leak in `SvtxVertexMap`
- removing `cluster_key != 0` check in evaluator as this key is can be a meaningful cluster now.

PR #509, opened 2018-10-30
TPC with a rigid central membrane

blackcathj:
Implementing a TPC membrane in a sandwich type structure to evaluate for a rigid central membrane design. The structure can be configured via parameters in TPC macro. The default parameter is FR4 surface of 0.0125 cm thickness sandwich a core of ROHACELL FOAM51. The overall thickness is 0.56 cm. 

This is under evaluation by Sourav with 1st report in [today's simulation meeting](https://indico.bnl.gov/event/4039/)

## Geant4 dispaly of the default membrane in this pull request

Central membrane in TPC with half cut:
tpc2

Side view of TPC with ROHACELL FOAM51 shown in white:
tpc

Zoom-in side view of TPC with ROHACELL FOAM51 shown in white sandwitched by FR4 in a thin blue layer: 
tpc1

PR #629, opened 2019-05-01
Add two layer TPC central membrane surface material with Cu+PCB

blackcathj:
Tom suggested the original design of Al coat on the central membrane may flake off over time. Therefore, we were requested to evaluate a thin-Cu-coated membrane that may double the radiation length (0.14% -> 0.26%). And thanks to Sourav, who plan to help test its impact to occupancy and tracking performance. 

Follow up with #509, this pull request introduce possibility of using two layer as the surface of the central membrane. The default option is 8.9 us of Cu and 127um of PCB (FR4). The core material has not been changed which is ~5.6 mm of rohacell.


blackcathj:
Sourav completed the performance study, summary will be discussed in future simulation meeting. Merging this pull request first.

PR #643, opened 2019-05-22
Decaps intt mvtx tpc

pinkenburg:
This PR makes the naming of the Intt, Mvtx and Tpc uniform. We had a mixture of all caps and mixed caps-lower case names which made maintenance tedious and really confusing. Form now on only the node names are all caps, everything else (source names and variables) is mixed case.


pinkenburg:
The valgrind failure needs an updated suppression file (the lower caps methods weren't propagated to the suppression file). The EIC macro will be fixed after building the code

PR #644, opened 2019-05-23
Truth tracking

adfrawley:
Debugging and implementing PHTruthTrackSeeding was overlooked in the upgrade to the new containers. It is now debugged and implemented. There is an accompanying modification to G4_Tracking.C in a separate PR.


blackcathj:
Problems reported in check are known from https://github.com/sPHENIX-Collaboration/coresoftware/pull/643, which are being fixed. Merging

PR #660, opened 2019-06-12
Configurable seeding

bogui56:
This pull request contains settesrs for the SvtxClusterMap name and configurable seeding layers as preparations for the iterative tracking. The default seeding configuration was changed from 4 out of 7 layers to 4 out of 12 layers. This should temporarily the tracking efficiency to roughly what it was before the switch to the modular tracking.

eff_eval_all_nopu_pt


bogui56:
Fixed merge conflicts and updated the SvtxEvaluator to do what it actually should do. In some cases it was still using the cluster arrays and iters instead of the new cluster_key arrays and iters.

pinkenburg:
I suppressed the valgrind error, it may be a static leak but this can be fixed later

PR #663, opened 2019-06-19
Fix cppcheck

pinkenburg:
This PR takes care of a handful of cppcheck warnings. We had some duplication of methods in PHG4Cell and PHG4CylinderCell and the invalid iterators from the shower/hit collections should be fixed now


pinkenburg:
I updated the valgrind suppression file with yet another static leak from the tracking

PR #666, opened 2019-06-21
remove g4hough remnants

pinkenburg:
This PR removes the g4hough from the Makefiles.  The only code which used a method from g4hough is offline/QA/modules/QAG4SimulationCalorimeterSum.cc for projections to the front of the emc. I commented this code out - that needs to be reimplemented with our current versions.


pinkenburg:
A macro used by the EIC tracking was removed (G4_Svtx_maps_ladders+intt_ladders+tpc_KalmanPatRec_EIC.C)
The valgrind error is another static leak which needs tracking down or suppressing. Merging this so the daily rebuilds don't crash since g4hough is not build anymore

PR #667, opened 2019-06-25
undo revert

bogui56:


bogui56:
reverting back the revert by chris. this should put the new seeding defaults back into the repo

PR #671, opened 2019-07-04
added deltapt, deltaeta, and deltaphi

zchwaknn:
added deltapt, deltaeta, and deltaphi

### DIFF
--- a/simulation/g4simulation/g4eval/SvtxEvaluator.cc
+++ b/simulation/g4simulation/g4eval/SvtxEvaluator.cc
@@ -161,7 +161,7 @@ int SvtxEvaluator::Init(PHCompositeNode* topNode)
                                                  "nhittpcall:nhittpcin:nhittpcmid:nhittpcout:nclusall:nclustpc:nclusintt:nclusmaps");
 
   if (_do_track_eval) _ntp_track = new TNtuple("ntp_track", "svtxtrack => max truth",
-                                               "event:trackID:px:py:pz:pt:eta:phi:charge:"
+                                               "event:trackID:px:py:pz:pt:eta:phi:deltapt:deltaeta:deltaphi:charge:"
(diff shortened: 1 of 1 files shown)


simulation/g4simulation/g4eval/SvtxEvaluator.cc
```diff
@@ -2501,6 +2501,15 @@ void SvtxEvaluator::fillOutputNtuples(PHCompositeNode* topNode)
...
         float eta = v.Eta();
         float phi = v.Phi();
+	float CVxx = track->get_error(3,3);
+	float CVxy = track->get_error(3,4);
+	float CVxz = track->get_error(3,5);
+	float CVyy = track->get_error(4,4);
+	float CVyz = track->get_error(4,5);
+	float CVzz = track->get_error(5,5);
+	float deltapt = TMath::Sqrt((CVxx*px*px+2*CVxy*px*py+CVyy*py*py)/(px*px+py*py));
+        float deltaeta = TMath::Sqrt((CVzz*(px*px+py*py)*(px*px+py*py)+pz*(-2*(CVxz*px+CVyz*py)*(px*px+py*py)+CVxx*px*px*pz+CVyy*py*py*pz+2*CVxy*px*py*pz))/((px*px+py*py)*(px*px+py*py)*(px*px+py*py+pz*pz)*(px*px+py*py+pz*pz)));
```
pinkenburg: Please do not use TMath - TMath has very strange ideas how to handle bad values

zchwaknn: I think I tried compiling without TMath::, but it gave an error. What should be used instead?

blackcathj: Good to try stl `sqrt()` with http://www.cplusplus.com/reference/cmath/sqrt/

PR #672, opened 2019-07-05
Pull

zchwaknn:
image


blackcathj:
Thanks for the immediate follow up! 

For the eta and phi pull distribution plots, could you also zoom into the x-axis range of [-5, 5]? The goal is to check the width of the Gauss core to be near 1.

zchwaknn:
image

blackcathj:
@zchwaknn  Is this the identical plot as in the 1st comment?

zchwaknn:
I was working in the same windows that were just generated, clicked on the X axes, right-clicked, and selected SetLimits. I entered -5 and 5.

blackcathj:
Please try something like this in the drawing stage:
```c++
ntp_track->Draw("(pt-gpt)/detapt >> hEtaPull(100,-5,5)")
```

zchwaknn:
image
image

PR #679, opened 2019-08-09
Outer tracker

adfrawley:
Added a detector, OuterTracker, consisting of an arbitrary number of silicon layers outside the TPC. It is completely configurable from the macro. By default the macro will be set up for 2 layers, each 100 microns thick, with a pixel segmentation that gives about 100 micron cluster resolution. It is added as a complete subsystem, including everything through clustering, and the evaluator has been modified to include the hits and clusters. It is not added to the tracking, since it is intended for space charge distortion correction studies.
There will be an accompanying pull request for the macro.


pinkenburg:
I don't think this should become part of our code base. It is basically a copy of the mvtx code with some minor changes. This will just add to maintenance load without a benefit. Given that this detector will never be build (20m^2 of silicon are just out of the question), it is not clear why a detailed silicon implementation is even necessary to achieve a space point measurement which we can generate using existing tools for cylinder detectors (~pinkenbu/workarea/sPHENIX/gitrepov5/macros/macros/OuterTracker/Fun4All_G4_OuterTracker.C creates 100um cell sizes for 2 silicon cylinders at 82/84cm). I would suggest to build on the generic tools (add to them the charge sharing as module if really needed) and create an interface module which copies the cells into the tracking scheme. This will allow us to reuse this in other places if needed (e.g. detector concepts at the eic). Alternatively one could generalize the maps implementation so this code can be reused rather than just copying it.

The G4_Tracking.C macro is becoming unwieldy, rather than adding more and more lines to the G4_Tracking.C macro we should see that we split it up along tracking detectors, just like the G4Setup macro. At least this detector just adds a few functions which can be declared in a separate macro.  This will make adjustments needed for our other implementations (fsPHENIX, EIC) a lot easier to maintain.

In general - it would be advantageous to move away from the hardcoded tracker object nodes. For other systems it turned out to be helpful to be able to run two implementations simultaneously and one only had to change the detectorname to get a fully functional chain. Replacing TRKR by a string which can be configured in the macro is a solution to this.

adfrawley:
Hello Chris,

It seems likely that there will be a tracking detector outside the TPC of some sort. This commit was not intended to make a realistic detector, obviously, but to create a place for an outer tracking detector in g4simulations and offline/packages.

The initial implementation of magically supported layers of silicon was just an easy way to gain the ability to do some tests, needed to evaluate the effect of having space points with a specified resolution outside the TPC. That can be replaced easily enough with something that looks like an actual detector at the appropriate time. I agree that what is there now could have been done with a cylinder-cell model. That was not my purpose.

Cheers
Tony

blackcathj:
New micromegas outter tracker implementation supersedes this PR. Closing.

PR #687, opened 2019-08-19
Move init vertexing

adfrawley:
Moved Sookhyun Lee's initial vertexing code (using the MVTX) from g4hough into offline/packages/trackreco and modified it to use the new Tracking containers. I removed some utility classes that were copies of classes from HelixHough, and switched to using the HelixHough versions.
I implemented the PHInitZVertexing module as an alternative to truth vertexing (there will be an accompanying macro pull request).
I tested the tracking with the initial vertex obtained from PHInitZVertexing using Pyhia events, and with 10, 20 and 500 pion events. It works very well.
There is only one user parameter, the minimum number of associated tracks required for a candidate vertex to be accepted. Setting it to 2 for p+p and 5 for Au+Au seems to work well. If it is set too low, you get a lot of spurious vertices. However I modified PHHoughSeeding so that it chooses the vertex with the largest number of associated tracks, so spurious vertices are ignored downstream.
Note that the tracking does not yet support events with multiple valid vertices, that will be the next step.


blackcathj:
@adfrawley the ROOT5 build failure lead me to found a few problem using the `CellularAutomaton`. For a start, it is not a storage object and current use drag transient object and Eigen into the io lib. And its use in vertex and tracking leads to memory leaks. 

Pushed a few commit trying to fix both. However, I have not tried in running tracking yet. Please give a try.

blackcathj:
@adfrawley with the new macro used in the test: 
https://github.com/sPHENIX-Collaboration/macros/pull/199 
It appear the code crash at the end of single particle simulation, and it produce a warning of no-vertex found at each event: 
https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-default-pipeline/3145/console
It will be great to modify the code to also tolerate single particle simulation where vertex would be missing.

PR #697, opened 2019-09-28
Enable Volume hierarchy

pinkenburg:
This PR enables hierarchies of volumes on the macro level (e.g. cylindrical magnetic field inside an iron cylinger == beam magnet). A change in the PHG4Detector and PHG4Subsystem base class makes this transparently possible. The main change in the detector classes is that the respective PHG4Subsystem needs to be added to the PHG4Detector ctor and the Construct() method is now ConstructMe(). The Construct() method is now implemented in PHG4Detecor and declared final. ConstructMe is now called for the detector implementations, the G4LogicalVolume is set to the mother volume (if set)

### DIFF
--- a/offline/QA/modules/QAG4SimulationCalorimeter.cc
+++ b/offline/QA/modules/QAG4SimulationCalorimeter.cc
@@ -15,7 +15,6 @@
 
 #include <g4eval/CaloEvalStack.h>
 #include <g4eval/CaloRawClusterEval.h>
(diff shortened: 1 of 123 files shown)


pinkenburg:
Before I forget - this change was tested with our three macros (Fun4All_G4_sPHENIX.C, Fun4All_G4_fsPHENIX.C and Fun4AllG4_EICDetector.C), the results are not affected. This did not test all detectors we have but problems will be very evident (no detector or segfault)

PR #705, opened 2019-10-11
Simulating TPC GEM gain fluctuation

blackcathj:
Implement per-electron TPC GEM gain fluctuation using an exponential PDF as suggested by Tom and Bob. 

The average gain is not changed (x2000, now also in `phparameter`), but we could expect more realistic cluster amplitude-shape fluctuations. This was used in SAMPA FEE input charge simulation (https://github.com/sPHENIX-Collaboration/tpc-fee-input-charge) for FEE development. 

Request tracking team to validate its impact to tracking resolution and occupancy prior to merging. @adfrawley @bogui56 . This added fluctuation could lead to a reduction of the arbitrary resolution fudge factors in the TPC macro. 

## Example simulated FEE input charge fluctuations

Click to enlarge. More in https://github.com/sPHENIX-Collaboration/tpc-fee-input-charge

Single track: 

PadCharge_side0_sector1_layer40

Single 0-4fm central AuAu collision track: 

PadCharge_side0_sector1_layer33

Typical higher occupancy scenario with 100kHz collisions: 

PadCharge_side0_sector1_layer40


adfrawley:
This modification worsens the TPC cluster resolution significantly. Therefore I retuned the TPC cluster resolution fudge factors so that the r-phi cluster resolution in the outer TPC becomes 150 microns again. The new fudge factors are (r-phi) 0.085 and (Z) 0.105 (reduced from 0.12 and 0.15). I will make a pull request with these values changed in the macro. With the new fudge factors, the upsilon resolution is 88 MeV in 100 pion events. The efficiency is unchanged by this pull request. It is ready to be merged.

PR #715, opened 2019-12-05
Added minimum number of constituents option to PYTHIA jet trigger

jdosbo:
Added simple functionality to the PYTHIA jet trigger class for users to require the jet has a minimum number of constituents. Tested by running default pythia jet macro while requiring a large number of constituents (14) at RHIC energies - each event had at least one jet with 14 or more constituents.

Useful for e.g. photon-jet events where one would like to trigger on a high pT photon and a high pT jet that isn't just composed of (for example) the prompt photon.

### DIFF
--- a/generators/PHPythia6/PHPy6JetTrigger.cc
+++ b/generators/PHPythia6/PHPy6JetTrigger.cc
@@ -2,20 +2,20 @@
 #include "PHPy6GenTrigger.h"
 
 #include <HepMC/GenEvent.h>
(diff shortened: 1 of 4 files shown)


blackcathj:
There was a Jenkins-CI hang last night. Please ignore the first two reports. @jdosbo

jdosbo:
Okay, I tested locally and it worked fine so just let me know if somehow the build doesn't work and I'll take a closer look.

PR #717, opened 2019-12-10
Enable printing of random seeds for rare problem debugging

blackcathj:
Following the discussion in the last meeting, enable printing of random seeds for rare problem debugging, such as the on-going debug for rare non-initialized variable in the tracking evaluator.


blackcathj:
Sure. As in https://github.com/sPHENIX-Collaboration/macros/pull/217

PR #723, opened 2020-01-07
Adding RTree seeding

bogui56:
First implementation of the rtree seeding


pinkenburg:
It is an include which makes root5 rootcint barf. It's easier to debug this after it's merged

PR #728, opened 2020-01-14
add include to resolve phgenfit::Track

pinkenburg:
This should take care of the  weird behavior under singularity with local cvmfs copies. phgenfit::Track was forward declared but used in a shared pointer. No idea why it behaved this way but including the declaration fixed this. This include has a comment now to ignore the recommendations from include checkers to remove this


pinkenburg:
skipping the wait for Jenkins to get this patch for the ubuntu crashes out asap

PR #733, opened 2020-01-16
"Accumulate modifications for new MVTX geometry and a lot of cleaning up"

ycorrales:
Add a class inherited  from ALICE with real chip and sensor size
Implement new MVTX stave model and layer configuration, including rotation offset for the first stave
Remove much-unused code referring to ALICE OB detector.
Fix few bugs in the algorithm to determine the pixel index to/from location in the sensor.


pinkenburg:
I fixed the root5 crash and submitted another PR with this code, closing this one

PR #737, opened 2020-01-17
Adding CA Seeding module

petermic:
Adds PHCASeeding class (and supporting ALICE code, which starts with GPUTPC for now) to the trackreco package. This class can be directly substituted for PHHoughSeeding or PHRTreeSeeding where they appear.


blackcathj:
@petermic Thanks for the contribution! 
There is few `cppcheck` warnings here: 
https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/cpp-check-pipeline/917/cppcheck/New/#issuesContent

PR #741, opened 2020-02-17
Acts implementation

adfrawley:
Adds module PHActsTrkFitter to offline/packages/trackreco. Includes accompanying changes to packages/mvtx and packages/intt. Implements Acts geometry building from within Fun4All, and creates Acts measurements for the sPHENIX MVTX, INTT and TPC. Built and tested against the new build.


blackcathj:
This is a bit tricky as the ACTS package is not built in the nominal builds, but require gcc-8.3 configuration.  It will be useful we have a chat how to best proceed without breaking the daily build in the simulation meeting tomorrow.

PR #742, opened 2020-02-20
Add EMCal layer in 3-D topoClustering

dvperepelitsa:
3-D topoCluster building can now extend across IHCal<->EMCal boundary. Configurable user options to run topoClustering on HCal layers only, EMCal only, or all three layers.

### DIFF
--- a/offline/packages/CaloReco/RawClusterBuilderTopo.h
+++ b/offline/packages/CaloReco/RawClusterBuilderTopo.h
@@ -15,6 +15,24 @@
 class PHCompositeNode;
 class RawClusterContainer;
 
+int RawClusterBuilderTopo_constants_EMCal_eta_start_given_IHCal[24] = 
+  {2,  6,  10, 14, 18, 22, 26, 30, 33, 37, 41, 44, 
+   48, 52, 55, 59, 63, 66, 70, 74, 78, 82, 86, 90 };
+
+int RawClusterBuilderTopo_constants_EMCal_eta_end_given_IHCal[24] = 
+  {5,  9,  13, 17, 21, 25, 29, 32, 36, 40, 43, 47,
+   51, 54, 58, 62, 65, 69, 73, 77, 81, 85, 89, 93 };
+
+int RawClusterBuilderTopo_constants_IHCal_eta_given_EMCal[96] = {
+  -1, -1,  0,  0,  0,  0,  1,  1,  1,  1,  2,  2, 
+   2,  2,  3,  3,  3,  3,  4,  4,  4,  4,  5,  5, 
+   5,  5,  6,  6,  6,  6,  7,  7,  7,  8,  8,  8, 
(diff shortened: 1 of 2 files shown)


offline/packages/CaloReco/RawClusterBuilderTopo.h
```diff
@@ -46,42 +65,185 @@ class RawClusterBuilderTopo : public SubsysReco
...
+    if ( ID < 256 * 96 ) return _TOWERMAP_E_LAYER_ETA_PHI[ get_ilayer_from_ID( ID ) ][ get_ieta_from_ID( ID ) ][ get_iphi_from_ID( ID ) ];
+    else return _EMTOWERMAP_E_ETA_PHI[ get_ieta_from_ID( ID ) ][ get_iphi_from_ID( ID ) ];
+  }
+
+  void set_status_by_ID( int ID , int status ) {
+    if ( ID < 256 * 96 ) _TOWERMAP_STATUS_LAYER_ETA_PHI[ get_ilayer_from_ID( ID ) ][ get_ieta_from_ID( ID ) ][ get_iphi_from_ID( ID ) ] = status;
+    else _EMTOWERMAP_STATUS_ETA_PHI[ get_ieta_from_ID( ID ) ][ get_iphi_from_ID( ID ) ] = status;
   }
+
+  std::vector<int> get_adjacent_towers_by_ID( int ID ) {
```
blackcathj: Probably just need to move the function implementations to the cc file to pass the `ROOT5` build.

offline/packages/CaloReco/RawClusterBuilderTopo.h
```diff
@@ -15,6 +15,24 @@
 class PHCompositeNode;
 class RawClusterContainer;
 
+int RawClusterBuilderTopo_constants_EMCal_eta_start_given_IHCal[24] =
```
blackcathj: It will be really nice if the numbers come from the geometry object or a calibration file  ;-)

blackcathj:
Please ignore the last Jenkins report. It was a git checkout issue with our new ACTS setup.

osbornjd:
Just FYI CERNs gitlab is down today, which is why the git checkout for ACTS is hanging.

offline/packages/CaloReco/RawClusterBuilderTopo.h
dvperepelitsa: Done, this has been moved to the .cc file.

offline/packages/CaloReco/RawClusterBuilderTopo.h
dvperepelitsa: Where possible, the total eta/phi dimensions are now being taken from the geometry during run-time. However, the particular mapping on this line is difficult to calculate in real time (since the EMCal has a different eta segmentation which even varies with |eta|) and is thus hard-coded at the moment.

PR #743, opened 2020-02-20
Fix nullptr check

adfrawley:
Now aborts if layergeom not found!


blackcathj:
Same CERN gitlab server issue as #742 . Restarting CI tests

PR #744, opened 2020-02-20
Upstream/master genfit trk fitter disable layers

hupereir:
This pull request adds the possibility to disable some detector layers when fitting the track using genfit. Track states resulting from the fit are then extra/interpolated back to the disabled layers in order to make it possible to look at unbiased residuals in the said layers between track and cluster.

Also included are some cleanup of the code. Unfortunately some unrelated changes sneaked in about cleaning empty lines, or replacing "tab" characters by actual space characters, something which my editor does automatically. 

If requested, I can put those in a separate pull request.

### DIFF
--- a/offline/packages/trackreco/PHGenFitTrkFitter.cc
+++ b/offline/packages/trackreco/PHGenFitTrkFitter.cc
@@ -107,7 +107,7 @@ namespace genfit { class AbsTrackRep; }
 #define LogError(exp) std::cout << "ERROR: " << __FILE__ << ": " << __LINE__ << ": " << exp << std::endl
 #define LogWarning(exp) std::cout << "WARNING: " << __FILE__ << ": " << __LINE__ << ": " << exp << std::endl
 
-#define WILD_FLOAT -9999.
+static constexpr float WILD_FLOAT = -9999;
 
 #define _DEBUG_MODE_ 0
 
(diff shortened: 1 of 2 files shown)


blackcathj:
Hi Hugo,

Thanks for the contribution! 

The CI error reports are related to : 
1. A memory loss in track output: https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-default-valgrind-pipeline/1443/valgrindResult/pid=200850,0x2e48c/
2.  `ROOT5 CINT` is allergic to modern C++ again: https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/Build-Master/2155/artifact/build/root5/rebuild.log/*view*/ . I think we could hide these lines from ROOT5 with something like `#if !defined(__CINT__) || defined(__CLING__)`

hupereir:
> Hi Hugo,
> 
> Thanks for the contribution!
> 
> The CI error reports are related to :
> 
>     1. A memory loss in track output: https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-default-valgrind-pipeline/1443/valgrindResult/pid=200850,0x2e48c/
> 
>     2. `ROOT5 CINT` is allergic to modern C++ again: https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/Build-Master/2155/artifact/build/root5/rebuild.log/*view*/ . I think we could hide these lines from ROOT5 with something like `#if !defined(__CINT__) || defined(__CLING__)`

Both should be fixed now. 
Hopefully we will abandon root5 soon :)

hupereir:
So looking at the log, it seems I really have no choice but getting rid of std::unique_ptr, since root5 cint requires a copy constructor for the class, which unique_ptr forbid. 
The alternative would be to use some shared-ptr. This adds some overhead but is still safe in terms of making sure all instances are properly deleted, without manually writing the class destructor. 
What would you recommand ?

PR #746, opened 2020-02-21
Clusterizer

shura95:
Overall reorganization to have the only interface RawClusterBuilderTemplate for all calorimeter clustering purposes (and get rid of any specific interfaces like RawClusterBuilderTemplateFEMC etc.). All detector specific code is now included in BEmcRec<CalName> modules, which now exist for CEMC, FEMC and EEMC. It includes energy and position correction and shower profile evaluation ("prob" variable). If BEmcRec<CalName> is not defined for a particular calorimeter <CalName>, clustering will be called through the base class BEmcRec, which doesn't correct energy for any effects (e.g. Ecore), position (just give a straight center of gravity), and doesn't evaluate shower profile parameter. In addition: 
1. BEmcProfile is slight modified to include additional uncertainty due to binning used in profile hists
2. BEmcCluster  is slight modified to include more arguments in BEmcRec->GetProb() call, needed to evaluate the impact angle
3. RawClusterBuilderTemplate<CalName> modules are not yet removed for safety. I'll do it in the next pull request session

### DIFF
--- a/offline/packages/CaloReco/BEmcCluster.cc
+++ b/offline/packages/CaloReco/BEmcCluster.cc
@@ -524,7 +524,10 @@ void EmcCluster::GetErrors(float* pde, float* pdx, float* pdy, float* pdz)
 
 float EmcCluster::GetProb(float& chi2, int& ndf)
 {
-  return fOwner->GetProb(fHitList, chi2, ndf);
+  float e, xg, yg, zg;
+  e = GetTotalEnergy();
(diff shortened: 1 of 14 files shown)


dvperepelitsa:
Hi @shura95 , may I ask how much of this PR is a just a change in the organization / front-end, and how much of it actually modifies the cluster reconstruction, prob assignment, etc.? Thanks!

PR #759, opened 2020-03-17
Acts Track Fitting

osbornjd:
This PR introduces the necessary foundation for calling the Acts track fitting using sPHENIX clusters and track seeds. There is still work to be done as the Kalman Fitter within Acts can't find the TrackingGeometry object (unsure why at the moment), but I'm pushing this code now so that there is a stable copy in the master repository that can successfully call the track fitting without any sPHENIX issues.

Still on the table is debugging the issues within Acts as to why the TrackingGeometry object can't be accessed.


blackcathj:
Please ignore the error which due to on-going work with tracking-QA integration into Jenkins.

PR #764, opened 2020-03-29
Acts Fitting Modularization

osbornjd:
This PR takes most of the logic out of PHActsTrkFitter and splits it into more modular classes. Each class now puts relevant information on the node tree for the next class to grab. This is better than the previous implementation as each stage of the Acts fitting process is now put on the node tree, such that other modules may be designed which take advantage of the various steps in the Acts fitting process. The ACTFW specific classes, TrkrClusterSourceLinks.hpp and TrkrClusterFittingAlgorithm.hpp, were moved into the ACTFW repository. This is where they belong since they don't depend on sPHENIX software at all; additionally, this improves the compile/link time of the trackreco package.

The design can be discussed in the simulations meeting, if necessary.


blackcathj:
Appears to be `gcc-8` compile error towards the end:
https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/Build-Master/2488//artifact/build/new/rebuild.log/*view*/

And a performance warning: 
https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/cpp-check-pipeline/1104//cppcheck/source.1276b010-afeb-4538-86a5-c089ebae0040/#37

osbornjd:
I think the clang formatting moved some include header files around which messed some Acts things up. I'll take a look

osbornjd:
Compiles locally now after fixing the include file ordering

PR #769, opened 2020-04-01
Add phlog

pinkenburg:
this PR adds a wrapper for log4cpp


pinkenburg:
Sorry - I hit merge without thinking so jenkins did not start up. It is just installing a standalone include, no code changes

PR #772, opened 2020-04-04
make printouts more visible, adding method to printout

pinkenburg:
The previous printouts drowned in other messages in the dryrun macro. More visible and explicit now


pinkenburg:
This is just a script change which does not affect the running. I want this in the next build which kicks off in an hour - so merging this prematurely

PR #786, opened 2020-04-21
Add beamline magnets

pinkenburg:
This PR adds more sophisticated beam line magnets than the PHG4BeamLineMagnet class


blackcathj:
Please ignore `CI-build-x8664_sl7-new-FAIL` which comes from high occupancy tracking QA runs overtime, and it is due to many jobs running on the server today. Extending time-out cut offs.

PR #789, opened 2020-04-21
Evaluator truth clustering

adfrawley:
Added ntp_g4cluster to the tracking evaluator output. It starts from all g4particles (including secondaries), clusters the g4hits to make all truth clusters, and associates them with reco clusters (where those were found). It estimates the z-size and phi-size of the truth clusters. Fixed a mismatch between  the cluster size recorded for the TPC in TpcClusterizer and those recorded for the silicon detectors, so that they are now consistent. Previously TrkrCluster returned a factor of 2 too large size for TPC clusters.
This is intended to do what ntp_g4hit used to do, before we moved to a single volume TPC. The truth clusters and reco clusters are now directly comparable. It should make it possible to extract the clustering efficiency easily. 
Note that ntp_cluster also clusters the truth hits for calculating truth positions of reco clusters (same algorithm) but that is only done for g4hits that are associated with a reco cluster.


blackcathj:
Please ignore this `CI-build-gcc-8.3-new-FAIL` as it is going through tuning for #794

PR #792, opened 2020-04-27
reject HCal sub-cluster seeds that overlap with EMCal seed

dvperepelitsa:


dvperepelitsa:
The motivation and impact of this PR was discussed in the 27 April 2020 Jet Structure TG meeting, link to agenda here: https://indico.bnl.gov/event/8369/ . In summary, showers from high-energy hadrons were found to frequently have a local maximum (which seeds sub-clusters in the splitting procedure) in both the EMCal and OHCal, which resulting in splitting the cluster longitudinally. This PR fixes such cases.

PR #793, opened 2020-04-29
Genfit lib64

pinkenburg:
This PR adds $OFFLINE_MAIN/lib64 to the AM_LDFLAGS for packages linking against genfit. The default location of libgenfit2.so has changed from $OFFLINE_MAIN/lib to $OFFLINE_MAIN/lib64. Depending on the build type it is one or the other. With this change both locations are searched for libgenfit2.so


blackcathj:
Jenkins service was going through an update during the last check, which leads to failures. Please ignore its last report. 

Restarting with https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/sPHENIX_CoreSoftware_PullRequest/

PR #802, opened 2020-05-04
Track covariance

osbornjd:
This PR fixes the track covariance matrix rotation to the proper basis that Acts expects. Previously we had a place holder to simply calculate the diagonal elements from the sPHENIX covariance matrix and hand them to Acts. This PR actually rotates the covariance matrix from global (sPHENIX) coordinates to local (Acts) coordinates, and in the proper basis that Acts expects.


blackcathj:
Oops. The tracking QA fails due to merging of https://github.com/sPHENIX-Collaboration/macros/pull/236 minutes before the QA checks for this PR starts....

PR #803, opened 2020-05-04
Track propagation

osbornjd:
This PR updates the track propagation module for calling the Acts::CombinatorialKalmanFilter. The CKF was added by the Acts developers recently, so we can now take advantage of it similarly designed classes in the ACTFW that construct a CKF and call it given a track seed and a list of source links (clusters) in an event. This PR introduces code to do the minimal amount of work to call the CKF. It still needs to be tested that the CKF returns sensible results; however, this PR puts the code in the master repository so that others can start to play with it as well.


blackcathj:
The failed check appears due to time out as server load was high. Time-out limit was extended

PR #805, opened 2020-05-06
Upstream/master qa clusters

hupereir:
This PR adds a #clusters/layer/track QA histogram in the tracking module, as shown at https://indico.bnl.gov/event/7413/contributions/37321/attachments/27916/42843/talk.pdf slide 10.
The actual normalization to the number of tracks is performed in the plotting macro based on the "Normalization" histogram. 

Resulting plot:
G4sPHENIX root_qa root_QA_Draw_Tracking_nClus_Layer_QAG4SimulationTracking

@blackcathj For consistency with other histograms, I have labelled the y axis "nHit". However these are really clusters, while hits are "something else" (same is true for the other histograms). Is it ok to relabel them all to nCluster (either in this PR or a separate one) ? 

Also performed some cleanup in the tracking module.

### DIFF
--- a/offline/QA/modules/QAG4SimulationTracking.cc
+++ b/offline/QA/modules/QAG4SimulationTracking.cc
@@ -33,6 +33,7 @@
 #include <TString.h>
 #include <TVector3.h>
 
+#include <array>
 #include <cassert>
 #include <cmath>
 #include <iostream>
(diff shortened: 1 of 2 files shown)


blackcathj:
Thanks! @hupereir 

> For consistency with other histograms, I have labelled the y axis "nHit". However these are really clusters, while hits are "something else" (same is true for the other histograms). Is it ok to relabel them all to nCluster (either in this PR or a separate one) ?

Good point. We should fix it for all plots. Would you like to append the fix to this pull request? That affects both coresoftware and macros. Thanks!  

Also thinking more on your plot of truth track -> nCluster count per layer plot: although we are not running truth tracking in the QA tests, the truth track -> nCluster matching may be a compact section of code that can be part of `QAG4SimulationTracking.cc` as a new check processing function. It can be copied from here: `PHTruthTrackSeeding::Process(PHCompositeNode* topNode)` [line 68-176](https://sphenix-collaboration.github.io/doxygen/d4/d0c/PHTruthTrackSeeding_8cc_source.html#l00068)

hupereir:
> Thanks! @hupereir
> 
> > For consistency with other histograms, I have labelled the y axis "nHit". However these are really clusters, while hits are "something else" (same is true for the other histograms). Is it ok to relabel them all to nCluster (either in this PR or a separate one) ?
> 
> Good point. We should fix it for all plots. Would you like to append the fix to this pull request? That affects both coresoftware and macros. Thanks!
> 
Will do in these PR yes

> Also thinking more on your plot of truth track -> nCluster count per layer plot: although we are not running truth tracking in the QA tests, the truth track -> nCluster matching may be a compact section of code that can be part of `QAG4SimulationTracking.cc` as a new check processing function. It can be copied from here: `PHTruthTrackSeeding::Process(PHCompositeNode* topNode)` [line 68-176](https://sphenix-collaboration.github.io/doxygen/d4/d0c/PHTruthTrackSeeding_8cc_source.html#l00068)

Yes. I'll do that. It will also be the occasion to see if something there can lead to the issues pointed out by @bogui56

blackcathj:
@hupereir , please ignore these "abort" errors which is due to server busy and jobs time out. The next check on  654747c will also fail due to the time out too. Time out interval has just been updated to avoid this again.

PR #817, opened 2020-05-14
suppress cppcheck warnings

pinkenburg:
This adds
// cppcheck-suppress *
before the line with the warning which suppresses all warnings (sadly I cannot find docs with the actual names of the warnings to be used for the suppression so this suppresses all


hupereir:
Hi Chris,
I think 'cppcheck --errorlist' gives you an xml files with all error ids, (that you can include in your suppression comment) and corresponding error message.
I have not tested it though ...

PR #819, opened 2020-05-18
Remove inadvertently added code

osbornjd:
I'm assuming that the build directory for the `trackreco` package was inadvertently added. There are also a number of `cout` statements that were not enclosed within verbosity statements added over the weekend. This PR removes the `trackreco` build directory from the master branch and also encloses the extraneous `cout` statements added within `Verbosity` checks.


offline/packages/trackreco/PHTruthVertexing.cc
```diff
@@ -59,11 +59,14 @@ int PHTruthVertexing::Setup(PHCompositeNode* topNode)
...
-    cout << "embed only true" << endl;
-  }else{
-    cout << "embed only true" << endl;
-  }
+  if(Verbosity() > 1)
+    {
+      if(_embed_only){
+	cout << "embed only true" << endl;
+      }else{
+	cout << "embed only true" << endl;
```
hupereir, line 67: Hi Joe, 
Thanks for fixing this. 
As a side remark, the text here is also incorrect. "true" should be "false"
Also, a shorter version of this "if" block could be 
  std::cout << "embed only: " << std::boolalpha << _embed_only << std::endl

PR #824, opened 2020-05-26
Mvtx energy threshold

zhengyunyou:
Add a pixel energy threshold for each mvtx hit. 
Hits with energy under threshold will be removed.
This change is to match the cluster size distribution in simulation with those from test beam data.

### DIFF
--- a/simulation/g4simulation/g4mvtx/PHG4MvtxDigitizer.cc
+++ b/simulation/g4simulation/g4mvtx/PHG4MvtxDigitizer.cc
@@ -30,7 +30,7 @@ using namespace std;
 
 PHG4MvtxDigitizer::PHG4MvtxDigitizer(const string &name)
   : SubsysReco(name)
-  , _energy_threshold(0.0)
+  , _energy_threshold(0.95e-6)
 {
   unsigned int seed = PHRandomSeed();  // fixed seed is handled in this funtcion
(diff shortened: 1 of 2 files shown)


zhengyunyou:
Add a pixel energy threshold for each mvtx hit. Hits with energy under threshold will be removed.
Collect hits to remove in a set, and remove them later in another loop.

zhengyunyou:
Add a pixel energy threshold for each mvtx hit. Hits with energy under threshold will be removed.
Collect hits to remove in a set, and remove them later in another loop.

blackcathj:
@zhengyunyou 
Thanks for the update. It passed all code checkers. 

Just to highlight the response from the MVTX simulation: 
* Low occupancy: https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-tracking-low-occupancy-qa/239//QA_20Report/ 
* High occupancy: https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-tracking-high-occupancy-qa/145//QA_20Report/ 

It appears the cluster size is much more close to the expectation from the test beams now. Also it improved the centering of the pull distribution, which is very interesting to see

blackcathj:
Merging per discussion of MVTX testbeam analysis and simulation meeting: https://indico.bnl.gov/event/8612/

PR #838, opened 2020-06-16
Acts Proto Tracks

osbornjd:
This PR puts the proto track information that is fed into the Acts Fitter onto the ActsEvaluator output tree. This can allow for all the hit information to be analyzed after the fact, i.e. all hits/info that goes into the fitter and all hits/info that come out of the fitter.


osbornjd:
I'm guessing this is failing because I was working on a branch that was way behind the master branch due to the ongoing issues with PHGenFitTrkProp. I'll close this PR and make a new branch from the master repository and `git cherry-pick` the commits into the new branch, since I have already tested them.

PR #839, opened 2020-06-16
Acts proto track eval

osbornjd:
This PR puts the proto track information that is fed into the Acts Fitter onto the ActsEvaluator output tree. This can allow for all the hit information to be analyzed after the fact, i.e. all hits/info that goes into the fitter and all hits/info that come out of the fitter.

This PR replaces [PR838](https://github.com/sPHENIX-Collaboration/coresoftware/pull/838) which failed due to me testing on an older version of `coresoftware` as there currently is an issue between the communication of `PHGenFitTrkProp` and `PHActsTrkFitter`.


blackcathj:
Please ignore DST readback error, which was from recent macros repository changes.

PR #842, opened 2020-06-17
Remove hardcoded tracknode name

pinkenburg:
This PR replaces the hardcoded Track Node Name by an argument to the ctor of TrackJetInput. This should help getting rid of these svtxtrack maps. It defaults to SvtxTrackMap so it does not break old code/macros


blackcathj:
This fail is from Jenkins setup update last night on the QA pipelines, and not related to this pull request

PR #846, opened 2020-06-22
Acts Track Propagation

osbornjd:
This PR updates and really makes the track propagation from Acts runnable in sPHENIX. The CKF track finder in Acts is called from `PHActsTrkProp` and the results are stored in the `SvtxTrackMap` and the `ActsFitResults` for use with the evaluator tools. This needs to be tuned pretty significantly with the source link selector object provided by Acts, as the fits are not that great at the moment.

The PR also fixes a bug in the communication between PHGenFitTrkProp and PHActsTracks which was causing some of the weird behavior we saw with the Acts track fitting previously. Apparently the GenFit track propagation code resets the track charge but does not re-initialize it, and GenFit does not care about this. Acts does and this was part of the reason there were very poor track fits previously.

This PR also corresponds to the [changes](https://github.com/sPHENIX-Collaboration/acts/pull/15) that will be merged shortly into the sPHENIX/Acts repo.

### DIFF
--- a/offline/packages/trackreco/ActsCovarianceRotater.cc
+++ b/offline/packages/trackreco/ActsCovarianceRotater.cc
@@ -177,7 +177,7 @@ Acts::BoundSymMatrix ActsCovarianceRotater::rotateActsCovToSvtxTrack(
 }
 
 
-void ActsCovarianceRotater::printMatrix(const std::string message,
+void ActsCovarianceRotater::printMatrix(const std::string &message,
 					Acts::BoundSymMatrix matrix)
 {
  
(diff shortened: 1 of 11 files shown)


blackcathj:
`CI-DST-readback-FAIL` tag is from missing this fix https://github.com/sPHENIX-Collaboration/coresoftware/pull/847 to the main repository after macro changes

osbornjd:
I pushed another commit with some verbosity statements for Tony and I to use, so if the DST-readback fails again I'll just ignore it

blackcathj:
The memory leak in https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-default-valgrind-pipeline/1797//valgrindResult/ is related to `PHTpcTracker` and GenFit2.

PR #848, opened 2020-06-24
only print fatal messages

pinkenburg:
This PR changes the default priority for log4cpp messages to print on the screen from INFO (everything) to FATAL (nothing)


pinkenburg:
The memory leak is somewhere else - merging this and fixing the leak in the next PR

PR #850, opened 2020-06-26
Fix branches in Evaluator Tree

osbornjd:
Some branches were accidentally constructed as leaves, instead of branches. This PR fixes that.


blackcathj:
The warning is a picked out from the `online_distribution` repository: 
https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/Build-Clang/879//clang/new/source.c54ff2ae-6384-4781-b533-a6d3dda12224/#48
Merging.

PR #855, opened 2020-07-02
Fix micromega readback

pinkenburg:
The micromega geometry gave readback warnings since its helper class MicromegasTile did not have a Streamer. Added the micromega io lib to libg4dst, fixed includes according to include what you use. Enabling the micromegas crashes the tracking evaluator (g4hit null pointer dereference)

### DIFF
--- a/offline/packages/micromegas/Makefile.am
+++ b/offline/packages/micromegas/Makefile.am
@@ -11,7 +11,7 @@ AM_CPPFLAGS = \
   -I$(includedir) \
   -I$(OFFLINE_MAIN)/include  \
   -I$(ROOTSYS)/include \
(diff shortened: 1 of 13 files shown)


hupereir:
On the evaluator crash: 
- yeah I did not test if things are running with the evaluator
- most likely what happens is that the evaluator finds g4hits keys that cannot be associated to any container (because it does not know about the container for micromegas, does not load the node, etc.) 
I had the same problem in trackprop before.
I will provide a fix later today.

Hugo

hupereir:
PS: in any case I think it is a good idea to have the micromegas disabled by default 
This is highly experimental and irrelevant for most people doing tracking and simulation studies.

PR #860, opened 2020-07-09
Simplify vmeson gen

pinkenburg:
This PR makes some minor mods to the upsilon generator: added simplified interface to select decay (just electron or muon), checks for only valid decays (e+ & e-, mu+ & mu-), use x*x instead of pow(x,2), declare protected datamembers private, Some include what you use suggestions were added for good measure


pinkenburg:
Merging before unrelated tests have finished. Need the new VMeson generator for the production for macro changes

PR #863, opened 2020-07-11
simplified the code. 

hupereir:
As hinted by the comments and verified using printouts, the second part of the error calculations, involving the weighted mean of drphi or dz, is always zero. I removed the corresponding code, to speed up the execution.

### DIFF
--- a/offline/packages/tpc/TpcClusterizer.cc
+++ b/offline/packages/tpc/TpcClusterizer.cc
@@ -487,28 +487,24 @@ int TpcClusterizer::process_event(PHCompositeNode *topNode)
 
       // Estimate the errors
       double dphi2_adc = 0.0;
-      double dphi_adc = 0.0;
       double dz2_adc = 0.0;
-      double dz_adc = 0.0;
       for (int iz = zbinlo[iclus]; iz <= zbinhi[iclus]; iz++)
       {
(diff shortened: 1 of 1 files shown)


blackcathj:
The check-error message comes from an on-going development of JenkinsCI. Report is accessible via 
https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/sPHENIX_CoreSoftware_PullRequest/1450/

Nonetheless, there is a noticeable change in tracking efficiency with this PR:
* Low occupancy: https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-tracking-low-occupancy-qa/356/QA_20Report/
* High occupancy: https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-tracking-high-occupancy-qa/261/QA_20Report/
The effect is more obvious in the low occupancy case: 
image
image

Suggest discuss prior to a merging decision.

hupereir:
@blackcathj 
Fully agree. I must have missed something. I'll double check before merging. In fact it is probably enough to just close this one

hupereir:
Closing for now. If I had fully understood the code, there should have been no change at all. The fact that it is not true means that I must have overlooked something. I will come with a new PR (if needed) after double checking. 
Sorry for the noise.

PR #864, opened 2020-07-12
Tpc cluster error v2

hupereir:
This is the second attempt at simplifying the TPC cluster error calculatio code.
First commit is similar to the original attempt, but with better handling of size one cluster. Not identifying size-one cluster due to numerical divergences is the reason why the first attempt lead to a deterioration of the tracking efficiency. The clusters positions and errors after this first commit are identical to what they were before with simpler and more robust code
The second commit remove hits for which adc - pedestal is <0. This fixes cases for which the resulting cluster error square was negative, and the corresponding error -nan. 
This has a bigger impact on all cluster errors, hopefully for the best. This should hopefully lead to either identical or improved performances. If not, this one can always be reverted before pushing.

### DIFF
--- a/offline/packages/tpc/TpcClusterizer.cc
+++ b/offline/packages/tpc/TpcClusterizer.cc
@@ -38,6 +38,11 @@
 #include <utility>  // for pair
 #include <vector>
 
+namespace 
+{
+  template<class T> inline constexpr T square( const T& x ) { return x*x; }
+}
+
 using namespace std;
 
 TpcClusterizer::TpcClusterizer(const string &name)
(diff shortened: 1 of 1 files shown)


sPHENIX-GitHub-l:
Hi Hugo,

Hits for which (adc-pedestal) is less than zero were kept because there will be such hits in real data. The TPC reads out a fixed number of time bins after the input exceeds threshold.

We should discuss this before dropping them.

Cheers
Tony

hupereir:
Hi Tony, 
I understand that we should keep those hits, 
my point is more than they still should not be accounted for in the centroid and error determination for clusters. 
1/ we already do not account for hits with ADC-pedestal == 0 (by construction they get a weight of zero, and the corresponding hit is not even associated to the cluster in the current code). I would think one should not handle clusters with ADC-pedestal < 0 differently from those for which ADC-pedestal ==0 in the centroid determination (but I might miss something) 
2/ when you calculate a weighted RMS, via sum(w.x**2) -sum((w.x)**2), you can end up with a negative value when some of your weights are not positive. (which happens if you keep the ADC-pedestal < 0 in your error calculation). This in turns leads to cluster errors = -NAN. (I have found some of such clusters in the existing code). One way or another we must prevent this -NAN to happen. Handling hits for which ADC-pedestal < 0 in the same way as hits for which ADC-pedestal == 0 was to me the most natural way to deal with this. 
Of course this is open for discussion.

sPHENIX-GitHub-l:
Hi Hugo,

I don't disagree that the errors should be handled differently.

Tony

PR #866, opened 2020-07-14
Fit WRT Initial Vertex Estimate

osbornjd:
This PR changes the Acts track fitting/propagation to fit with respect to the initial track vertex estimate given by the `PHInitVertexing` that is used (either `PHTruthVertexing` or `PHInitZVertexing`). Previously we were just fitting wrt to the global origin (0,0,0), which in discussion with the Acts developers is not the intended use of the fitter.

This PR also introduces a small change in commit `ee2f314` to return a `nullptr` instead of `exit` in the event an appropriate surface for a TPC cluster isn't found. This way the cluster is just skipped and not added to the fitting procedure rather than causing the whole code to exit.


osbornjd:
Is this on my PR? Looking through the Jenkins report it seems like there is some weird root memory leak?

blackcathj:
No. The leak is from `PHTpcTracker`, something we are still working on. Good to have a reminder the problem is still there  ;-)

PR #867, opened 2020-07-14
Removed redundant zero suppression by GeV

griffsk:
[https://indico.bnl.gov/event/7423/](url) from zero suppression presentation.


cfmcginn:
@blackcathj @pinkenburg This is a PR to try and resolve what was discussed today (see indico link from @griffsk above). It appears someone needs to add @griffsk per above to trigger testing.

blackcathj:
Invitation sent via GitHub to @griffsk.

PR #868, opened 2020-07-14
Fixed copy-from-baseclass constructor

hupereir:
It appears that PHG4Particlev2 constructor from pointer to base class was not propagating the track id, vertex id, primary id and energy, resulting in having the values being reset when doing a deep copy of the object. I don't think this is intentinal.
This PR fixes that.


blackcathj:
Jenkins server suffered a crash yesterday. Restarting all tests. Please ignore the last test results.

PR #870, opened 2020-07-16
Fixed uninitialized variable and modernized

hupereir:
variable showerid was never properly initialized in G4Hitv1 constructor, nor reset in the ::Reset method.
I took the opportunity of fixing this to modernize the code: 
- moved all member initialization to the header (this should help not making the same mistake again)
- marked the constructor (and the destructor) as "default" since they are both empty.


blackcathj:
The error comes from a known rare memory leak of `PHTpcTracker` which is still to be fixed. Merging

PR #878, opened 2020-07-31
Creating QAG4SimulationVertex module

rosstom2232:
Adding .cc, .h, and LinkDef.h files to create QAG4SimulationVertex class in coresoftware repo. Updating Makefile.am to account for those changes.
Also, fixing an issue with the way I made my loop in QAG4SimulationTracking to prevent it from double-filling some histograms.


blackcathj:
The valgrind error is some CA tracking module. Please ignore for this pull request @rosstom2232

PR #881, opened 2020-08-03
Add momentum to g4hits

hupereir:
This PR makes sure that script-generated stepping Action also passes the Geant4 hit momentum information to PHG4Hits
It also adds it to PHG4MicromegasSteppingAction
The rational is that: 
- since the G4Hits members are here, they better be filled with the correct information
- the momentum information might be useful for calculating "local" momentum resolution (as opposed to mom resolution at origin)
- the same code already exists for the TPC, INTT and MVTX


pinkenburg:
The momentum is added to the PHG4Hits via a property. The perl script creates a template which just fills the data members of PHG4Hit. The implicit understanding is that properties get added as needed. Since the properties are computationally expensive I don't want to make this part of the default behavior (assuming users will not strip this out if they do not need it)

hupereir:
> The momentum is added to the PHG4Hits via a property. The perl script creates a template which just fills the data members of PHG4Hit. The implicit understanding is that properties get added as needed. Since the properties are computationally expensive I don't want to make this part of the default behavior (assuming users will not strip this out if they do not need it)

True
So I will remove the part of the PR that modifies the script. Still: I need the info in the Micromegas (and it is copied in all our tracking detectors as far as I can tell)

pinkenburg:
Sure - the tracking detectors need this information - no objection whatsoever.

PR #901, opened 2020-08-28
Tpc tracker interface and Vertex SL

osbornjd:
This PR implements changes to have the Acts CKF run with the TpcTracker seeder. The fitting produces good fits in momentum but not in track position, which probably is related to the poor TpcTracker seed position. This is a problem we will have to discuss with all of the TPC seeders.

This PR also implements the option to add the event vertex as a measurement to add to the fitting procedure. The measurement is added and the CKF recognizes this; however, we will have to iterate with the ACTS developers on how to force the propagator to "see" this measurement as it is not part of the `Acts::TrackingGeometry`. Currently the CKF only visits surfaces associated to the geometry and not isolated surfaces (e.g. the one associated to the vertex measurement).

One other issue with TPC seeders that will need to be addressed is that they currently don't associate a vertex to the track seed. Acts requires that the track seed be associated to a vertex estimate already before going into the CKF. Currently, if the track seed vertex has not been set, this PR just resets it to the first `SvtxVertex` available.

This PR also makes the Hough seeder functionality obsolete with Acts tracking, since the hough seeding has a mismatch in the track sign. For future reference, in case we need to come back....

### DIFF
--- a/offline/packages/PHTpcTracker/externals/kdfinder.hpp
+++ b/offline/packages/PHTpcTracker/externals/kdfinder.hpp
@@ -699,7 +699,7 @@ namespace kdfinder
         T t40 = (-t3 * t38);
         if (t40 < 0.)
         {
-          std::cerr << "t40 < 0." << std::endl;
+          //std::cerr << "t40 < 0." << std::endl;
           return VALUE;
         }
         t40 = std::sqrt(t40);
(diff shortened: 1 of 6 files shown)


osbornjd:
I forgot to push this commit which reduces the space point covariance to that of the vertex. So after Jenkins runs this should be good to go.

PR #902, opened 2020-08-28
Keep zero e hits

pinkenburg:
This PR adds a switch to the cylinder subsystem which keeps zero energy hits. This is needed for reference cylinders for projections where we want to look at the actual G4 hit position.

Added override keyword to the PHG4CylinderSubsystem mathods

This adds a recoConst flag for the world material, so we can change it consistently (right now you'll find G4_AIR all over the place which then messes with the material scans where we use G4_Galactic.

Farewell to root5 in simulation/g4simulation/g4detectors


pinkenburg:
The failure is e memory leak in the PHTpcTracker

PR #904, opened 2020-09-01
New Clusterizer

bogui56:
Eliminates duplicates and should provide better cluster splitting capability

TpcTracker tpc only reconstruction efficiency is veru high in high multiplicity qa events:
tpceff_nuclus

Obvious cluster splitting issues should be fixed:

clu_split_match2_zoom_nu

Before:
clu_split_match2_zoom

### DIFF
--- a/offline/packages/tpc/TpcClusterizer.cc
+++ b/offline/packages/tpc/TpcClusterizer.cc
@@ -38,6 +38,11 @@
 #include <utility>  // for pair
 #include <vector>
 
+namespace 
+{
+  template<class T> inline constexpr T square( const T& x ) { return x*x; }
+}
+
 using namespace std;
 
 typedef std::pair<int, int> iphiz;
@@ -117,66 +122,6 @@ bool TpcClusterizer::is_in_sector_boundary(int phibin, int sector, PHG4CylinderC
   return reject_it;
 }
 
-
-void TpcClusterizer::get_cluster(int phibin, int zbin, int &phiup, int &phidown, int &zup, int &zdown, std::vector<std::vector<double>> &adcval)
-{
-  // search along phi at the peak in z
-
-  for (int iphi = phibin + 1; iphi < phibin + 4; iphi++)
-  {
-    if (iphi >= NPhiBinsMax) continue;
-
-    if (adcval[iphi][zbin] <= 0)
-      break;
-
-    if (adcval[iphi][zbin] <= adcval[iphi - 1][zbin])
-      phiup++;
-    else
-      break;
-  }
-
-  for (int iphi = phibin - 1; iphi > phibin - 4; iphi--)
-  {
-    if (iphi < NPhiBinsMin) continue;
-
-    if (adcval[iphi][zbin] <= 0)
-      break;
-
-    if (adcval[iphi][zbin] <= adcval[iphi + 1][zbin])
-      phidown++;
-    else
-      break;
-  }
-
-  // search along z at the peak in phi
-
-  for (int iz = zbin + 1; iz < zbin + 10; iz++)
-  {
-    if (iz >= NZBinsMax) continue;
-
-    if (adcval[phibin][iz] <= 0)
-      break;
-
-    if (adcval[phibin][iz] <= adcval[phibin][iz - 1])
-      zup++;
-    else
-      break;
-  }
-
-  for (int iz = zbin - 1; iz > zbin - 10; iz--)
-  {
-    if (iz < NZBinsMin) continue;
-
-    if (adcval[phibin][iz] <= 0)
-      break;
-
-    if (adcval[phibin][iz] <= adcval[phibin][iz + 1])
-      zdown++;
-    else
-      break;
-  }
-}
-
 void TpcClusterizer::remove_hit(double adc, int phibin, int zbin, std::multimap<double, ihit> &all_hit_map, std::vector<std::vector<double>> &adcval)
 {
   typedef multimap<double, ihit>::iterator hit_iterator;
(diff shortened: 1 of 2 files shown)


hupereir:
Hi Christof,
Thanks for the commit, this looks very promising. However I noticed that some of the "recent" changes to the code were reverted in the merging. In particular those concerning not taking into account hits with negative ADC value as opposed to those with zero ADC value. Did you start your changes from the latest version ? 
This is important because it was found in the past that keeping negative ADC values and using negative weights in weighted averages can resutl in negative RMS square, and thus NAN errors in the clusters. 
Could you double-check ? 
I'll add some inline comments about the parts of the code that seem "reverted" to me.

offline/packages/tpc/TpcClusterizer.cc
```diff
@@ -38,13 +38,11 @@
 #include <utility>  // for pair
 #include <vector>
 
-namespace 
-{
-  template<class T> inline constexpr T square( const T& x ) { return x*x; }
-}
-
```
hupereir, line 45: Using "square(x)" later in the code rather than x*x is to me a clear improvement: More explicit, less error prone, in particular when you want to square a complex expression. 
I don't think this should be removed

offline/packages/tpc/TpcClusterizer.cc
```diff
@@ -415,118 +789,104 @@ int TpcClusterizer::process_event(PHCompositeNode *topNode)
...
         for (int iz = zbinlo[iclus]; iz <= zbinhi[iclus]; iz++)
         {
-          // skip hits for which weight is non-positive
-          /*
-           * Having negative weights in the calculation of the error might result in negative variance
-           * and thus -nan error, that then propagate through the tracking. Since those hits correspond to
-           * noise anyway and in order to have the error and centroid calculation consistent,
-           * we skip the hit alltogether
-           */
-          if( adcval[iphi][iz] <= 0 ) continue;
```
hupereir, line 475: Here is a change that I think should not be reverted. This is exactly what prevents to have NAN uncertainties on otherwise perfectly valid clusters.
Also, as discussed in an earlier meeting, not keeping hits with ADC==0 but keeping those with ADC <0 does not make much sense.

offline/packages/tpc/TpcClusterizer.cc
```diff
@@ -415,118 +789,104 @@ int TpcClusterizer::process_event(PHCompositeNode *topNode)
...
+          double dphi = layergeom->get_phicenter(iphi) - clusphi;
+          dphi2_adc += dphi * dphi * adcval[iphi][iz];
+          dphi_adc += dphi * adcval[iphi][iz];
+
+          double dz = layergeom->get_zcenter(iz) - clusz;
+          dz2_adc += dz * dz * adcval[iphi][iz];
+          dz_adc += dz * adcval[iphi][iz];
+        }
+      }
+      double phi_cov = (dphi2_adc / adc_sum - dphi_adc * dphi_adc / (adc_sum * adc_sum));
```
hupereir, line 870: This is an exemple where using "square(dphi_adc/adc_sum)" would make the code more readible in my oppinion

offline/packages/tpc/TpcClusterizer.cc
```diff
@@ -415,118 +789,104 @@ int TpcClusterizer::process_event(PHCompositeNode *topNode)
...
+      double z_cov = dz2_adc / adc_sum - dz_adc * dz_adc / (adc_sum * adc_sum);
+
+      //cout << "   layer " << layer << " z_cov " << z_cov << " dz2_adc " << dz2_adc << " adc_sum " <<  adc_sum << " dz_adc " << dz_adc << endl;
 
-      // phi_cov = (weighted mean of dphi^2). The error is then:
+      // phi_cov = (weighted mean of dphi^2) - (weighted mean of dphi)^2,  which is essentially the weighted mean of dphi^2. The error is then:
```
hupereir, line 875: This change should not be reverted. The old comment is the correct one.

offline/packages/tpc/TpcClusterizer.cc
```diff
@@ -547,22 +907,22 @@ int TpcClusterizer::process_event(PHCompositeNode *topNode)
       DIM[0][1] = 0.0;
       DIM[0][2] = 0.0;
       DIM[1][0] = 0.0;
-      DIM[1][1] = square(0.5*phi_size);  //cluster_v1 expects 1/2 of actual size
+      DIM[1][1] = pow(0.5 * phi_size,2);  //cluster_v1 expects 1/2 of actual size
```
hupereir, line 910: Please do not change this. It is much more efficient and readible to use "square" than "pow(x,2)" The result is also more precise.

offline/packages/tpc/TpcClusterizer.cc
```diff
@@ -415,118 +789,104 @@ int TpcClusterizer::process_event(PHCompositeNode *topNode)
...
-      const double phi_err_square = (phibinhi[iclus] == phibinlo[iclus]) ?
-        square(radius*layergeom->get_phistep())/12:
-        square(radius)*phi_cov/(adc_sum*0.14);
-
-      const double z_err_square = (zbinhi[iclus] == zbinlo[iclus]) ?
-        square(layergeom->get_zstep())/12:
-        z_cov/(adc_sum*0.14);
+      double z_err = sqrt(z_cov / (adc_sum * 0.14));
+      if (z_err == 0.0)
+        z_err = layergeom->get_zstep() / sqrt(12.0);
```
hupereir, line 889: The old code is more efficient.

bogui56:
Hi Hugo, 

bugger, it looks like I started from an outdated version. 
Reverting wasn't intentional. 

Let me see if I can get a proper diff and put the updates back in. 

Cheers

   Christof

offline/packages/tpc/TpcClusterizer.cc
```diff
@@ -330,27 +676,77 @@ int TpcClusterizer::process_event(PHCompositeNode *topNode)
...
+      // -> add hits to truth association
+      // remove hits from all_hit_map
+      // repeat untill all_hit_map empty
+      //      print_cluster(ihit_list);
+      calc_cluster_parameter(ihit_list,nclus++, layergeom, hitset);
+      remove_hits(ihit_list,all_hit_map,adcval);
+    }
+  }
 
+#ifdef ISOLD
```
blackcathj, line 749: `cppcheck` complain unmatched `}` in the case of ISOLD`:
```
TpcClusterizer.cc:998
Unmatched '}'. Configuration: 'ISOLD'. 
```
https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/cpp-check-pipeline/1534/cppcheck/new/

blackcathj:
There is a Jenkins server problem at the moment. Please ignore the reports

blackcathj:
Hi @bogui56 

We were not able to capture the memory problem you mentioned via the Jenkins valgrind run (only tested two events of default macro). 

There are two minor nagging from `clang` build: 
* private field 'hit_nt' is not used at TpcClusterizer.h:59
* private field 'cluster_nt' is not used at TpcClusterizer.h:60

blackcathj:
Two `valgrind` memory leak errors: https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-default-valgrind-pipeline/1995//valgrindResult/

This one appears new: 
```
offline/packages/trackreco/PHGenFitTrkProp.cc:1574
```
This one has been there for a while: 
```
offline/packages/PHTpcTracker/PHTpcTrackFollower.cc:508
```

Please ignore the `cppcheck` errors this time. There appears to be a reference build issue.

PR #906, opened 2020-09-02
Change TPC gas properties to Ne/CF4 50/50

hupereir:
Title says it all. 
Diffusion and drift velocity numbers from http://skipper.physics.sunysb.edu/~prakhar/tpc/HTML_Gases/split.html
(Thanks Ross !)


blackcathj:
There is a Jenkins server problem at the moment. Please ignore the reports

PR #907, opened 2020-09-04
Update gain to match gas change from Ne/CF4 90/10 to Ne/CF4 50/50

hupereir:
This ensures that the average charge per particle per pad is kept unchanged.
The ratio between old and new gain correspond to the inverse ratio of the average number of primary electron per cm for a MIP


mohaas33 (approved):

blackcathj:
Both issues are ongoing and not related to this pull request.

PR #912, opened 2020-09-10
first version of event plane detector

bi-ran:
this is based on the design of the STAR EPD. the geant4 model is built using extruded solids (1 per scintillator tile), with coordinates from Prashanth, who worked on the STAR EPD.

### DIFF
--- a/simulation/g4simulation/g4epd/Makefile.am
+++ b/simulation/g4simulation/g4epd/Makefile.am
@@ -1,0 +1,46 @@
+AUTOMAKE_OPTIONS = foreign
+
+AM_CPPFLAGS = \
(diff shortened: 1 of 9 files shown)


pinkenburg:
I ran include what you use and applied its recommended changes, cppcheck found an unused variable. Ran clang-format to apply our code formatting standard.
What is missing is the dedicated handling of geantinos in the stepping action, they cause zero energy loss and their hits will be dropped in the current implementation. This doesn't prevent the PR from being accepted but we rely heavily on geantino scans so this needs to be added.
This PR can be merged if it passes jenkins

pinkenburg:
It is the random leak in the phtpc tracker

PR #918, opened 2020-09-16
Acts KF Timing Additions

osbornjd:
This PR adds some histograms/analysis for doing timing analysis with the Acts KF. The behavior can be turned off at the macro level and is off by default.


hupereir:
"I may never learn to check cppcheck"
Hi Joe, 
for what it's worth, with c++11 you can initialize all your members to their default values directly in the header when they are declared (like here: TH1 *h_fitTime = nullptr), in which case you then don't need to initialize them in the constructor any more and cpp check is still happy. (essentially: anything that is not initialized to something in the constructor gets the value defined in the header). 
Personally I find this most convenient to have everything properly initialized in the header, especially if you have multiple constructors, since you then avoid the duplication of default initialization. Also, I find this easier to read afterwards. 
my two cents :)

PR #925, opened 2020-09-23
implement skipping for non syncing input managers

pinkenburg:
This PR is a quick (but reasonable) fix to skip events in the pileup manager properly so skipping will preserve the event ordering one would see when just running. The same change was applied to the nosync manager, so it will also advance its event counter. The mechanics need some more cleanup to get event numbers right and pushing back the previous events in case of reread. This will take a bit longer and will come in a later PR.

### DIFF
--- a/generators/phhepmc/Fun4AllHepMCInputManager.h
+++ b/generators/phhepmc/Fun4AllHepMCInputManager.h
@@ -11,9 +11,7 @@
 
 #include <utility>                                  // for swap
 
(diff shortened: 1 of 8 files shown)


pinkenburg:
Ignore this one - there is a problem when mixing running and skipping

blackcathj:
Server was busy last night and a few of the test reached execution time limit. Relaxing the limit to avoid future errors

PR #927, opened 2020-09-23
Magnetic Field Configuration

osbornjd:
This PR introduces configurations for setting the magnetic field in Acts based on the field set in the Fun4All macro. The magnetic field can now just be propagated from the Fun4All macro into the needed modules.

If the field map is used, then the Acts magnetic field is just set to a default 1.4T constant field. We are still working on getting Acts and TPCTracker to coexist with the field map (they seem to interpret them with some differences e.g. in direction). 

There is a fudge factor added in the stub matcher to get the deltaphi matching correction correct, based on the sign of the magnetic field. We should return to this once the field map situation is sorted out, but this puts the magnetic field related code into a stable state for the purposes of opening a macros PR to use Acts by default.


blackcathj:
Failure due to overtime of running the macro that is not related to this PR

PR #933, opened 2020-09-28
Fix micromegas tracking

hupereir:
This PR modifies how Micromegas 1D clusters are handled in the Genfit track fit: 
Even if the clusters are 1D one must provide an estimate of the "other" coordinate to the measurement. This is to have the planar approximation around the measurement, performed by GenFit, actually tengent to the detector. (this is not the case for z measurements, for which phi is assigned to the midle of the detector).
This estimate is based on the measurement provided by the other micromegas cluster, in the same track, if present. 
There is no double use of the said measurements because the errors are unchanged and kept very large for the direction that one given micromegas layer does not measure. 

This fixes some large biases in the z residual for tracks including Micromegas detectors.

Ultimately, Micromegas detectors will be implemented as planes rather than cylinders, making this "fix" irrelevant.

### DIFF
--- a/offline/packages/trackreco/Makefile.am
+++ b/offline/packages/trackreco/Makefile.am
@@ -195,6 +195,7 @@ libtrack_reco_la_LIBADD = \
   -lg4tpc \
   -lg4intt \
   -lg4mvtx \
+  -lmicromegas_io \
   -lmvtx_io \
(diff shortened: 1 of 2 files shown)


pinkenburg:
The reported clang failure does not make sense

PR #940, opened 2020-10-07
CA Seeding update

petermic:
Fixed covariance matrix and a possible sign error in the coordinate rotation, added the ability to filter for "good" clusters, and also improved configurability. Efficiency problems in high-PU events still persist.


pinkenburg:
Not sure why my fixes did not make it into the check. I merge - if this is still around I'll fix it in a separate PR. Please make your PR editable by the maintainer (it's a tick mark when sending the PR, it should be the default, not sure why it didn't work here) - I could not push the fixes into the branch

PR #941, opened 2020-10-08
Add Micromegas to Acts tracking

adfrawley:
This PR adds fake plane surfaces inside the Micromegas cylinders to the TGeo geometry so that we can make Acts source links, and adds a module to associate TPC seed tracks to Micromegas clusters. The latter also provides enough information so we can connect the two planes of Micromegas clusters to the correct plane surface. The Micromegas have also been integrated into SvtxEvaluator and friends.
 
The Acts tracking chain now finds and fits full tracks, including the Micromegas.


osbornjd:
@adfrawley @pinkenburg - Part of the problem is that the macro is picking up the Acts response file that is in the `/share` directory, which doesn't have the updated response file that is in this PR that contains the Micromega geometry. I'm not sure how we get around this in the CI, because in order for the macro to run it needs the update in this PR to `tgeo-sphenix.response` to be propagated to `$OFFLINE_MAIN/share`, but in order for that to happen the PR has to be merged.

We should probably come up with a different strategy so that this doesn't happen in the future. Maybe this file should go in the Calibrations repository, so that it can be merged separately and then coresoftware CI jobs can run which pick up the updated calibrations repo?

PR #944, opened 2020-10-10
Configure PFlow matching criteria and emulate finite track efficiency

dvperepelitsa:


pinkenburg:
The cppcheck errors are from offline/packages/trackreco. Fixed in another PR

PR #945, opened 2020-10-12
Fix ca seeding charge

adfrawley:
Makes PHCASeeding output track charge correct for any magnetic field direction.


pinkenburg:
No problem here - the cppcheck errors were fixed in the previous PR. Merging this

PR #955, opened 2020-10-21
Remove root5

pinkenburg:
This PR removes the various leftovers from root5 builds including the ifdef CINTs in our header files. Running with this does not change the results when running the default Fun4All_G4_sPHENIX.C


blackcathj:
The current check run will partly fail due to on-going effort rebuilding the DST-readback tests according to the new macro setup. Please ignore that particular error.

PR #956, opened 2020-10-23
MBD/BBC added to macros in common and detectors, added parts of BBC housing

mchiu-bnl:


blackcathj:
gcc8 build will be marked fail by Jenkins due to a known warning in database headers. It will be cleared in future builds.

PR #976, opened 2020-11-04
Tpc distortion reconstruction

hupereir:
This is an overall update of the track-base space charge distortion code
- reordered axis and change axis range to better match the format of the input distortion map.
- added checks on tracks and clusters before filling the matrices 
- save the result of the matrix inversions (namely, the space charge corrections) to 3D histograms of the same format as the input distortions rather than TGraphs. The histograms are written to a rootfile that can then be used as input for correcting the cluster positions.

### DIFF
--- a/offline/packages/tpccalib/TpcSpaceChargeReconstruction.cc
+++ b/offline/packages/tpccalib/TpcSpaceChargeReconstruction.cc
@@ -12,7 +12,7 @@
 #include <trackbase_historic/SvtxTrackMap.h>
 
 #include <TFile.h>
-#include <TGraphErrors.h>
+#include <TH3.h>
 
 #include <memory>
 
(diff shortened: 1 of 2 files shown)


hupereir:
Not sure what's wrong on this one
Screenshot_20201105_090918
"Failure ... build is success"

pinkenburg:
The compiler report from x8664 under (new) shows the problem: TpcSpaceChargeReconstruction.cc:53
 [[maybe_unused]] TH3* create_histogram( TH3* hin, const TString& name )
gcc 4.8 (which we still need to support) does not know this attribute. To make it (and gcc 8.3) happy one needs top use:
__attribute__((unused))
like in
  static unsigned int bitshift_upperkey __attribute__((unused)) = 16;
in simulation/g4simulation/detectors/PHG4CellDefs.h

PR #1003, opened 2020-11-20
fix remaining mem leaks in Acts and stub matching

osbornjd:


offline/packages/trackreco/PHSiliconTpcTrackMatching.cc
```diff
@@ -292,8 +292,7 @@ int PHSiliconTpcTrackMatching::Process()
 	    {
 	      // more than one si stub matches
 	      // make a copy of the TPC track, update it and add it to the end of the node tree 
-	      
-	      SvtxTrack *newTrack = new SvtxTrack_v1();
+	      auto newTrack = std::make_unique<SvtxTrack_v1>();
```
hupereir, line 295: Hi Joe, 

Not knowing the deep details of this, this change looks very fragile (or event crashing): making new track a unique_ptr makes it being deleted when it gets out of scope (at exit of the method). But then later on, you insert the corresponding pointer in a map, which is a member of the class (and is not cleared at the end of the method). To me at least, it means that as soon as you exit the method the _track_map contains only dangling pointers. Or did I miss something ? 
also, (but unrelated to this change) the fact that a derived class (PHSiliconTpcTrackMatching) can access the members of the parent class (here _track_map) is also very fragile. In general I though it was safer (design wise) to have only methods as protected, and keeping members are private, in base class, to have better control over constness.

offline/packages/trackreco/PHSiliconTpcTrackMatching.cc
```diff
@@ -345,7 +344,7 @@ int PHSiliconTpcTrackMatching::Process()
 		  _assoc_container->SetClusterTrackAssoc(*clus_iter, newTrack->get_id());
 		}
 
-	      _track_map->insert(newTrack);
+	      _track_map->insert(newTrack.get());
```
hupereir, line 347: maybe newtrack.release() is what is intended here ? this way only the unique_ptr whose tracks are not inserted in the map are deleted (preventing dangling pointers in the map)

offline/packages/trackreco/PHSiliconTpcTrackMatching.cc
osbornjd: I think you are right, which is probably why Jenkins is very upset. Let me see if there is a better way to do this

osbornjd: I am puzzled though because this doesn't happen in my local tests...

hupereir: dangling pointers are nasty. It could be that in your local test the released memory has not been reused yet when accessed, so that the later code still access a proper track and produces no crash. But it might then crash on a later event, different machine or anything of that kind. (again, unlesss I missed something about the code intend)

osbornjd: If I look at the example [here](http://www.cplusplus.com/reference/memory/unique_ptr/get/) I guess you are saying `release` is better because it moves the ownership away from the unique pointer to the other object, is that right? Whereas `get` still keeps some relation between the two.

hupereir: Exactly

hupereir:
Sorry for the many comments. looking more at the code, I realised that using "release" will still crash, few lines later, because "newtrack" (which has been released) is accessed in a cout statement (with verbosity > 3). 
So you would need to move this before the call to "release", or remove the debug statement altogether. 
You cannot access the unique pointer any more once "release" has been called: it us nullptr.

osbornjd:
> Sorry for the many comments. looking more at the code, I realised that using "release" will still crash, few lines later, because "newtrack" (which has been released) is accessed in a cout statement (with verbosity > 3).
> So you would need to move this before the call to "release", or remove the debug statement altogether.
> You cannot access the unique pointer any more once "release" has been called: it is nullptr.

Oops, I didn't notice that - thanks for the catch

pinkenburg:
gcc 4.8 barfs with std::make_unique since this is c++14 (and we need to keep gcc 4.8 alive for the time being)
there is an easy fix for this by using boost (just applied it to PHCASeeding.cc) for compilers not supporting c++14:
the header:
#if __cplusplus < 201402L
#include <boost/make_unique.hpp>
#endif

....

 #if __cplusplus < 201402L
  t_fill = boost::make_unique<PHTimer>("t_fill");
  t_seed = boost::make_unique<PHTimer>("t_seed");
#else
  t_fill = std::make_unique<PHTimer>("t_fill");
  t_seed = std::make_unique<PHTimer>("t_seed");
#endif

PR #1009, opened 2020-11-25
Master hklest

hupereir:
This pull request implements new distortions from Henry Klest. 
The distortion code is moved out of PHG4TpcElectronDrift to PHG4Distortion, which is in turn passed to the electron drift code in the steering macro.
New code supports reading either static or per-event distortion maps, or both, from different files. 
Classes obsoleted by this change are removed. 
By default distortions are disabled
it was checked that using only static distortions, with the same input file, leads to the same result as with the old distortion code.

Should go together with https://github.com/sPHENIX-Collaboration/macros/pull/345


hupereir:
So it seems that for the failing tests, this is either due to 
- macros not being up to date (meaning that https://github.com/sPHENIX-Collaboration/macros/pull/345 should be merged first ?)
- some issue with AnaTutorial.cc (https://web.racf.bnl.gov/jenkins-sphenix/job/sPHENIX/job/Build-ScanBuild/730/clang-analyzer/new/), unrelated with this change.
Or do I miss something ?

PR #1016, opened 2020-12-02
Acts silicon seeding

osbornjd:
This PR adds a module that runs the Acts seeder. We will use this for the silicon stubs, so it it. currently runs only over the silicon measurements. 

The seeder is not tuned at all, so it is not ready for broader use. But we can start testing it with the stub matching chain for tuning/efficiency studies.

Currently the seeder sets a default vertex of 0. We will run the Acts vertex finder over the results of this module to collect initial reconstructed vertex estimates for the rest of the tracking chain. That needs to be tested and tuned also, although the machinery is available to do this right now.


osbornjd:
Not sure what is going on here... there are no compiler issues with my code (according to the compiler report) and regardless the macros don't call this class so I'm not sure why the default macro would not be able to run.

pinkenburg:
Don't worry - Hugo's PR needed macro changes which I merged to get his PR through jenkins. Hugos coresoftware PR is now merged, so this problem should go away by itself. I merged your PR

PR #1020, opened 2020-12-08
Initial vertexing

osbornjd:
This PR adds two major changes to the repository:

1. It adds an initial vertexing module that is interfaced with the silicon seeder. This can be run after the silicon seeder to produce an initial vertex estimate and update the `SvtxTrack` instances with the appropriate vertex ID.

2. It fixes/improves the silicon seeder to have the correct charge for the seeds, and phi and theta position. The seeder is still not tuned and we still need to learn what it is actually doing, but now that it produces correct output we can do that.

Note that the initial vertexer (unfortunately) dies within Acts a high percentage of times because the propagator doesn't converge. I suspect this is due to the poor momentum resolution of the seeds, but I'm not sure yet. I am investigating why this is and if it is related to something deeper within the Acts code.


osbornjd:
Any idea why this compiles locally for me, but does not on the CI?

PR #1047, opened 2021-01-07
Fix to KFParticle DST output

cdean-github:

## Types of changes

I noticed when generating MDC data that when creating a track map for the loose reconstruction, the tracks used disappear from the original track map rather than being copied from them. I cloned the original track map then made the new container from there which keeps the original tracks in place. This might be related to the crash from jet truth matching which could be trying to access the missing tracks.

I also increased the verbosity level for the DST class as it mostly duplicates the basic reconstruction output and would thus be annoying for most users.

## TODOs (if applicable)

Run a slow valgrind test after this patch to see if it fixes the conflict with the jet truth matching

### DIFF
--- a/offline/packages/KFParticle_sPHENIX/KFParticle_sPHENIX.cc
+++ b/offline/packages/KFParticle_sPHENIX/KFParticle_sPHENIX.cc
@@ -126,7 +126,7 @@ int KFParticle_sPHENIX::process_event(PHCompositeNode *topNode)
       if (m_verbosity > 0)
       {
         printParticles(mother[i], vertex[i], daughters[i], intermediates[i], nPVs, multiplicity);
-        if (m_save_dst) printNode(topNode);
+        if (m_save_dst && m_verbosity > 1) printNode(topNode);
       }
     }
   return Fun4AllReturnCodes::EVENT_OK;
--- a/offline/packages/KFParticle_sPHENIX/KFParticle_sPHENIX.cc
+++ b/offline/packages/KFParticle_sPHENIX/KFParticle_sPHENIX.cc
@@ -39,7 +39,6 @@ std::map<std::string, particle_pair> particleList = kfp_list.getParticleList();
 /// KFParticle constructor
 KFParticle_sPHENIX::KFParticle_sPHENIX()
   : SubsysReco("KFPARTICLE")
-  , m_verbosity(0)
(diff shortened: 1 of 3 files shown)


offline/packages/KFParticle_sPHENIX/KFParticle_sPHENIX.cc
```diff
@@ -126,9 +126,6 @@ int KFParticle_sPHENIX::process_event(PHCompositeNode *topNode)
       if (m_verbosity > 0)
       {
         printParticles(mother[i], vertex[i], daughters[i], intermediates[i], nPVs, multiplicity);
-      }
-      if (m_verbosity > 1)
```
blackcathj: Thanks for the update!

For Fun4All modules, there is a convenient verbosity flag in the base class: 
https://github.com/sPHENIX-Collaboration/coresoftware/blob/master/offline/framework/fun4all/Fun4AllBase.h#L65
that applies verbosity consistently for all modules.

```c++
 enum enu_Verbosity
  {

    //! Quiet mode. Only output critical messages. Intended for batch production mode.
    VERBOSITY_QUIET = 0,

    //! Output some useful messages during manual command line running
    VERBOSITY_SOME = 1,

    //! Output more messages
    VERBOSITY_MORE = 2,

    //! Output even more messages
    VERBOSITY_EVEN_MORE = 3,

    //! Output a lot of messages
    VERBOSITY_A_LOT = 4,

    // ... use your imagination ...

    //! Show all messages. Useful for step-by-step debugging
    VERBOSITY_MAX = INT_MAX - 10

  };
  /// Sets the verbosity of this module (0 by default=quiet).
  virtual void Verbosity(const int ival) { m_Verbosity = ival; }

  /// Sets the verbosity of this module (0 by default=quiet).
  virtual void Verbosity(enu_Verbosity ival) { m_Verbosity = ival; }

  /// Gets the verbosity of this module.
  virtual int Verbosity() const { return m_Verbosity; }
```

cdean-github: Thanks! I reverted the verbosity requirement back to the old version but I'll implement this in the next release as I assume I'll have a lot of requests during the MDC

cdean-github:
I've fixed the bug with missing tracks in the SvtxTrackMap and changed the verbosity logic so it is inherited from Fun4AllBase

PR #1052, opened 2021-01-09
Do not force Geant4 build options upon g4main libs

plexoos:
Certain releases of Geant4 may require specific C++ standards not compatible
with this project

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/simulation/g4simulation/g4main/Makefile.am
+++ b/simulation/g4simulation/g4main/Makefile.am
@@ -1,6 +1,5 @@
 AUTOMAKE_OPTIONS = foreign
 
-AM_CXXFLAGS = `geant4-config --cflags | sed s/-Woverloaded-virtual// | sed s/-W\ // | sed s/-Wshadow//`
 
 # set in configure.in to check if gcc version >= 4.8
 #if GCC_GE_48


pinkenburg:
This PR is problematic. There is a reason why the geant cflags are included - we would not go through this exercise for the fun of it.

If the geant build options are incompatible with our compiler version we will not be able to use that G4 version. This is the case for G4 10.06 already. G4 10.06 only works in conjunction with our default gcc 8.3. For SL7 (gcc 4.8.2) we use G4 10.02. This construct to use the geant cflags actually allows us to do this without ifdefs in our Makefile.

But the main reason for using the G4 cflags is that this package includes G4 includes. Specifically for selecting drivers for the G4 event display. Those includes use defines which are set by geant4-config --cflags and they do sometimes change with G4 version. Removing this will kill our G4 event display

plexoos:
And yet the tests have passed without this construct. I assume the current test builds include the gcc 8.3 + G4 10.06 and gcc 4.8.5 + G4 10.02 combinations. If this is the case, all one needs to say is that we want to support the older compiler without C++17 support. However, I am still puzzled how this code compiles with gcc 4.8.5
https://github.com/sPHENIX-Collaboration/coresoftware/blob/master/offline/packages/trackreco/PHActsSiliconSeeding.cc#L266
What am I missing?

pinkenburg:
The tests do not check the G4 event display (hard to do in batch anyway).

For code needing ACTS we use conditional compilation which only kicks in for gcc > 6. The configure.ac https://github.com/sPHENIX-Collaboration/coresoftware/blob/master/offline/packages/trackreco/configure.ac checks for the gcc version and sets MAKE_ACTS if it is >= 6.0. This is used in the Makefile.am https://github.com/sPHENIX-Collaboration/coresoftware/blob/master/offline/packages/trackreco/Makefile.am to compile those sources, including https://github.com/sPHENIX-Collaboration/coresoftware/blob/master/offline/packages/trackreco/PHActsSiliconSeeding.cc

The c++17 flag is added to the compiler flags by using a variable CONFIG_SITE which is set in the sphenix setup script to /cvmfs/sphenix.sdcc.bnl.gov/gcc-8.3/opt/sphenix/core/etc/config.site which tells configure to use those (also handy for 3rd party libs which use configure and don't have to be told). For gcc 4.8.4 is uses /cvmfs/sphenix.sdcc.bnl.gov/x8664_sl7/opt/sphenix/core/etc/config.site. This way all sources compiled with gcc 4.8 are compiled with -std=c++11, all sources compiled with gcc 8.3 get -std=c++17

The build script uses the sysname to decide if to build ACTS (which does not compile with gcc 4.8)

We will ditch the old compiler soon - but we cannot do this before the EIC yellow report is done.

plexoos:
It's good to know about all these special cases and unconventional workflows. Also I'll keep in mind that the existing tests are not representative

PR #1058, opened 2021-01-20
Adding true DCA calculations to KFParticle output (nTuple only)

cdean-github:
A request was made to calculate the true DCA in the nTuple output. This is calculated from the true track parameters and the VtxPoint, if it exists. If the true VtxPoint does not exist then the true DCA is calculated wrt to the reconstructed vertex and a warning is given.

I also added some new DCA branches in the nTuple: DCA_xy and DCA error to aid in resolution studies. Please note that the DCA_xy error is not calculated. I also changed the units of the decay time from ctau to ps.

Lastly, the magnetic field was set to negative 1.5T. I changed this to positive 1.5T which matches our magnetic field map plot but I should check that there isn't a flip in the sign somewhere.

## Types of changes

This PR introduces a new feature to the nTuple output of KFParticle if truth matching is enabled. This will not impact the MDC production but is wanted for user analysis.

## TODOs (if applicable)

### DIFF
--- a/offline/packages/KFParticle_sPHENIX/KFParticle_DST.cc
+++ b/offline/packages/KFParticle_sPHENIX/KFParticle_DST.cc
@@ -150,8 +150,9 @@ void KFParticle_DST::fillParticleNode_Track(PHCompositeNode* topNode, const KFPa
     }
   }
 
-  //I used to need to copy this track map to avoid crashes with truth matching
+  //When you build the daughter track, the track vanishes from the original track map, making a clone fixes this
   SvtxTrackMap* originalTrackMap = findNode::getClass<SvtxTrackMap>(topNode, "SvtxTrackMap");
+  SvtxTrackMap* originalTrackMap_copy = dynamic_cast<SvtxTrackMap*>(originalTrackMap->CloneMe());
   KFParticle* daughterArray = &daughters[0];
 
   for (unsigned int k = 0; k < daughters.size(); ++k)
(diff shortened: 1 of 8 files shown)


hupereir:
Hi @cdean-github 
On the last bullet, (magnetic field): it is true that the default magnetic field map corresponds to +1.5T however unless I am mistaken, the official sphenix (macros/detectors/sPHENIX/Fun4All_G4_sPHENIX.C) macro introduces a scale factor to that of -1.4/1.5, thus making the magnetic field effectively -1.4T. Maybe this is the value to be considered, or better, get the actual scale factor, namely G4MAGNET::magfield_rescale (= -1.4 / 1.5)
Hugo

cdean-github:
Hi @hupereir 

Ah, thanks! I wondered why I had put in a negative value. I can change this value to -1.4T instead.

Cameron

blackcathj:
@hupereir @cdean-github 

Good point. And here is a clean way to consistently get the field for vertex region, that is to use the field value at (0,0,0) which is a good approximation for near-vertex extrapolation:
```c++
PHField* field = PHFieldUtility::GetFieldMapNode(nullptr, topNode);
  if (field )
  {
    const double point[] = {0, 0, 0, 0};  // x,y,z,t
    double Bx = 0, By = 0, Bz = 0;
    double Bfield[] = {std::numeric_limits<double>::signaling_NaN(),
                       std::numeric_limits<double>::signaling_NaN(),
                       std::numeric_limits<double>::signaling_NaN(),
                       std::numeric_limits<double>::signaling_NaN(),
                       std::numeric_limits<double>::signaling_NaN(),
                       std::numeric_limits<double>::signaling_NaN()};
    field ->GetFieldValue(point, Bfield);
    Bx = Bfield[0];
    By = Bfield[1];
    Bz = Bfield[2];
    B_approximation_invertex_region = Bz / CLHEP::tesla;
}
```
In reference to https://sphenix-collaboration.github.io/doxygen/d0/d76/PHRaveVertexing_8cc_source.html#l00106 and https://sphenix-collaboration.github.io/doxygen/d2/d22/PHTpcTracker_8cc_source.html#l00212

When the simulation field changes, this value is changed consistently, as it is the same interfaced used by GenFit2 and Geant4.

PR #1074, opened 2021-02-05
Set Acts Vertex Finder to use SvtxVertexMap

osbornjd:

## Types of changes

This PR switches the ActsVertexFinder to use the SvtxVertexMap only, and then creates an SvtxVertexMapActs as a clone of the original vertex map. It also resets the track seed position to the vertex for the second fit pass

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackreco/PHActsVertexFinder.cc
+++ b/offline/packages/trackreco/PHActsVertexFinder.cc
@@ -93,7 +93,10 @@ int PHActsVertexFinder::Process(PHCompositeNode *topNode)
       delete track;
     }
 
-  m_svtxVertexMapActs = dynamic_cast<SvtxVertexMap*>(m_svtxVertexMap->CloneMe());
+  for(auto [key, svtxVertex] : *m_svtxVertexMap)
+    { 
+      m_svtxVertexMapActs->insert(dynamic_cast<SvtxVertex*>(svtxVertex->CloneMe()));
+    }
 
   if(Verbosity() > 0)
     std::cout << "Finished PHActsVertexFinder::process_event" << std::endl;
--- a/offline/packages/trackreco/PHActsVertexFinder.cc
+++ b/offline/packages/trackreco/PHActsVertexFinder.cc
@@ -313,6 +313,8 @@ void PHActsVertexFinder::fillVertexMap(VertexVector& vertices,
 	  const auto trackKey = keyMap.find(originalParams)->second;
 	  svtxVertex->insert_track(trackKey);
 
+	  const auto svtxTrack = m_svtxTrackMap->find(trackKey)->second;
+	  svtxTrack->set_vertex_id(key);
 	  updateTrackDCA(trackKey, Acts::Vector3D(vertexX,
 						  vertexY,
 						  vertexZ));
@@ -328,6 +330,17 @@ void PHActsVertexFinder::fillVertexMap(VertexVector& vertices,
       ++key;
     }
       
+
+
+  if(Verbosity() > 2)
+    {
+      std::cout << "Identify vertices in vertex map" << std::endl;
+      for(auto& [key, vert] : *m_svtxVertexMap)
+	{
+	  vert->identify();
+	}
+    }
+
   return;
 }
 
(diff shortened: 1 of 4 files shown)


osbornjd:
I also added a commit which fixes the units of the DCA uncertainties - they were being converted from Acts units even though they had already been converted 🤦

osbornjd:
Apparently the QA still didn't like the `SvtxVertexMapActs` clone, as seen [here](https://nbviewer.jupyter.org/github/sPHENIX-Collaboration/QA-gallery/blob/jenkins-sPHENIX-test-tracking-low-occupancy-qa-750-test-tracking_Event600_Sum10/QA-vertexing.ipynb#Vertex-resolution)...

offline/packages/trackreco/PHActsVertexFinder.cc
```diff
@@ -93,6 +93,8 @@ int PHActsVertexFinder::Process(PHCompositeNode *topNode)
       delete track;
     }
 
+  m_svtxVertexMapActs = dynamic_cast<SvtxVertexMap*>(m_svtxVertexMap->CloneMe());
```
blackcathj, line 96: I think this line is not putting the cloned vertex map on the node tree. 

We probably need something like
for (vertex in m_svtxVertexMap)
{
 m_svtxVertexMapActs . insert(vertex . clone() );
}
Or some variance of such

osbornjd: I'm confused, when I tested I added:

```
  std::cout << "svtx vertex map"<<std::endl;
  for(auto [key,vert] : *m_svtxVertexMap)
    vert->identify();
  std::cout<<"acts vertex map"<<std::endl;

  for(auto [key,actsVertex] : *m_svtxVertexMapActs)
    actsVertex->identify();

```

after doing the `dynamic_cast` in l96 and both vertices have the same `SvtxVertex` in them. If I do your suggestion of `m_svtxVertexMap->insert(vertex->CloneMe())` then the svtxVertexMapActs has two copies of each vertex in it.

blackcathj: Haha. There are two parts of the trick: 

> in l96 and both vertex maps have the same SvtxVertex in them. 

That is true. However, in l96, the SvtxVertex in `m_svtxVertexMap` is not copied onto the DST tree, but onto a new instance of `SvtxVertexMap` somewhere floating in the memory heap and its pointer is assigned to `m_svtxVertexMapActs`. The actual "DST/SvtxVertexMapActs" node is not updated, as now `m_svtxVertexMapActs` points to the new `SvtxVertexMap` object rather than the "DST/SvtxVertexMapActs" node.

To demonstrate this, when running valgrind, it reported definitely lost of memory at l96 (should be at the 2nd event):
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-default-detector-valgrind-pipeline/208/artifact/macros/detectors/sPHENIX/Fun4All_G4_sPHENIX.valgrind.xml/*view*/ search `PHActsVertexFinder.cc`
@pinkenburg we still need valgrind cleanup to be notified of new warnings like this  ;-) 


> If I do your suggestion of m_svtxVertexMap->insert(vertex->CloneMe()) then the svtxVertexMapActs has two copies of each vertex in it.

The suggestion is to remove l96, but replace it with the for loop: 
```c++
for (vertex in m_svtxVertexMap)
{
 m_svtxVertexMapActs . insert(vertex . clone() );
}
```
In this case, `m_svtxVertexMapActs` always points to the "DST/SvtxVertexMapActs" node. It is cleared at the beginning of the event by the Fun4All server. And in the above for loop, the "DST/SvtxVertexMapActs" node get fills of a single copy of vertexes from `m_svtxVertexMap`

osbornjd: > That is true. However, in l96, the SvtxVertex in m_svtxVertexMap is not copied onto the DST tree, but onto a new instance of SvtxVertexMap somewhere floating in the memory heap and its pointer is assigned to m_svtxVertexMapActs. The actual "DST/SvtxVertexMapActs" node is not updated, as now m_svtxVertexMapActs points to the new SvtxVertexMap object rather than the "DST/SvtxVertexMapActs" node.

Ah ha, that makes sense now. The pointer that the DST node originally was set to in create node was overwritten in l96, which is then why valgrind complains since there is a dangling pointer. I'll make the fix and add it to this PR.

PR #1118, opened 2021-03-24
use TDatabasePDG instead of G4ParticleTable to get charge from PID

hupereir:

Using G4ParticleTable to get particle charge from pid might result in a crash of the macro, because G4ParticleTable might not be initialized at the time of the first call. This happens for instance when running reconstruction only modules on DSTs containing G4Hits.

Instead of figuring out what needs to be initialized to prevent the crash, use TDatabasePDG instead of G4ParticleTable to get charge from PID, as done in e.g. QA/modules/QAG4SimulationTracking.cc

At the same time, reverted the changes to configure.ac about warning suppression, since they have become (hopefully) unnecessary here.

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackreco/PHTruthTrackSeeding.cc
+++ b/offline/packages/trackreco/PHTruthTrackSeeding.cc
@@ -30,8 +30,8 @@
 #include <phool/getClass.h>
 #include <phool/phool.h>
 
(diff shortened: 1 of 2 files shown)


blackcathj:
Error is due to today's update in ACTS repository. Rebuilding for the failed checks...

PR #1134, opened 2021-04-16
Tracking fixes

osbornjd:

## Types of changes

The weekly friday afternoon PR is here from the tracking side. This PR fixes several bugs from the fall out of the monday night merge which combined the Svtx and Acts objects. This PR fixes truth track seeding, GenFit track fitting, and the TPOT cluster adding to the acts track fits.

## TODOs (if applicable)

### DIFF
--- a/offline/packages/micromegas/MicromegasClusterizer.cc
+++ b/offline/packages/micromegas/MicromegasClusterizer.cc
@@ -394,6 +394,7 @@ int MicromegasClusterizer::process_event(PHCompositeNode *topNode)
         cluster->setError( i, j, error(i,j) );
       }
 
+      trkrClusterContainer->addCluster(cluster.release());
     }
 
   }
--- a/offline/packages/micromegas/MicromegasClusterizer.cc
+++ b/offline/packages/micromegas/MicromegasClusterizer.cc
@@ -209,9 +209,10 @@ int MicromegasClusterizer::process_event(PHCompositeNode *topNode)
 
       // create cluster key and corresponding cluster
       const auto cluster_key = MicromegasDefs::genClusterKey( hitsetkey, cluster_count++ );
+     
       auto cluster = std::make_unique<TrkrClusterv2>();
       cluster->setClusKey(cluster_key);
-      
+
       TVector3 world_coordinates;
       double weight_sum = 0;
 
@@ -311,6 +312,8 @@ int MicromegasClusterizer::process_event(PHCompositeNode *topNode)
       }
 
       /// Add Acts and local information
+      /// We need to use this generation of the hitsetkey as this is what is 
+      /// done in the geometry building, so then they match
       const unsigned int layer = TrkrDefs::getLayer(cluster_key);
       unsigned int tile = 0;
       MicromegasDefs::SegmentationType segtype;
@@ -323,6 +326,7 @@ int MicromegasClusterizer::process_event(PHCompositeNode *topNode)
           
       /// Get the surface key to find the surface from the map
       TrkrDefs::hitsetkey mmHitSetKey = MicromegasDefs::genHitSetKey(layer, segtype, tile);
+
       TrkrDefs::subsurfkey subsurfkey;
       auto surface = getMmSurfaceFromCoords(topNode,
 					    mmHitSetKey,
--- a/offline/packages/trackreco/PHActsTrkFitter.cc
+++ b/offline/packages/trackreco/PHActsTrkFitter.cc
@@ -18,6 +18,8 @@
 #include <trackbase_historic/SvtxVertexMap.h>
 #include <trackbase_historic/SvtxVertex.h>
 
+#include <micromegas/MicromegasDefs.h>
+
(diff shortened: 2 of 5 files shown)


offline/packages/micromegas/MicromegasClusterizer.cc
```diff
@@ -310,6 +311,15 @@ int MicromegasClusterizer::process_event(PHCompositeNode *topNode)
       }
 
       /// Add Acts and local information
+      /// We need to use this generation of the hitsetkey as this is what is 
+      /// done in the geometry building, so then they match
+      const unsigned int layer = TrkrDefs::getLayer(cluster_key);
```
hupereir, line 316: As far as I can tell, layer is unused.
If you want to keep it, it should be moved in the commented block below.
In fact I would vote to remove this comented code alltogether, since it is mileading: 
the segmentation is already retrieved upstream in the code, and should not be hardcoded to the layer number

offline/packages/trackreco/PHActsTrkFitter.cc
```diff
@@ -359,13 +361,31 @@ Surface PHActsTrkFitter::getSurface(TrkrDefs::cluskey cluskey,
...
+      const unsigned int layer = TrkrDefs::getLayer(cluskey);
+      unsigned int tile = 0;
+
+      MicromegasDefs::SegmentationType segtype;
+      if(layer == 55)
+	segtype  =  MicromegasDefs::SegmentationType::SEGMENTATION_PHI;
+      else
+	segtype = MicromegasDefs::SegmentationType::SEGMENTATION_Z;
+      hitsetkey = MicromegasDefs::genHitSetKey(layer, segtype, tile);
+
```
hupereir, line 379: If hard coding the segmentation type to the layer is really necessary then something is wrong
The difference between the hitsetkey from the cluster and the one you use for the surface is that for the former includes the tile id, while the later has tileid = 0 
So what you need to do is:
auto segmentation_type = MicromegasDefs::getSegmentationType( cluskey );
hitsetkey  = MicromegasDefs::genHitSetKey(layer, segtype, 0);
this is consistent with what is done in MicromegasClusterizer and with nothing hard coded.

hupereir (commented):
Some comments about the hitsetkey handling. I think the difference between the hitsetkey retrieve from clusters and the one you need to get the surface is that tileid == 0 in the later. (there is no tile in the current ACTS implementation of the micromegas geometry). However you do not need to hardcode the segmentation type to the layer number. 
(see MicromegasClusterizer.cc)

simulation/g4simulation/g4micromegas/PHG4MicromegasHitReco.cc
```diff
@@ -10,6 +10,7 @@
 
 #include <trackbase/TrkrDefs.h>
 #include <trackbase/TrkrHit.h>
+#include <trackbase/TrkrHitv2.h>
```
hupereir, line 13: Feels like both changes to this file are unnecessary. Please leave the file unchanged.

offline/packages/trackreco/PHActsTrkFitter.cc
osbornjd: I don't think hard coding the segmentation type is strictly necessary, so that can be changed. But there is still some fundamental difference about the hitkey generated for the hits and the hitkey generated for the Acts surfaces. The hit hitkeys are generated [here](https://github.com/sPHENIX-Collaboration/coresoftware/blob/be4c1bc32edfd4f0774198b6a7b12f025724e6b7/simulation/g4simulation/g4micromegas/PHG4MicromegasHitReco.cc#L269), and when we originally implemented the fake MM surfaces we generate the hitsetkey for identifying them [here](https://github.com/sPHENIX-Collaboration/coresoftware/blob/be4c1bc32edfd4f0774198b6a7b12f025724e6b7/offline/packages/trackreco/MakeActsGeometry.cc#L1134). Tony originally did this, so I don't know what the reason is to make all surfaces with tileid = 0 - we can ask him today. But I think this is where the difference comes in between the hitsetkeys used to identify the Acts surfaces and the hitsetkeys used to identify the actual hits

hupereir: Hi Joe,
so yes the difference between the two hitset ids is that in the one used for surface tileid=0. It makes sense since the ACTS surface don't know about the "tiles" (they are full acceptance cylinders, whereas the tiles are the 26 modules carved out of them) 
That is why you cannot use the hitsetid directly from the cluster and need to regenerate a new one; with tile =0 
"unsigned int tile =0" in the code above, or, (in order to spare you a temporary variable) calling hitsetkey = MicromegasDefs::genHitSetKey(layer, segtype, 0); directly
Hope it makes sense (and Tony can confirm in a few minutes)
So all in all the full code can be reduced to: 
(and I did check saturday that it leads to the same result):
const auto layer = TrkrDefs::getLayer(cluskey);
const auto segtype = MicromegasDefs::getSegmentationType( cluskey );
hitsetkey = MicromegasDefs::genHitSetKey(layer, segtype, 0);

PR #1136, opened 2021-04-19
Tracking fixes

osbornjd:

## Types of changes

This PR implements Hugo's comments from the last PR. Also, apparently Chris' revert didn't actually remove my commits from the master repository (??). So this PR only contains the commit corresponding to Hugo's comments from [1134](https://github.com/sPHENIX-Collaboration/coresoftware/pull/1134)

## TODOs (if applicable)

### DIFF
--- a/offline/packages/micromegas/MicromegasClusterizer.cc
+++ b/offline/packages/micromegas/MicromegasClusterizer.cc
@@ -310,14 +310,10 @@ int MicromegasClusterizer::process_event(PHCompositeNode *topNode)
       }
 
       /// Add Acts and local information
-      /// We need to use this generation of the hitsetkey as this is what is 
-      /// done in the geometry building, so then surfi\ace ID matches
       Acts::Vector3D globalPos(cluster->getX(), cluster->getY(), cluster->getZ());
-    
-      /// Get the surface key to find the surface from the map
-      TrkrDefs::hitsetkey mmHitSetKey = MicromegasDefs::genHitSetKey(layer, 
-								     segmentation_type, 0);
 
+      /// Get the surface key to find the surface from the map
+      TrkrDefs::hitsetkey mmHitSetKey = MicromegasDefs::genHitSetKey(layer, segmentation_type, 0);
       TrkrDefs::subsurfkey subsurfkey;
       auto surface = getMmSurfaceFromCoords(topNode, mmHitSetKey, subsurfkey, globalPos);
       if(!surface)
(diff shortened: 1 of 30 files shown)


offline/packages/micromegas/MicromegasClusterizer.cc
```diff
@@ -310,10 +310,16 @@ int MicromegasClusterizer::process_event(PHCompositeNode *topNode)
       }
 
       /// Add Acts and local information
-      Acts::Vector3D globalPos(cluster->getX(), cluster->getY(), cluster->getZ());
+      /// We need to use this generation of the hitsetkey as this is what is 
+      /// done in the geometry building, so then surfi\ace ID matches
+      const unsigned int actslayer = TrkrDefs::getLayer(cluster_key);
```
hupereir, line 315: This is not necessary
I have checked offline that the "actslayer" from here is == to the layer variable from above.
(to the difference that yours is unsigned int, while the original one is uint8_t. Note that using cout will give you different values, this is because cout of an ushort_8 is interpreted as a single character and not a number. But the underlying values _is_ the same)
If you stick to using layer instead of actslayer, you actually don't need any change to this file. 
Please remove 

Doing so, you need not change anything to this file

osbornjd: I see, I checked with cout and they appeared different which is why I left it like this - I didn't realize cout used different underlying representations

hupereir: you can do (to double check): std::cout << "acts layer: " << actslayer << " old: " << (unsigned int) layer << std::endl;
(or even put an assert(actslayer==layer); and you will see that the code does not crash)
This cout stuff is very misleading and bit me multiple times in the past.

offline/packages/micromegas/MicromegasClusterizer.cc
```diff
@@ -310,10 +310,14 @@ int MicromegasClusterizer::process_event(PHCompositeNode *topNode)
       }
 
       /// Add Acts and local information
+      /// We need to use this generation of the hitsetkey as this is what is 
+      /// done in the geometry building, so then surfi\ace ID matches
```
hupereir, line 314: Ce comment here is unrelated to the code below. 
Really no change here is relevant. I would prefer MicromegasCluster to be removed from this pull request entirely
Makes life easier later when tracking comits relevant to a given file while hunting bugs.

PR #1143, opened 2021-04-22
more tracking fixes

hupereir:
In order to fix GenFit track fitting with both truth or full track finding, one must make it agnostic to the version of the SvtxTrk object provided by the track finding. To achieve this one introduces a CopyFrom to the lastest SvtxTrk object that takes the base class as constructor, and makes use of it inside the GenFitTrackFitter.

CopyFrom method should also be implemented for the other older SvtxTrack objects. This will be the subject for another PR.

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackbase_historic/SvtxTrack.h
+++ b/offline/packages/trackbase_historic/SvtxTrack.h
@@ -52,10 +52,12 @@ class SvtxTrack : public PHObject
   using PHObject::CopyFrom;
   
   //! copy content from base class
-  virtual void CopyFrom( const SvtxTrack& ) {}
+  virtual void CopyFrom( const SvtxTrack& ) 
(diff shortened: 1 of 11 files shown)


hupereir:
From what I understand:
- the scanbuild error is unrelated (PHPointerList ?)
- the cppcheck error (uninitialized variable) is a false positive. The ACTS covariance matrix is initialized in the CopyFrom method, called in the constructor. I can initialize it here 'a second time'  to make cppcheck happy, but this should really not be necessary.

hupereir:
So, warning on _acts_trajectory_covariance is gone. 
If new cppcheck version also ignore the other warnings, then we should be good to go. 
(or is it the new version that ran already ?)

PR #1144, opened 2021-04-23
compiler check needs to use *g++ for daily build and jenkins/user to …

pinkenburg:
…behave identically

## Types of changes
- [x ] Bug fix (non-breaking change which fixes an issue)

This PR changes the check for the CXX compiler. It used to be g++ for g++ but the new gcc-8.3 compiler comes with the full path /cvmfs/..../g++ which does not match this. The build cron uses a special setup where g++ is passed down as the compiler. The result was that jenkins and user builds missed the g++ settings which typically add -Werror, so code passed jenkins and bombed out in the daily builds. This PR uses now *g++ which matches both. It has to come after the entry for clang++ since *g++ also matches clang++

## TODOs (if applicable)

### DIFF
--- a/generators/PHPythia8/configure.ac
+++ b/generators/PHPythia8/configure.ac
@@ -14,7 +14,7 @@ dnl specific flags
   clang++)
    CXXFLAGS="$CXXFLAGS -Wall -Werror -Wno-c11-extensions"
   ;;
-  g++)
+  *g++)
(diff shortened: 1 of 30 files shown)


hupereir:
Hi Chris, 

Why not change also configure.ac in g4eval ?

pinkenburg:
g4eval mast have fallen through the cracks. Used sed for this, not sure how this happened

PR #1148, opened 2021-04-26
KFParticle: Adding decay parser, new output and code polishing

cdean-github:

## Types of changes

2 PRs from 03/18/21 and 03/26/21 have become entangled in this request.

**Cleaning and bug fixing:**

There is some general code polishing and scope improvements (such as calling the vertex map when truth matching). Unused headers are removed (I left some in when testing the calorimeter information). Many variables which were initialised in the constructor have been taken out as they are now vectors, this caused issues when trying to recover the user values.

Fixed a problem where track masses could be overwritten with 0 GeV. Parenthesis was in wrong point.

**New features:** 

Track-vertex DCA and quality is calculated for all PVs in the event. Intermediate selection criteria can now be specified in a range, useful for prompt/non-prompt decays.

Array declarations have been removed and changed with vectors. This should make clang happy. I checked that the input parameters are still getting passed around as required.

ROOT does not like certain characters so if they appear I swap them out with OK characters. This used to be taken care of on a case-by-case basis and is now fixed using a map that is iterated over, less error prone this way.

Added a new function where we can accept tracks that have zero mass (previously they were rejected). KFParticle (the base class) likes to randomly forget the mass of tracks which impacts our selection efficiency. I need to email the authors about this.

**_FOR ECCE:_** Added a new function where we can create a dummy primary vertex instead of reading the vertex map. I've noticed that our ECCE software likes to not fill the vertex map, I don't know why but this causes KFParticle to skip all events. Setting "useFakePrimaryVertex" will get around this but we should figure out why the vertex map is empty, fix that, and then remove this function.

Particle list has been updated to give better functionality in the parser.

Added branches to give a tracks full truth history from the HepMC record. Starts listing from parent and goes down to initial quark. Basically you can tell if your track came from the parent you think it did and whether it is a prompt or non-prompt produced particle.

**MAJOR NEW FEATURE:** Decay parser has been added to the package.

Users can now write a string which describes the decay they are looking for (charge-conjugates, intermediate decays, charges etc). This replaces the old method of using several functions to define the type of decay and is far more user friendly, e.g. the string "B_s0 -> {J/psi -> e^+ e^-} K^+ K^-" would be used to search for this specific b-hadron decay

## TODOs (if applicable)

1. Contact KFParticle developers about tracks randomly being assigned a mass of zero
2. Figure out why ECCE doesn't fill the vertex map (maybe this is already fixed)
3. Add the ability to tell how many decay products a parent had. Useful to identify if our candidate is fully reconstructed or not


offline/packages/KFParticle_sPHENIX/KFParticle_sPHENIX.h
```diff
@@ -88,13 +95,13 @@ class KFParticle_sPHENIX : public SubsysReco, public KFParticle_nTuple, public K
     m_num_tracks_nTuple = num_tracks;
   }
 
-  void setNumberTracksFromIntermeditateState(int num_tracks[max_particles])
+  void setNumberTracksFromIntermeditateState(int num_tracks[max_particles] = 0)
```
blackcathj, line 98: Setter function like this probably can be netter served with stl vector instead of the C arrays: 
e.g. 
`void setNumberTracksFromIntermeditateState(const std::vector<int> & num_tracks)`

It serves a few purposes: 
1. Macros can make a vector and push more value into them or init with `{...}`. So I think easier to use
2. Memory safe and always initialized
3. Avoids built in arbitrary limit such as `max_particles`

cdean-github: I agree. This is an older bit of code and I think I put a bit too many array-like accesses attached to it. Changing this to std::vector will likely cause a lot of errors and changes which I didn't want to chase down before getting a new release. This error was unexpected but probably shows I need to change the code around. 

I was already experimenting with vectors in user functions (https://github.com/sPHENIX-Collaboration/analysis/blob/master/HF-Particle/AntiTrigger/AntiTrigger.h#L47) so I'll just need to sit down for a couple of hours and go through it all.

blackcathj:
manually rebuild with https://github.com/sPHENIX-Collaboration/macros/pull/413 : 
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/Build-Master-gcc8/1335/

PR #1173, opened 2021-05-17
New micromegas geometry acts

hupereir:
This is a follow up to https://github.com/sPHENIX-Collaboration/coresoftware/pull/1166. It implements the new Micromegas geometry in ACTS. Rather than having ACTS Micromegas surfaces defined as portion of cylinders that cover the full TPC acceptance, there is now on surface per micromegas module. This makes the correspondance from hitsetid to ACTS Micromegas surface straightforward, and fixes the problem of ACTS not picking the right surface when handling 1D measurements, located at the middle of the strip, for the other coordinate.
All tracking and QA code adjusted accordingly
See https://indico.bnl.gov/event/11903/contributions/49766/attachments/34569/56076/talk.pdf for details
New geometry in ACTS: 
image

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/QA/modules/QAG4SimulationMicromegas.cc
+++ b/offline/QA/modules/QAG4SimulationMicromegas.cc
@@ -60,6 +60,20 @@ namespace
     double weight = 1;
   };
   
+ //! rotate size or covariant matrix to polar coordinates and return component along provided phi angle
+ /** this is copied from TrkrClusterv2.cc, with the difference that one must use the tile azimuthal angle instead of that of the cluster */
+ template<float (TrkrCluster::*accessor)(unsigned int, unsigned int) const>
+    float rotate( const double phi, const TrkrCluster* cluster )
+  {
+    const auto cosphi = std::cos(phi);
+    const auto sinphi = std::sin(phi);
+
+    return
+      square(sinphi)*(cluster->*accessor)(0,0) +
+      square(cosphi)*(cluster->*accessor)(1,1) +
+      2.*cosphi*sinphi*(cluster->*accessor)(0,1);
+  }
+  
   //! calculate the interpolation of member function called on all members in collection to the provided y_extrap
   template<double (interpolation_data_t::*accessor)() const>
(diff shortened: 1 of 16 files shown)


osbornjd (commented):
Thanks for implementing this Hugo, this is really great to see for the MMs. I have one question/comment that we can choose to address now or at a later time in a future PR. It looks like currently the micromegas are still just "created" by modifying the Geant geometry with boxes, like what is done for the TPC (am I incorrect about this?). At some point, we should build the micromegas just like the silicon layers by letting Acts do everything itself. This will save some code on our side and also will delegate as much as possible to Acts, which is probably best in the long run so that whatever Acts is doing is self consistent. 

Again, this may be a longer term thing, which is totally fine. Just thought I would make the comment now while the PR is open.

hupereir:
> Thanks for implementing this Hugo, this is really great to see for the MMs. I have one question/comment that we can choose to address now or at a later time in a future PR. It looks like currently the micromegas are still just "created" by modifying the Geant geometry with boxes, like what is done for the TPC (am I incorrect about this?). 

correct

> At some point, we should build the micromegas just like the silicon layers by letting Acts do everything itself. This will save some code on our side and also will delegate as much as possible to Acts, which is probably best in the long run so that whatever Acts is doing is self consistent.

Agreed. The only temporary issue with that is that the current GEANT4 implementation, which is used to generate the G4Hit, still uses cylinders (and thus cannot be passed to ACTS directly)
The reason for this is that the actual Micromegas configuration (number and position of the modules (=tiles) or even their size, is not final. We want to be able to modify thoses without having to rerun geant. Hence the two step process: first full acceptance cylinders, at G4Reco stage, then convertion to modules and synchronization of ACTS with those, at HitReco stage. Does that make sense ? 

When the micromegas configuration is final, it will be integrated to our GEANT and then I agree that the MakeActsGeom part should go.

Hugo

osbornjd:
Sounds perfect, just wanted to make the comment and make sure my understanding was correct.

PR #1200, opened 2021-06-07
suppress warning

osbornjd:

## Types of changes

Add a verbosity statement to suppress this warning when micromegas aren't actually used

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackreco/MakeActsGeometry.cc
+++ b/offline/packages/trackreco/MakeActsGeometry.cc
@@ -1543,7 +1543,7 @@ int MakeActsGeometry::getNodes(PHCompositeNode *topNode)
   // load micromegas geometry
   // do not abort if not found
   m_geomContainerMicromegas = findNode::getClass<PHG4CylinderGeomContainer>(topNode, "CYLINDERGEOM_MICROMEGAS_FULL");
-  if (!m_geomContainerMicromegas)
+  if (!m_geomContainerMicromegas && Verbosity() > 0)
   {
     std::cout << PHWHERE 
 	      << " CYLINDERGEOM_MICROMEGAS_FULL  node not found on node tree"


hupereir:
Hi Joe,

I would rather have this hidden behind a flag of type "!m_useMicromegas", rather than verbosity
The rationale is that if you do plan to work with the micromegas, and the geometry is not found, then there is something seriously wrong, and you should get warned about it, disregading the verbosity of MakeActsGeom. 

Not that for the other modules, (e.g. the TPC), the code actually terminates if the geometry is not found. (which is quite more drastic than a one time warning)

Setting the "m_useMicromegas" flag in MakeActsGeom could then be conditionned to if( Enable::MICROMEGAS) in the calling macro (G4_Tracking.C)

hupereir:
And in fact, reading the code, it is probably enough to condition it to "m_buildMMs", which is set in ::editTPCGeometry()
having both m_buildMMs to true and MICROMEGAS_FULL not found, should never happen.

osbornjd:
The current problem is that this is printed out for users who aren't using the MMs and they are then thinking there is some problem with what they are running.

I agree with you that this could just be hidden behind `m_buildMMs`; however, this wouldn't solve the problem. For example, we could add the following:
```
if(m_geomContainerMicromegas)    {     
  m_buildMMs = true;      
  std::cout << " WILL  add MMs surfaces to ActsSurfaceMap " << std::endl;     
 }
else {
   std::cout << "CYLINDERGEOM_MICROMEGAS_FULL node not found on node tree" << std::endl;
}
```
[here](https://github.com/sPHENIX-Collaboration/coresoftware/blob/965c2842aa8f1fb84a1ad6e24a31bf92eee15f48/offline/packages/trackreco/MakeActsGeometry.cc#L158). But that will still print out every time a user who is not using the MMs runs a job, potentially causing confusion (as it has with one user who asked me about it). How can we get both of these things?

hupereir:
Sorry for the many comments. Thinking more, I think the whole logic of how Micromegas are handled in MakeActsGeom is flawed. 
Instead of trying to "guess" whether Micromegas are enabled or not, there should be an externally set flag set for including or not the Micromegas, based on Enable::MICROMEGAS in G4_Tracking
If this flat is not set, all the micromegas part should be disabled and you would have no warning.
If it is set, code should warn/bump/etc in case of failure to find the relevant nodes. 
Does that make sense ?

hupereir:
> The current problem is that this is printed out for users who aren't using the MMs and they are then thinking there is some problem with what they are running.
Ultimately that is the problem: MM are now on by default (and should be). That you get some warnings when you manually turn them off, is ok I think
At least as ok as the warnings (or aborts) you get if you turn of the MVTX (Enable::MVTX = false) (just tried)
I don't see why you should suppress these warnings (more than those coming for other detectors)

For what its worth: the Micromegas code has been broken multiple times over the last few month (and it has taken several weeks of work to unbrake this) in part due to users running with the MM turned off. That one gets a warning or two here and there when deciding to do so is IMHO a small price to pay wrt to the above.

osbornjd:
I guess the question is one of philosophy - what is the right number of warning messages to print out that may or may not worry users. I can close this PR if you think it should be left as is - it is my personal experience that warning messages can confuse people, so the intent was only to reduce the potential confusion.

hupereir:
What I want to make sure is that: 
if my macro has Enable::MICROMEGAS = true, and there is something wrong so that micromegas surfaces are not built in MakeActGeom, I get a warning, disregarding the verbosity, because it means something is wrong and needs fixing. 
This PR does not allow that, so yes I would be in favor of closing it.

If now one does not want a warning when Enable::Micromegas = false, and CYLINDERGEOM_MICROMEGAS_FULL is not found (which is indeed ok), then further changes are needed that are not included in this PR. The easiest way to achieve this is to make m_buildMM flag configurable, and set to Enable::Micromegas externally. And then, if set to true externally, MakeActsGeom should require the relevant nodes to be found (with warnings and bumps if not).

The same should then also be done for the other subsystems. Right now if you set Enable::MVTX = false, you cannot run the ACTS tracking at all. MakeActsGeom bumps with a warning that the corresponding Geo node is not found. Which makes sense since our whole tracking relies on the presence of the MVTX (just like our calibration strategy for the TPC relies on the presence of the Micromegas), but then one can question the usefulness of said flag. 

If need be, we can discuss that further during tomorrow's simulation and software meeting

osbornjd:
I don't think it's necessary to discuss, I can close it and we'll just leave it as is since I don't think having a flag to build the MMs is a long term solution either (as then the MMs are still on different ground from e.g. the TPC)

PR #1219, opened 2021-06-19
Add event level info to event evaluator

raymondEhlers:

Add event level info to the event evaluator (xsec, event weight, etc). To do so, the Pythia8 event weight is now stored in the HepMC event. Based on how these weights are used, I don't think this is used, but to be safe, the option is off by default.

Also add some track DCA values

@FriederikeBock 

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/generators/PHPythia8/PHPythia8.cc
+++ b/generators/PHPythia8/PHPythia8.cc
@@ -45,7 +45,7 @@ PHPythia8::PHPythia8(const std::string &name)
   , m_Pythia8(nullptr)
   , m_ConfigFileName("phpythia8.cfg")
   , m_Pythia8ToHepMC(nullptr)
-  , m_SaveEventWeightFlag(false)
+  , m_SaveEventWeightFlag(true)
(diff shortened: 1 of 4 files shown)


generators/PHPythia8/PHPythia8.cc
```diff
@@ -45,6 +45,7 @@ PHPythia8::PHPythia8(const std::string &name)
   , m_Pythia8(nullptr)
   , m_ConfigFileName("phpythia8.cfg")
   , m_Pythia8ToHepMC(nullptr)
+  , m_SaveEventWeightFlag(false)
```
blackcathj, line 48: Thanks! @raymondEhlers  
This is useful to many and would you like to make this option true by default?

raymondEhlers: Sure - I just pushed an update to set it on by default

blackcathj:
The `valgrind` error appears come from an earlier PR . Merging

PR #1281, opened 2021-09-11
Ghost track rejector module

adfrawley:

## Types of changes

This PR adds a module to reject duplicate ("ghost") tracks. It can be run on the TPC track seeds after vertex association, or on the full tracks after the first track fitter is run. Because of the way the CA seeder works, it is very efficient but produces roughly two (usually identical) seeds per truth track. This module compares (x,y,z), phi and eta for every pair of tracks. When it finds matches, it keeps the track having the best chisq/ndf and deletes the others. Tests with MB + 50 KHz pileup events show that tracking efficiency (defined as fraction of truth tracks reconstructed) is not affected.
Because this module deletes tracks from the track map, it was necessary to modify any code that finds the last key in the track map. The last key is no longer the track map size - 1. The last key is now std::prev(_track_map->end())->first. 
An unrelated change to the evaluator makes the cluster ntuple "hitID" the cluster index (rather than a cast of the cluster key to float, which is gibberish). For a given hitset (phi element and z element), the cluster index is unique.
Running the ghost rejection module on the TPC seeds reduces the time taken for all subsequent steps by about a factor of two, so that is the preferred use.
I will make a macros PR later. 

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackreco/Makefile.am
+++ b/offline/packages/trackreco/Makefile.am
@@ -44,6 +44,7 @@ pkginclude_HEADERS = \
   PHActsVertexFitter.h \
   PHActsTrackProjection.h \
   PHGenFitTrkProp.h \
+  PHGhostRejection.h \
   PHHoughSeeding.h \
   PHTrackCleaner.h \
   PHRTreeSeeding.h \
@@ -168,6 +169,7 @@ libtrack_reco_la_SOURCES = \
   PHGenFitTrkProp.cc \
   PHGenFitTrkFitter.cc \
   PHGenFitTrackProjection.cc \
+  PHGhostRejection.cc \
   PHRaveVertexing.cc \
   PHTpcTrackSeedVertexAssoc.cc \
   ALICEKF.cc \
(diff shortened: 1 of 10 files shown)


adfrawley:
The failure reported on the second Jenkins go-round does not seem to have anything to do with this PR (nothing reported under "new").

blackcathj:
> The failure reported on the second Jenkins go-round does not seem to have anything to do with this PR (nothing reported under "new").

Hi Tony

The compilation was fine but the default sPHENIX macro failed to execute with this build. Here is the output. Do you see something similar in your tests? 

https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-default-detector-pipeline/3606/artifact/macros/detectors/sPHENIX/Fun4All_G4_sPHENIX.log/*view*/

```
Processing Fun4All_G4_sPHENIX.C(30)...
...
[0;33m[TwoTrackMinimumDistance] comparing track with itself![;0m
[0;33m[TwoTrackMinimumDistance] Computation HelixHelix::CAIR failed.[;0m
18:19:39    HelTrkLinPro   ERROR     Step failed: EigenStepperError:3
Command terminated by signal 11
```

PR #1295, opened 2021-09-30
TrkrCluster Memory

osbornjd:

## Types of changes

This PR changes the `TrkrCluster` object to a much reduced memory format than before. Previously, The object was holding both local and global information. The global information is inherently more memory hungry than the local information, so this PR removes it and makes the entire tracking chain identify the global information from the Acts surface transformation from the local information.

There are many changes in this PR so I expect Jenkins to pull something up. I am also opening it as I'd like to see the QA results and make sure I have not broken/changed anything drastically. Please DO NOT merge until I give the okay.

## TODOs (if applicable)

### DIFF
--- a/offline/packages/intt/InttClusterizer.cc
+++ b/offline/packages/intt/InttClusterizer.cc
@@ -3,7 +3,7 @@
 #include "InttDefs.h"
 
 #include <trackbase/TrkrClusterContainerv3.h>
-#include <trackbase/TrkrClusterv2.h>
+#include <trackbase/TrkrClusterv3.h>
 #include <trackbase/TrkrDefs.h>
 #include <trackbase/TrkrHitSet.h>
 #include <trackbase/TrkrHitv2.h>
@@ -362,7 +362,7 @@ void InttClusterizer::ClusterLadderCells(PHCompositeNode* topNode)
 	
 	// make the cluster directly in the node tree
 	TrkrDefs::cluskey ckey = InttDefs::genClusKey(hitset->getHitSetKey(), clusid);
-	auto clus = std::make_unique<TrkrClusterv2>();
+	auto clus = std::make_unique<TrkrClusterv3>();
 	clus->setClusKey(ckey);
 
 	if (Verbosity() > 2)
@@ -475,10 +475,6 @@ void InttClusterizer::ClusterLadderCells(PHCompositeNode* topNode)
 
 	// Fill the cluster fields
 	clus->setAdc(clus_adc);
-	clus->setPosition(0, clusx);
-	clus->setPosition(1, clusy);
-	clus->setPosition(2, clusz);
-	clus->setGlobal();
 	
 	TMatrixF DIM(3, 3);
 	DIM[0][0] = square(0.5 * thickness);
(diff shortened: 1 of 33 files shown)


hupereir:
Hi Joe, 
That looks promising !
Some questions, comments:
- any chance you also fix the code in tpccalib, in particular PHTpcResiduals ? 
- I see you modified all the clusterizers except MicromegasClusterizer. Does that mean it needs no change or was it just left out ? 
Looking at the code change I could probably implement this myself once this PR lands, but it might better be part of the same PR instead. 
- I guess this PR kills the GenFit toolchain once and for all. Does it make sense to then deprecate it once and for all, and remove it from compilation ?

osbornjd:
Thanks for the comments Hugo, responses in line.

> any chance you also fix the code in tpccalib, in particular PHTpcResiduals ?
> I see you modified all the clusterizers except MicromegasClusterizer. Does that mean it needs no change or was it just left out ?
> Looking at the code change I could probably implement this myself once this PR lands, but it might better be part of the same PR instead.
No, I just always forget about the TPOT and SCD and thus left it out inadvertently. Thanks for catching this - I will add them as a part of this PR as things will probably break without it. I need to make some changes anyway to the PR, I believe.
> I guess this PR kills the GenFit toolchain once and for all. Does it make sense to then deprecate it once and for all, and remove it from compilation ?
I think we have to keep it in for the EIC, but we could potentially remove it from the sPHENIX build. I would argue we should keep it around though, just in case we need it for benchmarking. AFAIK it doesn't "hurt" anything to keep it around.

osbornjd:
Clang still seems to barf with an issue with Acts/Eigen, which I don't understand:

https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/Build-Clang/2275/console

osbornjd:
Jenkins shows that the momentum resolution degrades from this PR. We can try and fix the clang issue but this is going back to WIP, I need to understand why the momentum resolution degrades

hupereir:
Hi Joe, 
There are also some problems with the residuals in the various detector specific QA histograms (INTT, MVTX and MicromegaS) (new histograms are empty). Probably because the corresponding QA modules need update to handle global cluster coordinates. ( similar to what is done in the evaluators)

osbornjd:
My local tests seem to indicate everything looks like the main repository right now. Let's see what Jenkins says

osbornjd:
That may have been the fastest Jenkins failure ever. 1 minute between trigger and failure 😂

osbornjd:
@blackcathj I don't understand why this PR does not pass the builds at the very least. It builds locally for me and based on the logs it seems the build goes through without errors on Jenkins too. Do you have any insight?

pinkenburg:
The failure is slightly misleading. It fails when any of its subsequent (indented) tasks fails. In this case valgrind found 2 invalid reads which makes it fail and then sets the gcc-8.3/new to failed

hupereir:
Please don't update the QA plots yet with new reference though !
I would like to recover the Micromegas rphi uncertainty first
(should not take more than a couple of days, now that the code have actually been merged)

PR #1333, opened 2021-11-03
Add oHCAL tile map response

johnlajoie:

This PR uses the oHCAL tile map response measured by MEPHI to modify the oHCAL light yield generated from energy deposits in the scintillator to reflect the measured non-uniformity in the scintillating tiles. 

## Types of changes
- [X] New feature (non-breaking change which adds functionality)
- [X] I am a member of [GitHub organization of sPHENIX Collaboration](https://github.com/orgs/sPHENIX-Collaboration/people), EIC, or ECCE (contact Chris Pinkenburg to join)

This PR adds a more realistic light yield response from the oHCAL scintillating tiles based on the measured tile response. As such, the light yield will more accurately reflect the features of the fiber embedded in the tile, proximity to the light blocker, etc. Because the scintillator in G4 is implemented as a single subdivided volume (no boundaries) this will also have the effect of introducing boundaries between the tiles. 

## TODOs (if applicable)

At this point this is a draft pull request for testing with Jenkins. I'm not sure if the tile response will change things enough that Jenkins will puke, but it will be good to see. 

Something else that could also be considered would be interpolating between the map bins (0.5cm).  We thought about this, but decided against it.  There are fluctuations in the response along the boundaries that would propagate if we did that and they would potentially need to be smoothed.  The 0.5cm size is much larger than the typical hadronic shower scale that we felt is was better not to overly-process and massage the map. 

Currently running a large sample with the same random seed to compare directly the tile response in local coordinates with and without the map response added.  Will add that to the PR when completed. 

The outer HCAL tile map itself was introduced by the pull request: 

https://github.com/sPHENIX-Collaboration/calibrations/pull/85

which has already been merged,

### DIFF
--- a/simulation/g4simulation/g4detectors/PHG4OuterHcalSteppingAction.cc
+++ b/simulation/g4simulation/g4detectors/PHG4OuterHcalSteppingAction.cc
@@ -23,8 +23,10 @@
 // Root headers
 #include <TAxis.h>                             // for TAxis
 #include <TH2.h>
+#include <TH2F.h>
 #include <TNamed.h>                            // for TNamed
 #include <TSystem.h>
+#include <TFile.h>
 
 // Geant4 headers
 
@@ -47,6 +49,7 @@
 #include <Geant4/G4VPhysicalVolume.hh>         // for G4VPhysicalVolume
 #include <Geant4/G4VTouchable.hh>              // for G4VTouchable
 #include <Geant4/G4VUserTrackInformation.hh>   // for G4VUserTrackInformation
+#include <Geant4/G4Transform3D.hh>
 
 // finally system headers
 #include <cassert>
@@ -58,6 +61,9 @@
 class PHCompositeNode;
 
 using namespace std;
+
+TH2F *MapCorr = NULL;
+
 //____________________________________________________________________________..
 PHG4OuterHcalSteppingAction::PHG4OuterHcalSteppingAction(PHG4OuterHcalDetector* detector, const PHParameters* parameters)
   : PHG4SteppingAction(detector->GetName())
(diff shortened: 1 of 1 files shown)


blackcathj:
The results are actually pretty consistent, though bins does migrate around after the photon generation step as expected.

The PR QA would not have enough statistics to tell fine difference, e.g. hadronic energy resolution, though. 

image

image

johnlajoie:
This really looks about as I expected. the maxium deviation (from minimum to maximum) within a tile can be quite large (more than a factor of two), but this is looking at the extreme points in the tile maps. As a shower samples the tiles a lot averages out and you are just left with the secondary effect of did you sample a little more in one area than another.  The end result is the actual effect on things like energy distributions should look small. 

What will be very interesting to look at will be the constant term in the energy resolution. I've long suspected that the tile non-uniformity will show up there at some level. 

Let me just pull together the ratio plot from our tests and add that to the PR as a double-check. Once that is done I think we should be good to merge. 

Ejiro has started work on the iHCAL tile maps;  the procedure will mirror this one.

johnlajoie:
Here's the result of running 1000 single pion events (5 per event) in two ways - the top panel is without the new tile response map, the bottom is with the new tile response map.  These events were run with the same initial random seed so they are basically the same events modulo the tile response - this makes it easier to see the effect in a smaller event sample. 

compare

Hard to see the difference by eye.  If I take the ratio:

ratio

Now you can clearly see the effect of the tile response map.

johnlajoie:
I had a quick look at the valgrind report but the complaints seem to be in other parts of the simulation code. I think this is ready to merge.

PR #1341, opened 2021-11-17
Tpc distortion correction

hupereir:
[comment]: <> (Please tell us something about this pull request)
This PR introduces new classes to handle TPC distortion corrections in the tracking on a "on-demand" basis
Distortion corrections are loaded on the node tree at initialization via TpcLoadDistortionCorrection. Distortion corrections are stored in TpcDistortionCorrectionContainer

Subsysreco in need of distortion correction must retrieve the TpcDistortionCorrectionContainer from the node tree, and use the utility class TpcDistortionCorrection to correct a given 3D position in the TPC.

As an illustration, TPC distortion corrections have been implemented in PHCaSeeding, immediately after global coordinates are retrieved for a given cluster, via ActsTransform.

Also 
- removed broken (and ill-named) TpcSpaceChargeCorrection module
- cleaned up the PHCASeeding code a bit, to remove unused variables and code, using namespace std, and to add a number of const& that should prevent unnecessary object copies. 

There will be a separate commit to the macro directory needed to account for the change. 
Without this change, Jenkins will break, because of the removed subsysreco

## Types of changes
[comment]: <> ( What types of changes does your code introduce? Put an `x` in all the boxes that apply: )
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work for users)
- [x] Requiring change in macros repository (Please provide links to the macros pull request in the last section)
- [x] I am a member of [GitHub organization of sPHENIX Collaboration](https://github.com/orgs/sPHENIX-Collaboration/people), EIC, or ECCE (contact Chris Pinkenburg to join)

## What kind of change does this PR introduce? (Bug fix, feature, ...)

[comment]: <> ( What does this PR do? Linking to talk in software meeting encouraged )

## TODOs (if applicable)

[comment]: <> ( In case this is a draft PR, e.g. for running checks using Jenkins, please make the pull request as a draft: https://github.blog/2019-02-14-introducing-draft-pull-requests/  )

## Links to other PRs in macros and calibration repositories (if applicable)
https://github.com/sPHENIX-Collaboration/macros/pull/466

PR #1343, opened 2021-11-18
Cache global in TPC Seeding

osbornjd:

## Types of changes

Caches global cluster positions so that cluster position look up isn't performed many times. Improves speed of modules without loss in functionality.

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackbase_historic/ActsTransformations.cc
+++ b/offline/packages/trackbase_historic/ActsTransformations.cc
@@ -258,7 +258,13 @@ void ActsTransformations::calculateDCA(const Acts::BoundTrackParameters param,
   
 }
 
-
+Acts::Vector3F ActsTransformations::getGlobalPositionF(TrkrCluster* cluster,
+						       ActsSurfaceMaps* surfMaps,
+						       ActsTrackingGeometry* tGeometry) const
+{
+  const Acts::Vector3D doublePos = getGlobalPosition(cluster,surfMaps,tGeometry);
+  return Acts::Vector3F(doublePos(0), doublePos(1), doublePos(2));
+}
 Acts::Vector3D ActsTransformations::getGlobalPosition(TrkrCluster* cluster,
 						      ActsSurfaceMaps* surfMaps,
 						      ActsTrackingGeometry *tGeometry) const
--- a/offline/packages/trackbase_historic/ActsTransformations.h
+++ b/offline/packages/trackbase_historic/ActsTransformations.h
@@ -68,7 +68,10 @@ class ActsTransformations
 			   const size_t &trackTip,
 			   SvtxTrack *svtxTrack,
 			   Acts::GeometryContext geoContext) const;
-
+  
+  Acts::Vector3F getGlobalPositionF(TrkrCluster* cluster,
+				    ActsSurfaceMaps* surfMaps,
+				    ActsTrackingGeometry *tGeometry) const;
   Acts::Vector3D getGlobalPosition(TrkrCluster* cluster,
 				   ActsSurfaceMaps* surfMaps,
 				   ActsTrackingGeometry *tGeometry) const;
--- a/offline/packages/trackreco/ALICEKF.cc
+++ b/offline/packages/trackreco/ALICEKF.cc
@@ -867,7 +867,6 @@ void  ALICEKF::line_fit(std::vector<std::pair<double,double>> points, double &a,
 void  ALICEKF::line_fit_clusters(std::vector<TrkrCluster*> clusters, std::vector<Acts::Vector3F>& globalPositions, double &a, double &b)
 {
   std::vector<std::pair<double,double>> points;
-  ActsTransformations transformer;
   for (unsigned int i=0; i<clusters.size(); ++i)
     {
       const auto globalpos = globalPositions.at(i);
--- a/offline/packages/trackreco/ALICEKF.cc
+++ b/offline/packages/trackreco/ALICEKF.cc
@@ -44,7 +44,7 @@ double ALICEKF::get_Bz(double x, double y, double z)
   return bfield[2]/tesla;
 }
 
-double ALICEKF::getClusterError(TrkrCluster* c, int i, int j)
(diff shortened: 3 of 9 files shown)


hupereir:
Hi Joe,
The changes to PHCASeeding will likely conflict with https://github.com/sPHENIX-Collaboration/coresoftware/pull/1341, depending on what gets pushed first ...
Also: isn't this commit essentially equivalent to going back to storing all global position in the clusters, at least for the TPC clusters which dominate anyway ? 
What is the gain, then to have only the local ACTS coordinates in the clusters ? 
In more detail: I expect modules like the CA Seeder to loop over all TPC clusters at least one, so that essentially all global coordinates get cached, and we are left to having (at least for all the time spent seeding) both the global and local coordinates stored in memory, with extra overhead for maps lookup.

osbornjd:
> The changes to PHCASeeding will likely conflict with #1341, depending on what gets pushed first ...

Yes, if yours is merged first then I'll fix conflicts. I think this will fail as I needed to remove an include from the G4_Tracking macro that isn't used anyway.

>Also: isn't this commit essentially equivalent to going back to storing all global position in the clusters, at least for the TPC clusters which dominate anyway ?
What is the gain, then to have only the local ACTS coordinates in the clusters ?
In more detail: I expect modules like the CA Seeder to loop over all TPC clusters at least one, so that essentially all global coordinates get cached, and we are left to having (at least for all the time spent seeding) both the global and local coordinates stored in memory, with extra overhead for maps lookup.

To some degree, yes. To completely get away from this we will have to move the seeding to use local coordinates - perhaps in discussion with @petermic Michael this could be done since the seeding is performed in fixed radial layers anyway. But he is the expert so I think we need to discuss with him more.

The move to storing local coordinates only was also for compressed DST writing. Your comment is valid though - all of the seeding (both silicon and TPC) is performed in global coordinates. So we will always have a trade off between memory savings (storing local only) and speed (having to spend time transforming between local to global) as long as we seed in global coordinates.

Also even though this PR caches the global coordinates there is still an overall reduction in memory since the previous `TrkrCluster` also stored the global covariance, which is no longer needed.

hupereir:
> > The changes to PHCASeeding will likely conflict with #1341, depending on what gets pushed first ...
> 
> Yes, if yours is merged first then I'll fix conflicts. I think this will fail as I needed to remove an include from the G4_Tracking macro that isn't used anyway.
> 
> > Also: isn't this commit essentially equivalent to going back to storing all global position in the clusters, at least for the TPC clusters which dominate anyway ?
> > What is the gain, then to have only the local ACTS coordinates in the clusters ?
> > In more detail: I expect modules like the CA Seeder to loop over all TPC clusters at least one, so that essentially all global coordinates get cached, and we are left to having (at least for all the time spent seeding) both the global and local coordinates stored in memory, with extra overhead for maps lookup.
> 
> To some degree, yes. To completely get away from this we will have to move the seeding to use local coordinates - perhaps in discussion with @petermic Michael this could be done since the seeding is performed in fixed radial layers anyway. But he is the expert so I think we need to discuss with him more.
> 
> The move to storing local coordinates only was also for compressed DST writing. Your comment is valid though - all of the seeding (both silicon and TPC) is performed in global coordinates. So we will always have a trade off between memory savings (storing local only) and speed (having to spend time transforming between local to global) as long as we seed in global coordinates.

Concerning writing local DSTs I think we agreed that the written clusters could be reduced even further in size, and that you would in anyway need a conversion module to convert converted objects to usable objects anyway. 

Now for the useable object my impression is that moving to storing only local coordinates led to many non-anticipated difficulties, such as for dealing with TPC distortions, and timing degradation, that one now needs to correct for with the caching introduced here.

Additionally, not having looked at the code in too much detail, I wonder how this caching plays with the "on-demand" distortion corrections one have to add to the cluster global position as introduced in #1341. Should the non distortion-corrected position be cached ? The corrected one ? 

When working on adding the corrections to the CA seeder, I essentially hit the same issue that you have with the local-to-global transform: because of how the code is written, both are done _many_ times for every single clusters (I counted maybe 10 times ? More ?). Hence the caching. 
Even if the correction lookup is fast, having to redo it so often also has a significant impact (on top of the ActsTransform) on the seeder timing ... 
 
Maybe we could discuss this further during the Monday meeting before pushing this change ?

Hugo
> 
> Also even though this PR caches the global coordinates there is still an overall reduction in memory since the previous `TrkrCluster` also stored the global covariance, which is no longer needed.
Makes sence. So maybe a solution that stores the local and global positions in the cluster and only the local covariance would give us the best of both worlds without all the extra complication introduced by only storing the local coordinates ?

hupereir:
> > The changes to PHCASeeding will likely conflict with #1341, depending on what gets pushed first ...
> 
> Yes, if yours is merged first then I'll fix conflicts. I think this will fail as I needed to remove an include from the G4_Tracking macro that isn't used anyway.
> 
> > Also: isn't this commit essentially equivalent to going back to storing all global position in the clusters, at least for the TPC clusters which dominate anyway ?
> > What is the gain, then to have only the local ACTS coordinates in the clusters ?
> > In more detail: I expect modules like the CA Seeder to loop over all TPC clusters at least one, so that essentially all global coordinates get cached, and we are left to having (at least for all the time spent seeding) both the global and local coordinates stored in memory, with extra overhead for maps lookup.
> 
> To some degree, yes. To completely get away from this we will have to move the seeding to use local coordinates - perhaps in discussion with @petermic Michael this could be done since the seeding is performed in fixed radial layers anyway. But he is the expert so I think we need to discuss with him more.
That still would not work with the distortion corrections, which move the cluster away from the TPC surfaces ...

osbornjd:
Sure sounds like a good idea, I'll mark the PR as draft and we can discuss on Monday. I think you are raising the more general concern that we would like to do things in local coordinates but there is still a lot especially in the seeding that is dependent on global coordinates.

hupereir:
> Sure sounds like a good idea, I'll mark the PR as draft and we can discuss on Monday. I think you are raising the more general concern that we would like to do things in local coordinates but there is still a lot especially in the seeding that is dependent on global coordinates.

Yes ! That's a good (and concise) summary :)

PR #1344, opened 2021-11-19
Cluster silicon in local coordinates

osbornjd:

## Types of changes

This PR changes the silicon clustering to local coordinates instead of global coordinates. It removes the need for transforming between global/local, which is unnecessary and wastes time.

It also fixes  a bug in the silicon uncertainty calculations. Previously the silicon uncertainty was calculated in local coordinates. It was then rotated to global coordinate uncertainty via the stave/ladder azimuth. When stored in the `TrkrCluster` object, to get the local uncertainty with e.g. `getRPhiError()`, it was rotated from global to local based on the cluster azimuth and not the stave/ladder azimuth. This PR fixes this by just storing the local uncertainty as it is determined directly from the geometry, which corresponds to the pitch of the stave/ladder.

Therefore, it should be expected that the silicon cluster QA looks different; however, any large change in the overall tracking QA would be unexpected as the silicon uncertainties will change (affecting the tracking) but it should not be substantial.

## TODOs (if applicable)

### DIFF
--- a/offline/packages/intt/InttClusterizer.cc
+++ b/offline/packages/intt/InttClusterizer.cc
@@ -295,11 +295,9 @@ void InttClusterizer::ClusterLadderCells(PHCompositeNode* topNode)
     // we have a single hitset, get the info that identifies the sensor
     int layer = TrkrDefs::getLayer(hitsetitr->first);
     int ladder_z_index = InttDefs::getLadderZId(hitsetitr->first);
-    int ladder_phi_index = InttDefs::getLadderPhiId(hitsetitr->first);
-
+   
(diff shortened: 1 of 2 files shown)


osbornjd:
This PR needs to be discussed further with MVTX experts

osbornjd:
@hupereir please take a look, but I think the QA is good enough to merge now. The MVTX pulls are a little wider and the tracking resolution does not degrade anymore, and the low pT DCA seems to even improve slightly.

osbornjd:
@pinkenburg this is ready to merge now, and will require new references to be set once it is merged (I can set the references)

PR #1353, opened 2021-11-30
updated simulation injection of distortions to cylindrical format etc.

rccorliss:

## Types of changes

 This changes the expected format of the input distortion maps applied in simulated data to match the format supplied in reconstructed data.  The new maps contain only cylindrical coordinate elements.  This reduces their size and makes it easier to compare same-in, same-out which previously did not read the same format.  Additionally, a temporary fix regarding how to handle positions quite close to the CM has been resolved by breaking the distortion maps up into a z>0 and z<0 histogram, which no longer attempts to interpolate between the two separate halves.

## TODOs (if applicable)

This still needs testing with actual files.  The old G4_TPC.C macro default map filenames will need to be revised to use a working sample.

### DIFF
--- a/simulation/g4simulation/g4tpc/PHG4TpcDistortion.h
+++ b/simulation/g4simulation/g4tpc/PHG4TpcDistortion.h
@@ -109,24 +109,22 @@ class PHG4TpcDistortion
   //!@name static histograms
   //@{
   bool m_do_static_distortions = false;
-  bool m_static_map_onesided = false;
   std::string m_static_distortion_filename;
   std::unique_ptr<TFile> m_static_tfile;
-  TH3 *hDRint[2];
-  TH3 *hDPint[2];
-  TH3 *hDZint[2];
+  TH3 *hDRint[2]={nullptr,nullptr};
+  TH3 *hDPint[2]={nullptr,nullptr};
+  TH3 *hDZint[2]={nullptr,nullptr};
   //@}
 
   //!@name time ordered histograms
   //@{
   bool m_do_time_ordered_distortions = false;
-  bool m_time_ordered_map_onesided = false;
   std::string m_time_ordered_distortion_filename;
   std::unique_ptr<TFile> m_time_ordered_tfile;
-  TTree *TimeTree;
-  TH3 *TimehDR[2];
-  TH3 *TimehDP[2];
-  TH3 *TimehDZ[2];
+  TTree *TimeTree=nullptr;
+  TH3 *TimehDR[2]={nullptr,nullptr};
+  TH3 *TimehDP[2]={nullptr,nullptr};
+  TH3 *TimehDZ[2]={nullptr,nullptr};
   //@}
 };
 
--- a/simulation/g4simulation/g4tpc/PHG4TpcDistortion.h
+++ b/simulation/g4simulation/g4tpc/PHG4TpcDistortion.h
@@ -40,8 +40,8 @@ class PHG4TpcDistortion
   //! radial distortion for a given cylindrical truth location of the primary ionization
   double get_r_distortion(double r, double phi, double z) const;
 
-  //! phi distortion for a given cylindrical truth location of the primary ionization
-  double get_phi_distortion(double r, double phi, double z) const;
+  //! R*phi (hence the unitful phi-hat) distortion for a given cylindrical truth location of the primary ionization
+  double get_rphi_distortion(double r, double phi, double z) const;
 
   //! z distortion for a given cylindrical truth location of the primary ionization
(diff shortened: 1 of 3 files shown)


hupereir:
Hi Ross,
looking at the code change, it seems there is a logical change in the new code with respect to the old, namely that you cannot have both static ( == one for all event) and timeordered corrections at the same time: the code in PHG4TpcDistortion::get_distortion uses either one or the other, but not both, with precedence given to the static.
The previous logic would consider both static and timeordered correction, returning the sum of the two, if this is what is asked for. 
It is not clear to me what logic is the best: the first assumed that the time-ordered correction were small fluctuations on top of the static (hence the need for being able to set both flags to true). 
With the new logic you would have to have the time-ordered corrections be "full", namely also include the static. 
Comment ? 

Hugo

simulation/g4simulation/g4tpc/PHG4TpcDistortion.h
```diff
@@ -110,10 +123,10 @@ class PHG4TpcDistortion
   bool m_time_ordered_map_onesided = false;
   std::string m_time_ordered_distortion_filename;
   std::unique_ptr<TFile> m_time_ordered_tfile;
-  TTree *TimeTree = nullptr;
-  TH3 *TimehDX = nullptr;
-  TH3 *TimehDY = nullptr;
-  TH3 *TimehDZ = nullptr;
+  TTree *TimeTree;
```
hupereir, line 126: Should initialize TimeTree to nullptr (as in the previous code) otherwise it is dangling. 
Here it creates a crash the first time ::load_event() is called 
It should also be good (for safety) to also initialize all the histograms in the same way, e.g.: 
TH3 *TimehDR[2] = {nullptr, nullptr};

simulation/g4simulation/g4tpc/PHG4TpcElectronDrift.cc
```diff
@@ -372,24 +372,21 @@ int PHG4TpcElectronDrift::process_event(PHCompositeNode *topNode)
 
       if (m_distortionMap)
       {
-        const double x_distortion = m_distortionMap->get_x_distortion(x_start, y_start, z_start);
-        const double y_distortion = m_distortionMap->get_y_distortion(x_start, y_start, z_start);
-        const double z_distortion = m_distortionMap->get_z_distortion(x_start, y_start, z_start);
+	const double r_distortion = m_distortionMap->get_r_distortion(radstart, phistart, z_start);
+	const double phi_distortion = m_distortionMap->get_phi_distortion(radstart, phistart, z_start);
```
hupereir, line 376: should be: 
   const double phi_distortion = m_distortionMap->get_phi_distortion(radstart, phistart, z_start)/radstart;
because what is stored in the histogram are actually r.phi distortions.
Additionally you could rename  PHG4TpcDistortion::get_phi_distortion into "get_rphi_distortion" to make it explicit that one must then divide by "r".
Alternatively you could apply the division in PHG4TpcDistortion::get_phi_distortion directly, to make the output consistent with the function name.

rccorliss: Good catch.  f510baf resolves this per your "get_rphi_distortion" suggestion.

hupereir: > 

Thanks Ross !
There remains the matter of the histogram and TTree initialization in PHG4TpcElectronDrift.h, and removing the two unused members: m_static_map_onesided and m_time_ordered_map_onesided.

Hugo

rccorliss:
> Hi Ross, looking at the code change, it seems there is a logical change in the new code with respect to the old, namely that you cannot have both static ( == one for all event) and timeordered corrections at the same time: the code in PHG4TpcDistortion::get_distortion uses either one or the other, but not both, with precedence given to the static. The previous logic would consider both static and timeordered correction, returning the sum of the two, if this is what is asked for. It is not clear to me what logic is the best: the first assumed that the time-ordered correction were small fluctuations on top of the static (hence the need for being able to set both flags to true). With the new logic you would have to have the time-ordered corrections be "full", namely also include the static. Comment ?
> 
> Hugo

Hi Hugo,

The distortions we generate from geant ions directly will include the static terms, but you're right that if we wish to check that inputting the reco distortions produces the reco distortions, we may wish to factor the input into a static and event-by-event term (or even static, slow, and fluctuation?).  I will restore this functionality.

simulation/g4simulation/g4tpc/PHG4TpcDistortion.h
rccorliss: Resolved this in commit 39b72d4

PR #1367, opened 2021-12-15
Fix nan

hupereir:
It sometimes happens that CircleFitByTaubin fails, returning nan Radius and center position. This gets propagated in the CASeeder ultimately resulting in calling the magnetic field at (nan,nan,nan) positions as reported by Chris: 

PHField3DCartesian::GetFieldValue: Invalid coordinates: x: nan, y: nan, z: nan bailing out returning zero bfield
previous point: x: 62.5924, y: -36.0241, z: 17.0333

This PR checks success of the CircleFit to move forward with a given seed. Also clarifies the CircleFit code and added some const. This removes said error messages.

Failed circle fit can also happen in CASeeder::RemoveBadClusters. In the past this was resulting in all clusters failing the residual test. To remain consistent we now skip the seed entirely when this happends

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackreco/ALICEKF.cc
+++ b/offline/packages/trackreco/ALICEKF.cc
@@ -132,17 +132,17 @@ std::vector<SvtxTrack_v2> ALICEKF::ALICEKalmanFilter(const std::vector<keylist>&
     // Pre-set momentum-based parameters to improve numerical stability
     const auto& secondpos = globalPositions.at(trackKeyChain.at(1));
 
-    double second_x = secondpos(0);
-    double second_y = secondpos(1);
-    double second_z = secondpos(2);
-    double first_phi = atan2(y0,x0);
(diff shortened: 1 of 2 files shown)


osbornjd:
Hi Hugo,

Thanks for debugging this new crash. There appears to be a real degradation in the DCA from this PR, probably from (somehow) the track matching getting screwed up. The QA reference histograms need to be updated now that the new material maps have been put into calibrations, but looking through the last several PRs it appears that the DCA degradation is from this one. I also just run some pions with the default macro just to double check that the material map change didn't somehow screw up the DCA, and it doesn't appear to be so.

pinkenburg:
The clang warnings come from the inclusion of dd4hep - that needs some jenkins update to increase the error limit for clang. The valgrind error seems to be caused by root checking an invalid pointer (remember where we have to install the boost headers read only for local installs). That's outside our control but also there is nothing wrong with this (besides being completely weird by root to do that)

PR #1370, opened 2021-12-16
Rare Crash Fix

osbornjd:

## Types of changes

This PR fixes a few rare crashes that Chris found well exercising the new MDC hardware with the tracking.

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackreco/PHTpcTrackSeedCircleFit.cc
+++ b/offline/packages/trackreco/PHTpcTrackSeedCircleFit.cc
@@ -70,7 +70,7 @@ namespace
     line_fit(points, a, b);
   }
   
-  void CircleFitByTaubin (std::vector<std::pair<double,double>> points, double &R, double &X0, double &Y0)
+  void CircleFitByTaubin (std::vector<Acts::Vector3D> points, double &R, double &X0, double &Y0)
     /*  
     Circle fit to a given set of data points (in 2D)
     This is an algebraic fit, due to Taubin, based on the journal article
@@ -91,10 +91,10 @@ namespace
     double meanX = 0;
     double meanY = 0;
     double weight = 0;
-    for(unsigned int i = 0; i < points.size(); ++i)
+    for(const auto& position : points)
     {
-      meanX += points[i].first;
-      meanY += points[i].second;
+      meanX += position(0);
+      meanY += position(1);
       weight++;
     }
     meanX /= weight;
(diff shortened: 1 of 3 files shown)


offline/packages/trackreco/PHTpcTrackSeedCircleFit.cc
```diff
@@ -350,13 +348,16 @@ int PHTpcTrackSeedCircleFit::process_event(PHCompositeNode*)
...
       double R, X0, Y0;
-      CircleFitByTaubin(cpoints, R, X0, Y0);
+      CircleFitByTaubin(globalClusterPositions, R, X0, Y0);
       if(Verbosity() > 2) 
       { std::cout << "PHTpcTrackSeedCircleFit::process_event - track: " << track_key << " R=" << R << " X0=" << X0 << " Y0=" << Y0 << std::endl; }
 
       // set the track x and y positions to the circle PCA
       double dcax, dcay;
       findRoot(R, X0, Y0, dcax, dcay);
+      if(std::isnan(dcax) or std::isnan(dcay))
```
hupereir, line 358: Hi Joe, this reminds me of the other commit about nan in the Magnetic field lookup. 
Maybe the reason that ::findRoot fails is the R, x0 and y0 are already nan. (as in the other commit), due to the CircleFit actually failing. Can you double check ? 
In this case it should be enough to test on isnan(R) and stop there already. (spares you some extra calculations)

Hugo

osbornjd: I did check this, in the case of the crash that Chris found R X0 and Y0 were actually real valued. It turned out that in the case of this particular track X0 was very small (<1e-7) and Y0 was large, so basically R==Y0 and the subtraction returned a negative value which when sqrt-ed gave the nan. So the circle fit succeeded, it was just some very strange edge case where X0<<<Y0

hupereir: > I did check this, in the case of the crash that Chris found R X0 and Y0 were actually real valued. It turned out that in the case of this particular track X0 was very small (<1e-7) and Y0 was large, so basically R==Y0 and the subtraction returned a negative value which when sqrt-ed gave the nan. So the circle fit succeeded, it was just some very strange edge case where X0<<<Y0

Ok. Sorry for the noise !
Hugo

osbornjd: No worries, thanks for the double check!

pinkenburg:
we have a valgrind complaint in our recent PRs. It does look like the boost header issue we had in spring, cling checking an invalid pointer for no good reason. It seems related to the inclusion of dd4hep into our acts version. I'll have a look later and if it turns out to be bogus (chances are it will be) I'll suppress it

PR #1371, opened 2021-12-17
Fixing cluster mover surface finding

adfrawley:

## Types of changes

This PR fixes issues with getting the TPC subsurface key from global cluster coordinates after distortion corrections. The subsurface is found from coordinates in PHTpcClusterizer and in PHTpcClusterMover, so both are affected by this PR.
1) The cause of the occasional failure to find a TPC surface for distortion corrected clusters is that they sometimes change hitsets. In that case,  the surface vector returned using the hitset key is the wrong one. Discussed this with Joe, and we decided to use a single surface vector for each layer in the TPC (240 surfaces per layer), rather than subdividing the surfaces by hitset. This is the only way we could think of to keep the same cluster key for the corrected clusters. So now the TPC surface map uses the layer, an unsigned int, as the key instead of a hitset key. The user interfaces did not change, since the layer can be determined from the hitsetkey.  
2) Searching 240 surfaces in a layer every time we need to get the subsurface key from coordinates is pretty slow. I replaced that search with a direct calculation of the surface index, based on the world phi and z values. Seems to work reliably. It does rely on the surface vector being ordered in phi and z as it is now. The code screams loudly if it does not find a surface though. This change was made in PHTpcClusterMover and PHTpcClusterizer.
3) The surface Z step size set in MakeActsGeometry was slightly smaller than the surface center value, so clusters within 0.14 cm of zero were not assigned to a surface. That is now fixed.

## TODOs (if applicable)

### DIFF
--- a/offline/packages/tpc/TpcClusterizer.cc
+++ b/offline/packages/tpc/TpcClusterizer.cc
@@ -249,26 +249,56 @@ namespace
 	  
 	  double surfStepPhi = tGeometry->tpcSurfStepPhi;
 	  double surfStepZ = tGeometry->tpcSurfStepZ;
-	
+
 	  for(unsigned int i=0;i<surf_vec.size(); ++i)
 	    {
 	      Surface this_surf = surf_vec[i];
-	  
+	      
 	      auto vec3d = this_surf->center(tGeometry->geoContext);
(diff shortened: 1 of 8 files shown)


pinkenburg:
valgrind is the new problem which came in with dd4hep - not related to this PR

PR #1401, opened 2022-02-01
Tpot new geometry

hupereir:
This implements the latest TPOT geometry directly into geant 4. 
See https://indico.bnl.gov/event/14546/contributions/59761/attachments/39625/65748/talk.pdf for details
Depends on https://github.com/sPHENIX-Collaboration/macros/pull/488 and https://github.com/sPHENIX-Collaboration/coresoftware/pull/1400

## Types of changes
- [] Breaking change (fix or feature that would cause existing functionality to not work for users)

## TODOs (if applicable)

https://github.com/sPHENIX-Collaboration/macros/pull/488

### DIFF
--- a/offline/packages/micromegas/CylinderGeomMicromegas.cc
+++ b/offline/packages/micromegas/CylinderGeomMicromegas.cc
@@ -152,41 +152,6 @@ int CylinderGeomMicromegas::find_tile_planar( const TVector3& world_coordinates 
 }
 
 //________________________________________________________________________________
-void CylinderGeomMicromegas::convert_to_planar( uint tileid, PHG4Hit* hit ) const
-{
-  
-  // get extrapolation direction based on in and out G4Hit positions
-  const TVector3 world_direction( hit->get_x(1)-hit->get_x(0), hit->get_y(1)-hit->get_y(0), hit->get_z(1)-hit->get_z(0) );
-
-  // same treatment is applied to both coordinates of the G4Hit 
-  for( int i = 0; i < 2; ++i )
-  {
-
-    const TVector3 world_coordinates( hit->get_x(i), hit->get_y(i), hit->get_z(i) );
-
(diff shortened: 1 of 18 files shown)


osbornjd:
@hupereir there are a lot of nontrivial changes to the low multiplicity tracking QA from this PR, e.g. the momentum resolution degrades by almost a factor of 2. Is it understood what this is from?

hupereir:
@osbornjd 
The changes in the Micromegas QA (small modifications of the residuals distributions) are understood: new position of the TPOT, new pitch. And are not dramatic

The dramatic changes in the tracking QA and the Vertex QA, I suspect they come from inconsistent material map (again), because volumes definitions and hierarchy have changed with the new geometry (no more fake surfaces, no more cylinder micromomegas), but I cannot test/verify that this is indeed the case

Hugo

osbornjd:
I could see that being the case for the momentum resolution but the DCA distribution changes are surprising. I will try to take a look at the new volume assignments later today and see if there is an easy way to check if this might be the cause.

hupereir:
The reduction of the number of micromegas clusters on the "reco cluster per layer per track" plots is due to the fact that there is now only 8 TPOT modules instead of 26 before. To me it says that the track finding works essentially the same as before (which I have verified independently). Now, more than 60% of the TPC acceptance has no TPOT at all: it should have no impact on the momentum resolution, whatsoever

osbornjd:
Sure, I expected that the TPOT QA would change but was surprised to see such drastic changes in the tracking/vertexing. But let me take a look this afternoon and I'll report back.

osbornjd:
It does look like the new volume elements have different mappings, so the material map is broken. I will try to update it tomorrow with the new geometry. This should definitely fix the momentum resolution and hopefully the DCA as well as shown in the QA for this PR.

BTW @blackcathj let's not update the reference so that we have a clear indication of what the performance was pending an upcoming PR to update the material map.

PR #1404, opened 2022-02-03
Truth Seeding Bug Fix

osbornjd:
Update helical track fit option to use new global cluster position API

This PR makes two changes:

1. The default truth track seeding parameters somehow were changed to not be smeared out at all and taken as the truth values. This is not representative of an actual seeding algorithm, so this was changed back to a 5% smearing.
2. The optional helical fit code had become deprecated with the new cluster global position API. This is now fixed so that there is the option to perform a helical fit on the truth seed and get a more realistic set of initial track parameters.

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackreco/PHTruthTrackSeeding.cc
+++ b/offline/packages/trackreco/PHTruthTrackSeeding.cc
@@ -9,9 +9,11 @@
 #include <trackbase/TrkrCluster.h>
 #include <trackbase/TrkrClusterContainer.h>
 
+
 #include <trackbase/TrkrHitTruthAssoc.h>
 #include <trackbase_historic/SvtxVertexMap.h>
 #include <trackbase_historic/SvtxVertex.h>
+#include <trackbase_historic/ActsTransformations.h>
 
 #include <g4eval/SvtxClusterEval.h>
 #include <g4eval/SvtxEvalStack.h>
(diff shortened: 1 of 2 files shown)


osbornjd:
PS - this PR can also check that the tracking QA has returned to normal with the new material maps, in which case we can update the reference

osbornjd:
QA also appears back to normal after the new material map merge, resetting references

hupereir:
@osbornjd 
It seems this is the first jenkins report with your newest material map, and new TPOT geometry. 
I looked at the QA plots for low multiplicity, and they  look "reasonable", but please double check. 
I do see some deterioration, in particular for the momentum resolution vs pT. I would attribute this to the fact that TPOT acceptance has been significantly reduced with respect to the reference. 
With the old geometry (26 Modules) I did see some improvement when turning ON TPOT with respect to not having it. This improvement is now largely gone, with the descoped TPOT, which covers about 1/3 of the previous TPOT acceptance. 
One way to double check that, would be once more to compare momentum resolution w/ w/o TPOT, with latest geometry and material map. 

Hugo

hupereir:
PS: thanks for the updated material map !

osbornjd:
Hi Hugo, that makes sense as the largest pT resolution degradation is at high pT where I imagine having a O(100) micron space point at such large radius really helps identify the bend of the track better. In any case the low pT resolution looks pretty comparable to before, so we can test out with/without TPOT as you suggested.

PR #1440, opened 2022-03-14
MVTX and KFParticle update

cdean-github:

Changed three projects (because I didn't realise I had staged the MVTX code for a push...)

## Types of changes

Decay Finder: Removed useless code (stuff that was commented out) and added an improved object naming system

KFParticle: We couldn't read back the truth history as we initially only read from the HepMC record. For MDC2 we limited the pythia8 decay volume to 1nm so there was no truth association available in the reco. I added Geant4 truth tracing which should transition to the HepMC record when we get to the decay volume limit.

MVTX: Copied support structure from macros to full class. 

## TODOs (if applicable)

See if I can update DecayFinder to also use the Geant4 record when available. This package works backward to KFParticle so I need to think about it a little.

Remove as many assembly volumes from MVTX as I can. This is why I staged the changes in the first place but it's more important to get the KFParticle update out.

[Corresponding macros PR](https://github.com/sPHENIX-Collaboration/macros/pull/497)


pinkenburg:
The PHG4MvtxSupport defines its own version of elements:
G4Element* elH  = new G4Element("Hydrogen",symbol="H" , 1., 1.01*g/mole);
  G4Element* elC  = new G4Element("Carbon"  ,symbol="C" , 6., 12.01*g/mole);
  G4Element* elN  = new G4Element("Nitrogen",symbol="N" , 7., 14.01*g/mole);
  G4Element* elO  = new G4Element("Oxygen"  ,symbol="O" , 8., 16.00*g/mole);
Is that really necessary?

cdean-github:
> The PHG4MvtxSupport defines its own version of elements: G4Element* elH = new G4Element("Hydrogen",symbol="H" , 1., 1.01_g/mole); G4Element_ elC = new G4Element("Carbon" ,symbol="C" , 6., 12.01_g/mole); G4Element_ elN = new G4Element("Nitrogen",symbol="N" , 7., 14.01_g/mole); G4Element_ elO = new G4Element("Oxygen" ,symbol="O" , 8., 16.00*g/mole); Is that really necessary?

Ah, you've just highlighted something I was supposed to fix here. The carbon fibre composite object is also called several times instead of once

pinkenburg:
There is a duplication of visattributes created in the code and then adding  the volume to the display action. When just creating visattributes by default they just eat up memory (and leak at the end). The displayaction makes sure the memory only gets allocated when the display is started and it cleans the attributes up when quitting.

cdean-github:
> valgrind shows memory leaks in the mvtx support. That class has an empty dtor but the assembly volumes need to be deleted (check with the outer hcal detector which also uses assembly volumes). Rotation matrices just seem to create a memory leak - so far we haven't found a way to get around that. Those can be ignored

I'll take a look at the outer HCal code. I also saw no way around the rot matrix beyond giving it a rotation of 0 around z. Just calling delete caused a seg fault

pinkenburg:
> > valgrind shows memory leaks in the mvtx support. That class has an empty dtor but the assembly volumes need to be deleted (check with the outer hcal detector which also uses assembly volumes). Rotation matrices just seem to create a memory leak - so far we haven't found a way to get around that. Those can be ignored
> 
> I'll take a look at the outer HCal code. I also saw no way around the rot matrix beyond giving it a rotation of 0 around z. Just calling delete caused a seg fault

yes - there seem to be no way around the rotation matrix (as you say - it segfaults if you try to delete it). Don't worry about them - those are taken care off in the suppression file (which needs to be updated for every single occurrence). I'll take care of that at some point, the new root version also introduced some new warnings

PR #1476, opened 2022-04-14
Cluster container

hupereir:
[comment]: <> (Please tell us something about this pull request)
This PR simplifies the TrkrClusterContainer public interface in preparation for the new clusterContainer to minimize the number of function calls that make explicit assumptions to the underlying structure.
In particular: 
- ::findOrAddCluster has been removed 
- ::getClusterMap has been removed
- ::addCluster now returns void rather than an interator to the underlying map.
In addition ::getClusterRange( hitsetid ) has been made non const. 
The code calling these functions has been adapted accordingly, in principle with no loss of functionality nor performance
TpcSimpleClusterizer has been updated to incorporate recent changes to TpcClusterizer.

## Types of changes
[comment]: <> ( What types of changes does your code introduce? Put an `x` in all the boxes that apply: )
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work for users)
- [ ] Requiring change in macros repository (Please provide links to the macros pull request in the last section)
- [x] I am a member of [GitHub organization of sPHENIX Collaboration](https://github.com/orgs/sPHENIX-Collaboration/people), EIC, or ECCE (contact Chris Pinkenburg to join)

## What kind of change does this PR introduce? (Bug fix, feature, ...)

[comment]: <> ( What does this PR do? Linking to talk in software meeting encouraged )

## TODOs (if applicable)

[comment]: <> ( In case this is a draft PR, e.g. for running checks using Jenkins, please make the pull request as a draft: https://github.blog/2019-02-14-introducing-draft-pull-requests/  )

PR #1459, opened 2022-04-05
Add hit time

adfrawley:
[comment]: <> (Please tell us something about this pull request)

## Types of changes
[comment]: <> ( What types of changes does your code introduce? Put an `x` in all the boxes that apply: )
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work for users)
- [ x] Requiring change in macros repository (Please provide links to the macros pull request in the last section)
- [x] I am a member of [GitHub organization of sPHENIX Collaboration](https://github.com/orgs/sPHENIX-Collaboration/people), EIC, or ECCE (contact Chris Pinkenburg to join)

## What kind of change does this PR introduce? (Bug fix, feature, ...)

[comment]: <> ( What does this PR do? Linking to talk in software meeting encouraged )
This PR is a first step to enable extended readout reconstruction in pp collisions:
- Adds trackbase/TrkrHitv3 with a bunch crossing field.
- Adds trackbase_historic/SvtxTrack_v3 with bunch crossing field.
- Adds trackbase/TrkrClusterCrossingAssoc to  node tre to contain an (INTT cluster, crossing) multimap.
- Adds a class tpc/TpcClustetZCrossingCorrection to correct TPC cluster Z given bunch crossing and readout side.
- Adds code to intt/InttClusterizer to make cluster-crossing associations.
- Adds code to PHSiliconTpcTrackMatching to process tracks from any arbitrary bunch crossing.
- Removes all implicit assumptions that the TPC readout side can be determined from the sign of cluster Z.
- Removes some obsolete code used for PHTpcTracker TPC seeding.
- Requires a minor change to the tracking macro.

Things still to be done to enable full extended readout reconstruction:
- The CA seeder has to be modified to be vertex agnostic (currently uses eta cut in searches).
- The workflow has to be modified to accommodate TPC distortions (without knowing the true z cluster positions at the start).

## TODOs (if applicable)

[comment]: <> ( In case this is a draft PR, e.g. for running checks using Jenkins, please make the pull request as a draft: https://github.blog/2019-02-14-introducing-draft-pull-requests/  )

PR #1481, opened 2022-04-20
align datamembers to 64bit

bogui56:

## Types of changes

## TODOs (if applicable)


hupereir:
Hi Christof
Looking at reasons why the compilation of this PR fails, it seems that many changes introduced in https://github.com/sPHENIX-Collaboration/coresoftware/pull/1476 have been reverted. This is possibly the result of a wrong conflict solving when merging with master. In any case this is what breaks compilation. 
I would try to merge the new commits more carefully on top of the current master ...

sPHENIX-GitHub-l:
Hi Hugo, 

thanks for your remarks. Obviously all my fault. 
The compilation fail was due to the fact that I accidentally submittted the wrong makefile (i.e. the one from tracckbase). I have to be more carefull indeed.

Regading the TpcClusterizer I ran a detailled diff again and other than a few places where you replced explicit declarations with auto there are only two real differences
that between the two versions:

518c487
<       if(my_data.clusterlist)     
---
>       if(my_data.cluster_vector)  

this is a good catch on your part. This was an artifact from switching back and forth from maps to vectors to get the memory leaks fixed.

And this here:
887c839,840
<     unsigned short NZBinsSide = NZBins/2;
---
>     unsigned short NZBinsSide = NZBins;

I am not sure I understand this. NZBins/2 gives the right number of bins on each side of the TPC.
Why do we have to extend this?
We should check this, this may explain some more of the memory increase we see.
 
Cheers

   Christof

hupereir:
> And this here: 887c839,840 < unsigned short NZBinsSide = NZBins/2; ---
> unsigned short NZBinsSide = NZBins;
> I am not sure I understand this. NZBins/2 gives the right number of bins on each side of the TPC. Why do we have to extend this? We should check this, this may explain some more of the memory increase we see. Cheers Christof

Hi Christof
I agree. This change is strange. It does not come from https://github.com/sPHENIX-Collaboration/coresoftware/pull/1476 though. Some must have been there for a longer time. Still it rings a bell. I will try to trace back where it comes from

Hugo

hupereir:
> > And this here: 887c839,840 < unsigned short NZBinsSide = NZBins/2; ---
> > unsigned short NZBinsSide = NZBins;
> > I am not sure I understand this. NZBins/2 gives the right number of bins on each side of the TPC. Why do we have to extend this? We should check this, this may explain some more of the memory increase we see. Cheers Christof
> 
> Hi Christof I agree. This change is strange. It does not come from #1476 though. Some must have been there for a longer time. Still it rings a bell. I will try to trace back where it comes from
>

So double checking: the code change about NZBins date from @adfrawley 's PR https://github.com/sPHENIX-Collaboration/coresoftware/pull/1459/
So probably he should comment on that before merging this pull request, or at least one should re-add the comments at line 844 of TPCClusterizer to keep Tony's original code unchanged.

pinkenburg:
There are changes compared to the reference. The most noticeable is the dca which seem to have grown wider at negative dca's

pinkenburg:
This PR has a merging conflict which needs to be resolved before we can do anything about it

sPHENIX-GitHub-l:
There seems to be a problem with the macros?? No clue how to read that.
Could someone help me parse the error messages here?

offline/packages/trackbase/TrkrClusterv4.h
```diff
@@ -115,18 +115,14 @@ class TrkrClusterv4 : public TrkrCluster
...
-  unsigned short int m_adc;           //< cluster sum adc
-  
-  float m_local[2];          //< 2D local position [cm]
-  bool m_valid;
-  char m_size;
-  char m_phisize;
-  char m_zsize;
-  char m_overlap;
-  char m_edge;
+  TrkrDefs::cluskey m_cluskey;  //< unique identifier within container 64bit - cumul 1*64
```
hupereir, line 118: m_cluskey can be safely removed (in this PR or another): it cannot be set or accessed by the outside world anymore

offline/packages/tpc/TpcClusterizer.cc
```diff
@@ -611,7 +646,16 @@ namespace
...
+      ihit_list.clear();
     }
+
+    //std::multimap<TrkrDefs::cluskey, TrkrDefs::hitkey> empty_map;
+    // empty_map.swap(*my_data.clusterhitassoc);
+    // my_data->clusterhitassoc->clear();//HERE
+
+    adcval.clear();
+    all_hit_map.clear();
+    hit_vect.clear();
```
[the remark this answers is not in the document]
bogui56: Agreed, we can remove this in the next iteration

offline/packages/trackbase/TrkrClusterv4.h
bogui56: Yes, the ClusterKey I will remove later. i wanted to get the Clusterizer with your changes out first. It compiles now and all your changes should be in now.

pinkenburg:
looks like a hardcoded path somewhere. The macros die with:
Error in <TFile::TFile>: file /phenix/u/bogui/data/Field/sphenix3dtrackingmapxyz.root does not exist
our filesystems are not mounted on the node where jenkins runs

pinkenburg:
This PR introduces massive changes for the p values for the QA. Even for the calorimeters. We need to understand what's going on here

bogui56:
I had a look at the tpc and tracking plots. 
With the pull request we get a few more hits on track and per layer per track. 
There is nothing catastrophic, as far as I can see.

Regarding the calorimeter plots, there are a few plots that are simply not filled.
I.e. 3x3 and 5x5 sum / truth are not filled. Not sure we can blame this on the 
tracking field and the tpcclusterizer.

The current setup with the small fields should be used for the tracking step 
alone, the calo reco shouldn't run in this step. 

It would be good to merge this soon, or at least put together a playbuild, 
so we can benchmark the memory and cpu perforrmance of the new setup.

Cheers

   Christof

pinkenburg:
right - the truth info was broken for this PR. Forget about the calorimeters. Merging this now

adfrawley:
Dear Hugo and Christof,

This change is needed to accommodate hits from non-zero crossings. Hits from later crossings can be read out on the north side but have negative z values because they appear to have drifted much longer. If we want to handle nonzero crossings, we have to deal with this - we can no longer assume that the sign of z determines which side of the TPC the hit was read out on. These hits will appear that way in the data stream.

Tony

PR #1489, opened 2022-04-26
Allow Multiple tpc corrections

rccorliss:
This pull request allows three separate corrections:  Static, Average, and Fluctuation to coexist on the node tree.  The default behavior is to load a single correction into the Static node, for now.
## Types of changes

We eventually need to handle three different TPC distortion corrections:
Static - Large correction.  changes rarely
Average - Small correction. changes regularly
Fluctuation - Tiny correction. changes rapidly
This allows each of these to be held in the node tree in separate items, to be updated as desired.  It also allows 2D and 3D corrections to be handled interchangeably.
## TODOs (if applicable)
I was unable to get an ACTS dependency to resolve correctly in my private repo.  This may need some adjustment to compile correctly.


hupereir:
See coments above.
Also I would use an enumeration (rather than a simple index) to describe which distortion is what. e.g:
enum DistoritonType
{
   StaticOnly = 0,
   BeamInducedAveraged = 1,
   BeamInducedFluctuation = 2 
};
something along that line

You could also add a method "get_node_name( index ); this way calling modules like SimpleKFProp does not have to hardcode the node name

pinkenburg:
This PR has a merging conflict. It likely needs to be merged with the current master, the conflict resolved and then pushed again

offline/packages/tpc/TpcDistortionCorrection.cc
```diff
@@ -27,13 +27,34 @@ Acts::Vector3 TpcDistortionCorrection::get_corrected_position( const Acts::Vecto
...
   
-  // update cluster
+    // update cluster
   const auto x_new = r_new*std::cos( phi_new );
   const auto y_new = r_new*std::sin( phi_new );
 
   return {x_new, y_new, z_new};
 }
+
+TVector3 TpcDistortionCorrection::get_corrected_position( const TVector3& source, const TpcDistortionCorrectionContainer* dcc, unsigned int mask) const
```
hupereir, line 56: I would refrain to add convenience functions unless they are actually used somewhere. 
This is not the case here. 
I would recommand to remove this, and only add it when (and if) some module actually uses this.
This just cluters the code for no reason ... 
It also opens the doors to unwanted conversions in the code from float to double and vice versa.

offline/packages/tpc/TpcDistortionCorrection.h
```diff
@@ -28,9 +35,12 @@ class TpcDistortionCorrection
   };
   
   //! get cluster corrected 3D position using given DistortionCorrectionObject
-  Acts::Vector3 get_corrected_position( const Acts::Vector3&, const TpcDistortionCorrectionContainer*, 
+ Acts::Vector3 get_corrected_position( const Acts::Vector3&, const TpcDistortionCorrectionContainer*, 
 					unsigned int mask = COORD_ALL ) const;
-  
+
+ //! get cluster corrected 3D position using given DistortionCorrectionObject
+ TVector3 get_corrected_position( const TVector3&, const TpcDistortionCorrectionContainer*, unsigned int mask = COORD_ALL ) const;
```
hupereir, line 42: Same remark: unless actually used, I would remove

hupereir:
Hi Ross, 
Thanks for the update, 

Two more comments.
Please refrain adding "convenience" functions just for the sake of it, this clutters the code and make it harder to maintain. 

Hugo

hupereir:
> Hi Hugo,
> 
> Is Acts::Vector the desired interchange format for space points? This is an unused convenience function for the duration of this pull, but I was intending to use it in dealing with the central membrane immediately following the merge, since I am more used to working with TVector3. If the guidance is to avoid TVector3, I can do so, but it's been the object of choice for me for lots of 3D manipulations.
> 
It's what's mostly used yes. 
There is no guidance to avoid TVector3. I would think the guidance is: "do not write code that is not used elsewhere" (because it will break and nobody will notice)
Again: when the same interface will actually be needed with TVector3 instead of ACTS::Vector3, it should be easy enough to add it at that point. (or to avoid having to add it)

hupereir:
See comment above about loading the "fluctuation" distortion in the CorrectionLoader

pinkenburg:
The failure was caused by a bad macro which snuck in. Restarting jenkins

PR #1495, opened 2022-04-28
Updated DecayFinder to look at Geant record and handle pile-up

cdean-github:

## Types of changes

DecayFinder was unable to handle any decays that were performed by Geant. These decays were never in the HepMC record. As a consequence, the module would crash when accessing the end vertex of some particle (because they were null pointers).

This PR updates the module so it will now jump back and forth between HepMC and Geant4 as required. This is transparent to users.

New features:

1. Seamless transition between HepMC and Geant4 records
2. New output info to see if your mother produced pi0s and photons on decay, these particles would cause incorrect mass reconstruction and could throw decays out our windows
3. We now trigger on decays that would be reconstructable. Previously we triggered on events that had the decay, regardless of whether all products were in the acceptance or not
4. We also loop over the pile-up events to see if the decay was in there too. 

## TODOs (if applicable)

### DIFF
--- a/generators/decayFinder/DecayFinder.cc
+++ b/generators/decayFinder/DecayFinder.cc
@@ -13,6 +13,7 @@
 #include "DecayFinderContainerBase.h"    // for DecayFinderContainerBase::Iter
 #include "DecayFinderContainer_v1.h"     // for DecayFinderContainer_v1
 
+#include <g4main/PHG4TruthInfoContainer.h>
 #include <phhepmc/PHHepMCGenEvent.h>
 #include <phhepmc/PHHepMCGenEventMap.h>
 
@@ -87,13 +88,12 @@ int DecayFinder::process_event(PHCompositeNode* topNode)
 
   if (decayFound)
   {
-     m_counter += 1;
-    if (Verbosity() >= VERBOSITY_MORE) printNode(topNode);
+    if (m_save_dst && Verbosity() >= VERBOSITY_MORE) printNode(topNode);
   }
 
   if (m_triggerOnDecay && !decayFound)
   {
-    if (Verbosity() >= VERBOSITY_MORE) std::cout << "The decay, " << m_decayDescriptor << " was not found in this event, skipping" << std::endl;
+    if (Verbosity() >= VERBOSITY_MORE) std::cout << "The decay, " << m_decayDescriptor << " was not found or reconstructable in this event, skipping" << std::endl;
     return Fun4AllReturnCodes::ABORTEVENT;
   }
   else
@@ -109,6 +109,13 @@ int DecayFinder::End(PHCompositeNode* /*topNode*/)
   return 0;
 }
 
+/*
+ *
+ *This function tries to intepret the string you wrote
+ *Will be semi-independent of the rest of this class
+ *(semi- because we need this to set the charges and strings)
+ *
+ */
 int DecayFinder::parseDecayDescriptor()
 {
   bool ddCanBeParsed = true;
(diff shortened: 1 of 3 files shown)


cdean-github:
I think the failure is due to the compilation order. DecayFinder now needs ```PHG4TruthInfoContainer``` which is in ```-lphg4hit``` and is compiled after. We could try to rearrange the compilation order or move DecayFinder out of ```generators``` and into ```packages```

```
/cvmfs/sphenix.sdcc.bnl.gov/gcc-8.3/opt/sphenix/core/binutils/2.30-e5b21/x86_64-centos7/bin/ld: cannot find -lphg4hit
collect2: error: ld returned 1 exit status
make[1]: *** [libdecayfinder.la] Error 1
make[1]: Leaving directory `/var/lib/jenkins/workspace/sPHENIX/Build-Master-gcc8/build/new/build/coresoftware/generators/decayFinder'
make: *** [all] Error 2
system  make   -l 32 -j 32 >> /var/lib/jenkins/workspace/sPHENIX/Build-Master-gcc8/build/new/rebuild.log 2>&1 failed: 512
```

@blackcathj @pinkenburg do you have a preference for this?

pinkenburg:
The build order can be reshuffled. This is more of a problem what a generator constitutes. In principle it produces input for the sims or in the original case massages the input.
This change requires the truth information which is only generated during the simulation. That is more evaluation. Besides you have to decide when to run this module (G4 is just a module). If you want to decay the input of the sims before G4 runs, you do not have the truth info. If you need the truth info you cannot modify the G4 input anymore. My feeling is that this should be split into the generator modification part and the eval part

cdean-github:
> The build order can be reshuffled. This is more of a problem what a generator constitutes. In principle it produces input for the sims or in the original case massages the input. This change requires the truth information which is only generated during the simulation. That is more evaluation. Besides you have to decide when to run this module (G4 is just a module). If you want to decay the input of the sims before G4 runs, you do not have the truth info. If you need the truth info you cannot modify the G4 input anymore. My feeling is that this should be split into the generator modification part and the eval part

OK, I see the complication. As a base module, it just checks whats in the HepMC record and Geant record but, because it can trigger on what's in there, it could influence the generator. This trigger only calls an abort event though and doesn't force a re-run of the generator stage. 

Separating the two could be tricky, or maybe impossible, as it now tries to jump between HepMC and Geant4 if the HepMC record doesn't have all the information.

cdean-github:
> > > The build order can be reshuffled. This is more of a problem what a generator constitutes. In principle it produces input for the sims or in the original case massages the input. This change requires the truth information which is only generated during the simulation. That is more evaluation. Besides you have to decide when to run this module (G4 is just a module). If you want to decay the input of the sims before G4 runs, you do not have the truth info. If you need the truth info you cannot modify the G4 input anymore. My feeling is that this should be split into the generator modification part and the eval part
> > 
> > 
> > OK, I see the complication. As a base module, it just checks whats in the HepMC record and Geant record but, because it can trigger on what's in there, it could influence the generator. This trigger only calls an abort event though and doesn't force a re-run of the generator stage.
> > Separating the two could be tricky, or maybe impossible, as it now tries to jump between HepMC and Geant4 if the HepMC record doesn't have all the information.
> 
> You still have the problem that you can only run this after G4 so this is a filter based on the Truth Info - not a generator or generator filter. Applying a filter after G4 is done doesn't save us cpu, depending on how much is thrown out this will turn into a major waste of resources (since this is run as part of the production). Or is the plan to run this twice (which means registering two independent modules)? But then you would also exercise different parts of the module before and after G4.

What the module does is it first looks at the generator level info and decides if it has all the information it needs to make a decision, if not then it goes to the G4 info. So if it's run just after the generator, and we don't limit the generator, then it wouldn't look at the G4 info. I think I may load up the G4 nodes when we process the first event regardless so I could switch it to only load if needed, this won't fix the issue of needing the library at compilation though.

This is the new change I had to introduce because I limited the generators decay volume for MDC2. 

I wonder if it's worth just making a new package, something like ```geant4Finder``` for users and just leaving the module already in coresoftware as is. This could be quite confusing for users though.

pinkenburg:
That failure was caused by the G4_Mvtx.C macro. Restarting jenkins

pinkenburg:
looks like the restart did not pick up the modified macros. merging this anyway

PR #1558, opened 2022-07-06
Scan embed clusters bug

adfrawley:

## Types of changes
- [x ] Bug fix (non-breaking change which fixes an issue)

This PR does three things:
1) It fixes the evaluator bug that caused the cluster ntuple to not be filled when the scan-for-embedded flag was set.
2) Fixes a bug in the calculation of the flight time correction in PHTpcDeltaZCorrection.
3) Changes the SAMPA shaping offset correction in TpcClusterizer to be a time in ns, and tunes it so that z values match in the TPC north and south at z = 0.

## TODOs (if applicable)

### DIFF
--- a/offline/packages/tpc/TpcClusterizer.cc
+++ b/offline/packages/tpc/TpcClusterizer.cc
@@ -76,9 +76,8 @@ namespace
     unsigned short maxHalfSizeT = 0;
     unsigned short maxHalfSizePhi = 0;
     double m_tdriftmax = 0;
-    double par0_neg = 0;
-    double par0_pos = 0;
-    int cluster_version = 3;
+    double sampa_tbias = 0;
+     int cluster_version = 3;
     std::vector<assoc> association_vector;
     std::vector<TrkrCluster*> cluster_vector;
     int verbosity = 0;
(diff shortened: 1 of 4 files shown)


osbornjd:
@adfrawley I have been looking through the QA to track what things have changed over the last several days. A couple of notes:

1. The TPC z cluster uncertainty QA is screwed up because of the change to clustering in time instead of z. My guess is that the QA module didn't get updated to reflect that change.
2. It seems that this PR did not fix the high pT DCA_z resolution that showed up in the initial change to clustering in time. Is there another reason why that might be (but the TPC cluster residuals do appear to look normal after this PR)

sPHENIX-GitHub-l:
Hi Joe,

I missed this:

The calculation of z position from TPC readout time is handled properly in ActsGeometry::getGlobalPosition(cluskey, cluster).

However the (many) modules that use the TPC cluster error all get it directly from the cluster using cluster->getZError(). After the recent changes, the cluster z error is the square of the error in the arrival time, not the square of the error in z. It has to be multiplied by the square of the drift velocity to convert it to cm^2.

Possible solutions are:

  1.  Convert the error to cm in TpcClusterizer before writing it to the cluster.
  2.  Multiply the error by the drift_velocity^2 everywhere it is used.
  3.  Add a method to ActsGeometry like getZError(cluster) that will (for the TPC) convert the z error to cm^2 before returning it.

Option 1 is convenient but a bit ugly, since the cluster will store the local position as time but the error as cm^2, which may be confusing.  Option 2 is both inconvenient and ugly. Option 3 is inconvenient but not confusing.

Laziness lures me toward option 1. Any objections to doing that?

Thanks
Tony

osbornjd:
@adfrawley option (1) works for me for now - as long as we keep straight that the cluster time is stored in the local coordinate position but the actual cluster z uncertainty is stored in the covariance. This may also change or be expressed differently once we move to the parameterized uncertainties.

Do you have any ideas why the high pT DCA_z didn't return to it's previous performance?

PR #1564, opened 2022-07-07
Track Timing Fix

osbornjd:

## Types of changes

This PR fixes a few bugs I introduced which caused the track timing to explode. Thanks to Jin and Jenkins for finding the PR it was associated to.

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackbase/ActsGeometry.h
+++ b/offline/packages/trackbase/ActsGeometry.h
@@ -12,17 +12,6 @@ class ActsGeometry {
   ActsGeometry() = default;
   ~ActsGeometry() {} 
 
-  void setGeometry(const ActsTrackingGeometry tGeometry) 
-    { m_tGeometry = tGeometry; }
-
-  void setSurfMaps(const ActsSurfaceMaps surfMaps)
-    { m_surfMaps = surfMaps; }
-  
-  ActsTrackingGeometry geometry() const 
-    { return m_tGeometry; }
-  ActsSurfaceMaps maps() const 
-    { return m_surfMaps; }
-
   void set_drift_velocity(double vd) {_drift_velocity = vd;}
   double get_drift_velocity() {return _drift_velocity;}
 
@@ -49,7 +38,6 @@ class ActsGeometry {
       Acts::Vector3 world,
       TrkrDefs::subsurfkey& subsurfkey) const;
 
- private:
   ActsTrackingGeometry m_tGeometry;
   ActsSurfaceMaps m_surfMaps;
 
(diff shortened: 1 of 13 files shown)


hupereir:
@osbornjd The diagnostic is correct, (essentially your accessors were making a full copy each time), but the fix, that consists in removing the accessors and making the members public is not. 
You should use const references instead. This would prevent creating copies, without having to change the public API nor expose the members.  See inline examples below

offline/packages/trackbase/ActsGeometry.h
```diff
@@ -12,17 +12,6 @@ class ActsGeometry {
   ActsGeometry() = default;
   ~ActsGeometry() {} 
 
-  void setGeometry(const ActsTrackingGeometry tGeometry) 
-    { m_tGeometry = tGeometry; }
-
```
hupereir, line 17: I would keep this but change to: 
void setGeometry( const ActsTrackingGeometry& tGeometry) 
The "&" prevents unnecessary copy

offline/packages/trackbase/ActsGeometry.h
```diff
@@ -12,17 +12,6 @@ class ActsGeometry {
...
-  void setSurfMaps(const ActsSurfaceMaps surfMaps)
-    { m_surfMaps = surfMaps; }
-  
-  ActsTrackingGeometry geometry() const 
-    { return m_tGeometry; }
```
hupereir, line 22: You can keep the accessor but change to 
const ActsTrackingGeometry& geometry() const
This would prevent unnecessary copy of the member

offline/packages/trackbase/ActsGeometry.h
```diff
@@ -12,17 +12,6 @@ class ActsGeometry {
...
-  ActsSurfaceMaps maps() const 
-    { return m_surfMaps; }
```
hupereir, line 24: here also: const ActsSurfaceMaps& maps() const would prevent the copy.

offline/packages/trackbase/ActsGeometry.h
```diff
@@ -49,7 +38,6 @@ class ActsGeometry {
       Acts::Vector3 world,
       TrkrDefs::subsurfkey& subsurfkey) const;
 
- private:
```
hupereir, line 52: With the changes above, you should be able to keep the members private, while keeping the same speed. 
Also: you would then not need to change any of the rest of the code: the public interface remains the same.

osbornjd:
> @osbornjd The diagnostic is correct, (essentially your accessors were making a full copy each time), but the fix, that consists in removing the accessors and making the members public is not. You should use const references instead. This would prevent creating copies, without having to change the public API nor expose the members. See inline examples below

@hupereir I actually tried this at first. What I don't understand is that passing by const reference did not improve the timing at all. I could not understand why, hence I made them public which is what we had before. If you could explain to me why the const ref passing did not improve the timing I would be thrilled as I could not find anything on stack overflow...

osbornjd:
I recompiled everything, hence was very surprised it did not fix things. As you say, it only requires like 4 lines of changed code. Maybe I made a mistake, I will try it again and see if I can reproduce it.

PR #1590, opened 2022-07-25
change to cluster v4. For real this time

bogui56:
[comment]: <> (Please tell us something about this pull request)

## Types of changes
[comment]: <> ( What types of changes does your code introduce? Put an `x` in all the boxes that apply: )
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work for users)
- [ ] Requiring change in macros repository (Please provide links to the macros pull request in the last section)
- [x] I am a member of [GitHub organization of sPHENIX Collaboration](https://github.com/orgs/sPHENIX-Collaboration/people), EIC, or ECCE (contact Chris Pinkenburg to join)

## What kind of change does this PR introduce? (Bug fix, feature, ...)

[comment]: <> ( What does this PR do? Linking to talk in software meeting encouraged )

## TODOs (if applicable)

[comment]: <> ( In case this is a draft PR, e.g. for running checks using Jenkins, please make the pull request as a draft: https://github.blog/2019-02-14-introducing-draft-pull-requests/  )

PR #1589, opened 2022-07-25
fix cluster z error

osbornjd:
[comment]: <> (Please tell us something about this pull request)

## Types of changes
[comment]: <> ( What types of changes does your code introduce? Put an `x` in all the boxes that apply: )
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work for users)
- [ ] Requiring change in macros repository (Please provide links to the macros pull request in the last section)
- [x] I am a member of [GitHub organization of sPHENIX Collaboration](https://github.com/orgs/sPHENIX-Collaboration/people), EIC, or ECCE (contact Chris Pinkenburg to join)

## What kind of change does this PR introduce? (Bug fix, feature, ...)

This fixes the TPC cluster z error and should fix both the TPC cluster QA and the high pT DCA_z resolution.

## TODOs (if applicable)

[comment]: <> ( In case this is a draft PR, e.g. for running checks using Jenkins, please make the pull request as a draft: https://github.blog/2019-02-14-introducing-draft-pull-requests/  )

PR #1578, opened 2022-07-14
change default to cluster v4

bogui56:
[comment]: <> (Please tell us something about this pull request)

## Types of changes
[comment]: <> ( What types of changes does your code introduce? Put an `x` in all the boxes that apply: )
- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work for users)
- [ ] Requiring change in macros repository (Please provide links to the macros pull request in the last section)
- [x] I am a member of [GitHub organization of sPHENIX Collaboration](https://github.com/orgs/sPHENIX-Collaboration/people), EIC, or ECCE (contact Chris Pinkenburg to join)

## What kind of change does this PR introduce? (Bug fix, feature, ...)

[comment]: <> ( What does this PR do? Linking to talk in software meeting encouraged )

change default to cluster v4

## TODOs (if applicable)

[comment]: <> ( In case this is a draft PR, e.g. for running checks using Jenkins, please make the pull request as a draft: https://github.blog/2019-02-14-introducing-draft-pull-requests/  )

PR #1592, opened 2022-07-26
Update QA and TrackEvaluation for cluster v4

bogui56:

## Types of changes

## TODOs (if applicable)


osbornjd:
Hi Christof, the micromegas pulls and uncertainties still look a little off. I see the QA module got updated, do we understand why there are differences? The pulls in the MVTX and INTT are also a little sharper at 0 but don't look egregious. The pulls in the TPC are also wider. This may contribute to the worse track momentum resolution at low pT, maybe.

sPHENIX-GitHub-l:
... Indeed. The TPOT errors seem to have only one value, corresponding to pitch / sqrt(12), which is what you get for cluster size 1.


From: sPHENIX-GitHub-l ***@***.***> on behalf of Joe Osborn ***@***.***>
Sent: Tuesday, July 26, 2022 12:03:27 PM
To: sPHENIX-Collaboration/coresoftware
Cc: Subscribed
Subject: [EXTERNAL] Re: [Sphenix-github-l] [sPHENIX-Collaboration/coresoftware] Update QA and TrackEvaluation for cluster v4 (PR #1592)


Hi Christof, the micromegas pulls and uncertainties still look a little off. I see the QA module got updated, do we understand why there are differences? The pulls in the MVTX and INTT are also a little sharper at 0 but don't look egregious. The pulls in the TPC are also wider. This may contribute to the worse track momentum resolution at low pT, maybe.

osbornjd:
The previous QA had a distribution of cluster uncertainties though? Were there different cluster sizes before?
Screen Shot 2022-07-26 at 2 11 55 PM

hupereir:
To be honest, in the avalanche of successive commit I have lost track of what got broken and when. 
What I know is that the TPOT error and pulls are correct in https://github.com/sPHENIX-Collaboration/coresoftware/pull/1578, possibly because it is still v3 being used despite the PR title ? 
It is definitly broken in https://github.com/sPHENIX-Collaboration/coresoftware/pull/1590, event though the change has been pushed already ... 
We really need to improve our pull request merging policy.

hupereir:
The cluster size itsefl is correct, the distribution has not changed since I updated the Micromegas geometry to ACTS. It is correct.

osbornjd:
@hupereir the last correct QA report is from this [PR](https://github.com/sPHENIX-Collaboration/coresoftware/pull/1589). The only thing missing from this is the efficiency, which bombed out after the MVTX strobe PR was merged. Any change after this PR is from the new cluster version4

hupereir:
yeah so, looking at the current ClusterErrorPara.cc code (in trackbase): 
    case TrkrDefs::micromegasId:
      if(layer==55){
	phierror = pitcherr_phi_mm1;
	zerror = pitcherr_z_mm1;
      }else if(layer==56){
	phierror =pitcherr_phi_mm2 = 31.6;
	zerror = pitcherr_z_mm2 = 0.2;
      }
So: same error hardcoded for all TPOT clusters. Hopefully this code is unfinished. 

Hugo

bogui56:
Yes, this code is unfinished, its a placeholder for the time being. As I said on multiple occasions the errors in the parameterization 
are the 1st approximation. For the tpc they seem to work but need finetuning. For the MVTX the errors 
are more or less identical since they are only based on the cluster width in bins in z and phi direction.

For the INTT and the micromegas we have to see how to propagate the errors in case you extract the positions
from charge sharing in case multiple strips are hit. in the INTT in most cases one particle fires one strip.
In case you have to adjacent strips hit it is 50 - 50 whether is was 2 particles nearby or one particle sharing the charge. 

For the micromegas charge sharing seems to be the default situation. I'll see if a parametrization
based on cluster size and angles makes sense or if, for strip detectors we need a separate cluster class 
that still contains the errors derived from the cluster shape, which in this case actually make sense, contrary 
to the tpc case. The strip hit multiplicity is low, so we could afford a fatter cluster class.

Maybe we should schedule a discussion in the next tracking meeting. 

Cheers

   Christof

osbornjd:
If these aren't tuned yet then we should revert back to cluster v3 and not be using cluster v4

sPHENIX-GitHub-l:
I agree, we should revert to cluster v3 until the new version has been tuned.

Chris, can you do that please?

Thanks
Tony

hupereir:
> For the micromegas charge sharing seems to be the default situation. I'll see if a parametrization based on cluster size and angles makes sense or if, for strip detectors we need a separate cluster class that still contains the errors derived from the cluster shape, which in this case actually make sense, contrary to the tpc case. The strip hit multiplicity is low, so we could afford a fatter cluster class. Maybe we should schedule a discussion in the next tracking meeting. Cheers Christof

Christof: 

it might be enough to just set a cluster size dependent scale factor corresponds to the RMS of the (cluster-truth) residuals for this cluster size, in order to set the pull width to unity (again, for each cluster size). Additional dependence on track angle should not be necessary because the cluster size and track angle should be very correlated, except for overlaping clusters

Other than that: micromegas clusters being 1D you really need only one float (at most) per cluster if you need to store the uncertainty, and then you don't need the size. And finally: both micromegas layers correspond to ~ one layer of the TPC, over ~10% of the acceptance. So you expect the micromegas clusters to be only 1/640 of the TPC clusters. Indeed, I believe they can be a little bigger without any significant impact on the memory footprint. 

I agree it might be good to discuss this next monday.

bogui56:
I know all that and I am working on it.

sPHENIX-GitHub-l:
Hi Everybody, 

I may need a little help here. 
Build gcc-8.3 is in failure, but I can't spot any errors in the report files.
Can anybody see what the issue is?

Thanks 

   Christof

blackcathj:
Follow up to Chris' comment, the QA job time out is due to a segmentation fault of the QA simulation/reco process 10min into the run, which hang the later drawing process and lead to an apparent time out. The raw output for the  QA simulation/reco process looks like: 

```
 drop si_track 28 with eta 0.827518 and z 8.01981 because crossing is undefined 
 drop si_track 27 with eta 0.708346 and z 7.47044 because crossing is undefined 
 ....
Command terminated by signal 11
	Command being timed: "timeout --preserve-status --kill-after=1s --signal=9 1d root.exe -b -q Fun4All_G4_sPHENIX.C(1000,"NullInput","G4sPHENIX_job_7")"
	User time (seconds): 788.31
```
More at https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-tracking-low-occupancy-qa/2389/console

adfrawley:
Hi All,

The message is a diagnostic output that should have been covered with a verbosity flag. I will make a PR to turn it off.

Tony

offline/QA/modules/QAG4SimulationMicromegas.cc
```diff
@@ -363,49 +396,106 @@ void QAG4SimulationMicromegas::evaluate_clusters()
...
+    std::cout << " g4part is primary..." << std::endl;
+    float gprimary = trutheval->is_primary(g4particle);
+
+    if (Verbosity() > 0)
+      std::cout << PHWHERE << " PHG4Particle ID " << gtrackID << " gembed " << gembed << " gflavor " << gflavor << " gprimary " << gprimary << std::endl;
+    std::cout << "get truth clusters..." << std::endl;
+    // Get the truth clusters from this particle
+    const auto truth_clusters = trutheval->all_truth_clusters(g4particle);
+
+    std::cout << "fill position vectors..." << std::endl;
```
blackcathj, line 425: Please note there is a few places generate `cout` without `Verbosity()` based option to suppress them: 
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-default-detector-pipeline/lastSuccessfulBuild/artifact/macros/detectors/sPHENIX/Fun4All_G4_sPHENIX.log
```
evaluate clusters...
evaluate clusters...
got trutheval
got cluseval
got histmanager
histos assigned
g4part #1
accessing g4part...
 g4part is embedd...
 g4part is primary...
get truth clusters...
fill position vectors...
get clusters from g4part...
done cfg4
g4part #2
accessing g4part...
```

bogui56:
Hi Jin, 

the printouts will be removed again. 

Currently the pythia qa job in jenkins is the only place that consistently crashes and I can access the output file. 
So I put a few print statements in to see where things actually crash. Running my own Fun4All jobs I do not see this crash. 

Its in the Micromega QA module in a part where the trutheval module is called to return a list of the g4hits associated with a given g4particle. 
The same evalmodule is used in the Evaluator and doesn't crash

Where can I find that actual fun4all macro that is run in the pythia qa so I can run it in my dir (with the appropriate random seed) 
to reproduce the crashes?

Cheers

   Christof

osbornjd:
@bogui56 It looks like Jenkins can run through now. It would probably be useful to have a discussion on this PR before we merge it. It looks like in the QA there are some physics degradations, e.g. in the tracking efficiency and in the low pT track momentum resolution. This may be related to the different shape in the number of TPC clusters per track as well. The uncertainty distributions of some of the clusters are also substantially different, e.g. in the TPOT - does this just mean the uncertainty parameterization needs tuning? That may also be a cause for the change in pT resolution.

blackcathj:
Two things to note from the newly introduced "trend" tool for the last check on this pull request with [tracking QA with upsilon embedded in charmged jet events, build **#855**](https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-tracking-pythiajet-qa/855/): 

1. The Upsilon peak width is higher: 
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-tracking-pythiajet-qa//plot/#Upsilon_width_Summary.csv
image

2. And the code is taking noticeably longer to run 
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-tracking-pythiajet-qa//plot/#test-default-detector.csv_Time_(s)_Summary.csv
image


Apparently from a jump in time used in `PHSimpleKFProp`: 
image

Are these intended changes?

bogui56:
Hi Jin, 

the upsilon resolution is slightly deteriorated, but should go down again once the cluster error parametrizion is fine tuned some more. 
We discussed this in the tracking meeting and decide to go ahead with this as is for the time being. 

The timing of the KFProp is a bit surprising, since by default it isn't using the cluster errors from the parameterization.
There is an option to use them, hence the code was touched, but the with the current defaults the block of code in question isn't executed. 
The ACTS surfaces etc however are and there is a timing issue with the GeoContext. 

- Christof

osbornjd:
If you are satisfied with the current state @bogui56 we can merge this as discussed yesterday. Please also make a PR to the macros repository so that we have a global flag to switch between cluster v3 and v4 also.

adfrawley:
Hello Jin and Christof,

The timing jump for KFProp seems to be related to the implementation of the alignment transformations. I am looking into it. It is possibly just map lookup speed, which can be further improved.

Tony

osbornjd:
As far as I can tell the QA looks mostly the same as before but now the clang build fails. @bogui56 can you fix this and then we can merge it?

sPHENIX-GitHub-l:
Hi Joe, 

it looks a little better now. At least the upsilon mass resolution does, maybe. 
I think we can merge now, I suspect the pt resolution deterioration happens
in track we would reject anyway in the end. The cluster error parameterization 
is sensitive to the track selection you apply. We can refine that later. 

The clang error is due to a scalefactor thats defined, but not used in the current tune. 
I'll comment them out and resubmit.

cheers

   Christof

osbornjd:
Sounds good, I'll wait for the clang build to finish succesfully on the next push and then merge.

osbornjd:
Merging this, start the fireworks

sPHENIX-GitHub-l:
Hi, 

I can't do all the tuning for v4 myself. 
I am working on the micromegas part now. 

Do you have a plan how to finish all this is you revert to v3 now?

cheers

   Christof

PR #1595, opened 2022-07-27
Revert "change to cluster v4. For real this time"

osbornjd:
Reverts sPHENIX-Collaboration/coresoftware#1590 and moves us back to cluster v3


osbornjd:
@pinkenburg can we just merge this right away, which puts us back at cluster version v3

pinkenburg:
sure - no problem

PR #1642, opened 2022-08-31
Implement global alignment constants

rboucher43:

## Types of changes

Adds alignment parameters to Acts local to global transformations, and takes alignment parameters from the database. Alignment parameters are currently all zero in the database. 

## TODOs (if applicable)


pinkenburg:
Please follow the include file conventions in https://wiki.sphenix.bnl.gov/index.php/Codingconventions#Include_Files
Breaking them can lead to very hard (means impossible for even most expert users) to debug issues
Please remove -L$(ROOTSYS)/lib from the LIBADD, it is already in LDADD (where the -L flags belong)

offline/packages/trackreco/PHActsTrkFitter.cc
```diff
@@ -469,7 +479,88 @@ SourceLinkVec PHActsTrkFitter::getSourceLinks(TrackSeed* track,
...
+	  float globphi = atan2(global(1),global(0))*180.0/M_PI;
+	  std::cout << "Check in TrkFitter: global phi " << globphi << " hitsetkey: " << hitsetkey <<" global: " << std::endl << global << std::endl;
+
+	  auto x = cluster->getLocalX() * 10.0;   // mm
+	  auto y = cluster->getLocalY() * 10.0;
+	  
+	  if(trkrid == TrkrDefs::tpcId)
+	    {
+	      // must convert local Y from cluster average time of arival to local cluster z position
+	      double drift_velocity = 8.0e-3;  // cm/ns
```
osbornjd, line 502: We should obtain this from the globally settable variable in ActsGeometry

osbornjd (changes requested):
1. In general there is a lot of old code that is commented out - we should remove it completely to not confuse ourselves later (e.g. in ActsGeometry). 
2. A lot of the methods in ActsGeometry had their constness removed while trying to dig through an Acts compiler error about constness. But, if I recall, the compiler error didn't actually have to do with the constness in our class. If so, we should reimplement them. @adfrawley any comment on this, since I think (if I remember correctly) I talked to you about it?
3. We should also understand the drop in efficiency and pT resolution that Jenkins sees. It is probably related to the z position of the TPC clusters, as in the TPC QA the cluster z residuals look almost inverted.

Otherwise this looks like a great PR, I made a few other small comments but this is excellent to see!

sPHENIX-GitHub-l:
Hi Joe,

I don't recall the part about constness not being the problem. If I recall correctly,  the problem we had was that we could not update the geometry context if the ActsTrackingGeometry class was const, since const classes cannot have non-const members. I had to remove the constness from all members to get things to compile. Possibly I misunderstood something.

You are correct about map2, that was overlooked when deleting some debugging code. I agree, I don't know how it compiled.
We will fix the ActsGeometry include.
Yes, sPHENIXActsDetectorElement::use_alignment is obsolete, we missed it.

The block of code in PHActsTrkFitter enabled by the test_transform flag was never intended to be left there permanently. But we will fix that, since we decided to keep that block there for a while.

Thanks for the useful comments!
Tony


osbornjd:
Okay, I couldn't remember on the const issue. That's fine then, I just needed a refresher. 

My guess is once the TPC z cluster residual issue gets understood it will fix the issues at the track level.

adfrawley:
I do not see any reported issues that are related to this PR.

Tony

osbornjd:
Are the changes in the TPC cluster residuals understood? That (I suspect) negatively impacts the tracking and vertexing QA, especially in the z direction.

osbornjd:
Looks good to me, @pinkenburg I assume the overlap failures are from the ongoing hcal issue so I will go ahead and merge this.

PR #1645, opened 2022-09-01
Acts fit fixes

hupereir:
This PR: 
- cleans up GenFit trackfit
- fixes the sourcelink order in Acts Trackfit. This is necessary because the SVTX+MM fit check this order and arbitrarily remove surfaces if it is not met
- fix crash in SVTX+MM fit, due to incorrect insertion of the fitted track in the map. 
- improves (speed up) track insertion in the maps, print a warning (and fixes a memory leak) in case of duplicated keys.

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackbase_historic/SvtxTrackMap_v2.cc
+++ b/offline/packages/trackbase_historic/SvtxTrackMap_v2.cc
@@ -9,9 +9,6 @@
 #include <ostream>      // for operator<<, endl, ostream, basic_ostream, bas...
 #include <utility>      // for pair, make_pair
 
-using namespace std;
-
-
 SvtxTrackMap_v2::SvtxTrackMap_v2()
   : _map()
 {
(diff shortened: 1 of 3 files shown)


blackcathj:
The failed `Tracking QA at low occupancy` is due to larger memory foot print with the new hadronic calorimeter. Please ignore that part for now. 

There is a serious broadening of the DCA distribution, and please check: 
https://nbviewer.org/github/sPHENIX-Collaboration/QA-gallery/blob/jenkins-sPHENIX-test-tracking-pythiajet-qa-757-test-tracking_Event600_Sum10/QA-vertexing.ipynb
image

hupereir:
Hi Jin,

Thanks for spotting this. Indeed this is worrisome. There is no 
difference on the single track momentum resolution though as far as I 
can tell.

So the question is to Joe I guess: in which order does acts expects to 
have the surfaces so that the fit gives the best DCA resolution  ?

Depending on the answer, I can revert to the old ordering, and remove 
the (then invalid) check on the surface ordering done in the 
silicon+TPOT fit.


Hugo

osbornjd:
> So the question is to [@osbornjd](https://github.com/osbornjd) I guess: in which order does acts expects to
have the surfaces so that the fit gives the best DCA resolution  ?

I believe Acts expects the source links (I.e. surfaces) to be ordered in order of radius. So my guess is that the MVTX layers aren't being included in the fit (or if they are, in some weird way), giving similar single track pT resolution from the TPC component of the track but poor DCA resolution from the MVTX part of the track.

hupereir:
@osbornjd 
So I forced pushed an update to the branch fixing the DCA issue (hopefully. Let's see what Jenkins says) 
It has nothing to do with reordering the source links, but from the fact that a given track must have a valid track id before calling getTrackFitResult(), which some of my cleanup broke.

osbornjd:
@hupereir the DCA has returned to its nominal performance from the latest commit, as far as I can tell (the p values are from an outdated reference, which I just updated).

hupereir:
> @hupereir the DCA has returned to its nominal performance from the latest commit, as far as I can tell (the bad p values are from an outdated reference, which I just updated).

@osbornjd 
yes. I also checked the other histograms and could not spot any problem. (but feel free to double check). As far as I can tell, this is good to go. 
Lessons learned: apparently ACTS does not care about the order by which the sourcelinks (and measurements) are passed: I changed it and it had no impact on momentum resolution. The Silicon + MM fit does care, because we make an explicit check on this, and remove surfaces that are not properly ordered (In fact, I think we could safely remove this part of the code in a separate PR). 
Still, I think it makes more sense (to avoid future confusion) to pass the clusters, sourcelinks, measurement in the right order, namely outwards starting from the beam.
Also: this PR does fix a crash :) for the silicon + mm fit (essentially an incorrect cast)

Hugo

PR #1653, opened 2022-09-08
Tracking memory leak clean up

osbornjd:

## Types of changes

This PR aims to clean up some of the memory leaks in the tracking.

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackbase/TrkrHitSetContainerv1.cc
+++ b/offline/packages/trackbase/TrkrHitSetContainerv1.cc
@@ -10,14 +10,13 @@
 #include "TrkrHitSetv1.h"
 
 #include <cstdlib>
-#include <memory>
 
 void TrkrHitSetContainerv1::Reset()
 {
-  for( const auto& pair:m_hitmap )
-  { delete pair.second; }
-  
-  m_hitmap.clear();
+  for(auto&& [key, hitset] : m_hitmap)
+    { delete hitset;}
+  Map empty;
+  m_hitmap.swap( empty );
 }
 
 void TrkrHitSetContainerv1::identify(std::ostream& os) const
@@ -82,8 +81,7 @@ TrkrHitSetContainerv1::findOrAddHitSet(TrkrDefs::hitsetkey key)
   auto it = m_hitmap.lower_bound( key );
   if( it == m_hitmap.end() || (key < it->first ) )
   {
-    std::unique_ptr<TrkrHitSetv1> hs = std::unique_ptr<TrkrHitSetv1>();
-    it = m_hitmap.insert(it, std::make_pair(key, hs.release()));
+    it = m_hitmap.insert(it, std::make_pair(key, new TrkrHitSetv1));
     it->second->setHitSetKey( key );
   }
   return it;
--- a/offline/packages/trackbase/TrkrHitSetContainerv1.cc
+++ b/offline/packages/trackbase/TrkrHitSetContainerv1.cc
@@ -56,7 +56,7 @@ void TrkrHitSetContainerv1::removeHitSet(TrkrDefs::hitsetkey key)
     {
       TrkrHitSet* hitset = iter->second;
       delete hitset;
-      m_hitmap.erase(key); 
+      m_hitmap.erase(iter); 
     }
 }
 
--- a/offline/packages/trackbase/TrkrHitSetContainerv1.cc
+++ b/offline/packages/trackbase/TrkrHitSetContainerv1.cc
@@ -10,6 +10,7 @@
 #include "TrkrHitSetv1.h"
 
 #include <cstdlib>
+#include <memory>
 
 void TrkrHitSetContainerv1::Reset()
 {
@@ -81,7 +82,8 @@ TrkrHitSetContainerv1::findOrAddHitSet(TrkrDefs::hitsetkey key)
   auto it = m_hitmap.lower_bound( key );
   if( it == m_hitmap.end() || (key < it->first ) )
   {
-    it = m_hitmap.insert(it, std::make_pair(key, new TrkrHitSetv1));
+    std::unique_ptr<TrkrHitSetv1> hs = std::unique_ptr<TrkrHitSetv1>();
+    it = m_hitmap.insert(it, std::make_pair(key, hs.release()));
     it->second->setHitSetKey( key );
   }
   return it;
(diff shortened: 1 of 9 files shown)


osbornjd:
Please do not merge, would like to see the valgrind report from Jenkins

offline/packages/trackreco/ALICEKF.h
```diff
@@ -28,26 +28,24 @@ class ALICEKF
...
+    _B = B;
     //    _B = PHFieldUtility::GetFieldMapNode(nullptr,topNode);
     _cluster_map = cmap;
     _fieldDir = fieldDir;
     _max_sin_phi = max_sin_phi;
     _v = verbosity;
     _min_clusters_per_track = min_clusters;
   }
+
+  ~ALICEKF() { delete _B; }
```
hupereir, line 47: @osbornjd 
In principle _B could be made a unique pointer in the header.
You would assign it in the constructor via either _B(B) in the initializer list (so before the first {)
or via _B.reset(B); if elsewhere.
Doing so, you would not need to delete in the destructor.

offline/packages/trackreco/PHSimpleKFProp.h
```diff
@@ -39,7 +39,7 @@ class PHSimpleKFProp : public SubsysReco
 {
  public:
   PHSimpleKFProp(const std::string &name = "PHSimpleKFProp");
-  ~PHSimpleKFProp() override = default;
+  ~PHSimpleKFProp() { delete _field_map; }
```
hupereir, line 42: you might have a double deletion here. _field_map is passed to ALICEKF, and deleted with the destructor of ALICEKF. 
Here you would then delete it a second time ... 
In any case here also you could make _field_map a std::unique_ptr, so that the deletion would be automatic.

simulation/g4simulation/g4mvtx/PHG4MvtxHitReco.cc
```diff
@@ -579,13 +579,13 @@ int PHG4MvtxHitReco::process_event(PHCompositeNode *topNode)
           // generate the key for this hit
           TrkrDefs::hitkey hitkey = MvtxDefs::genHitKey(vzbin[i1], vxbin[i1]);
           // See if this hit already exists
-          TrkrHit *hit = nullptr;
-          hit = hitsetit->second->getHit(hitkey);
+	  std::unique_ptr<TrkrHit> hit(nullptr);
```
hupereir, line 582: std::unique_ptr is automatically initialized to null so you don't need to say it explicitly
std::unique_ptr<TrkrHit> hit; is enough

simulation/g4simulation/g4mvtx/PHG4MvtxHitReco.cc
```diff
@@ -579,13 +579,13 @@ int PHG4MvtxHitReco::process_event(PHCompositeNode *topNode)
...
+          hit = std::unique_ptr<TrkrHit>(hitsetit->second->getHit(hitkey));
           if (!hit)
           {
             // Otherwise, create a new one
-            hit = new TrkrHitv2();
-            hitsetit->second->addHitSpecificKey(hitkey, hit);
+            hit = std::make_unique<TrkrHitv2>();
+            hitsetit->second->addHitSpecificKey(hitkey, hit.get());
```
hupereir, line 588: That's definitly a leak fix yes :)
As far as I remember addHitSpecificKey clones the hit (but feel free to double check), so the source was not deleted.

hupereir (commented):
Hi Joe, 
A few comments on the changes you made ...

simulation/g4simulation/g4mvtx/PHG4MvtxHitReco.cc
hupereir: Actually, double checking in https://github.com/sPHENIX-Collaboration/coresoftware/blob/master/offline/packages/trackbase/TrkrHitSetv1.cc line 58, there is no copy of the hit done. So I suspect this would crash: you delete the hit after it has been inserted.
hit.release() would fix it.

osbornjd:
@hupereir thanks for the comments, I mostly just made this to see the valgrind report from Jenkins to see if this actually fixed the reported leaks from Chris. But I will also keep playing and implement your suggestions.

simulation/g4simulation/g4mvtx/PHG4MvtxHitReco.cc
osbornjd: I worried about release here because the pointer is accessed right below on line 592. If I read the unique_ptr documentation correctly, release will then render the smart pointer as null and then the `addEnergy` call will be called on a nullptr.

hupereir: @osbornjd you are right. After you do release the unique_ptr is invalid so that will break. 
On the other hand what you want to make sure is that the hit in the unique_ptr does not get deleted as soon as the unique_ptr gets out of scope, because you would then end up with a dangling pointer in the hitset.
All in all I am not sure you need to use a unique_ptr here at all, then ...
Whatever you do, the ownership of the hit is to the hitset, and it should be responsible for deleting it. 
I can look in the code in more detail if needed.

osbornjd: Ha, in fact using either `get` or `release` seems to lead to a seg fault. So the memory leak must come in from some way that the container is deleting the memory, I guess...

pinkenburg:
I have seen TObjects not playing nicely with uniq_ptr's - at least I have a vague memory of that. That was more in the context that a uniq_ptr didn't release its memory properly, but I would be careful with more fancy constructs with TObjects

offline/packages/trackbase/TrkrHitSetContainerv1.cc
```diff
@@ -50,7 +50,15 @@ TrkrHitSetContainerv1::addHitSetSpecifyKey(const TrkrDefs::hitsetkey key, TrkrHi
...
 
 void TrkrHitSetContainerv1::removeHitSet(TrkrDefs::hitsetkey key)
-{ m_hitmap.erase(key); }
+{ 
+  auto iter = m_hitmap.find( key );
+  if(iter != m_hitmap.end()) 
+    {
+      TrkrHitSet* hitset = iter->second;
+      delete hitset;
+      m_hitmap.erase(key);
```
hupereir, line 59: it is more efficient to use .erase(iter) than erase(key). The latter redo the search in the map, not the former.

PR #1654, opened 2022-09-09
Cluster edit

david-stewart:

## Types of changes

This pull request is a major refactoring of the request that was #1619 which it supersedes and is now closed.
It adds the PHObject TrkrTruthTrackContainer (full of TrkrTruthTracks) for each event. Each track contains phi,eta,pt, kinematics of the truth track, as well as 3D location of the vertex. These are to allow for simple, naive, track matching. It also contains a vector of references for Clusters in the TPC GEMs which can be used to get TrkrClusters from a newly filled TrkrClusterContainer.

The TrkrClusters for the truth tracks and generated from all the hits in each layer in the TPC (maximum of 54 clusters, therefore, per track). This is done by taking a weighted average phi and T(or Z) location of all electron hits, as well as a "width" in phi and Z from between the maximum and minimum bins in all hits, as calculated in the PHG4TpcPadPlaneReadout.cc module.

## TODOs (if applicable)


pinkenburg:
It dies because libtrack_reco does not exist yet (it is build later). I can't find anything in the code which needs this (and the dependency on trackbase existed before). Likely just removing -ltrack_reco in the Makefile.am will fix this.
Please replace the tabs you used in the file lists in the Makefile.am by the appropriate number of spaces. It doesn't matter right now but tabs are seen as indicating a command, if someone accidentally removes a continuation line character (\) ahead of a tab indented file, it is interpreted as a command which then gives completely misleading error messages making finding this really hard

david-stewart:
Thanks for the debugging, Chris! So I went back to file the Makefile.am and fixed the tab vs spaces. I suspect that this is a case-in-point for your insight that the \tab vs \space make it hard to reason about the file, because when I checked the main branch, the `-ltrack_reco` is also in the `g4tpc\Makefile.am` file. So, I just fixed by \tab vs \space issue and pushed the commit. If this doesn't work, I will go and remove the `-track_reco` like alltogether.

pinkenburg:
TrkrHitTruthClusters.h needs to be added to pkginclude_HEADERS in the Makefile.am so it gets installed where other packages can find it

david-stewart:
I removed some of the old code under the Dump directory which was using the old file names of the initial commit. I may have to go back and add in some of the functionality later for the new structures.

david-stewart:
I'm not sure about the Valgrind error, when I looked for the log I got an error that "A problem occurred while processing the request."
That being said, I have looked at the other pull requests and see that there are also Valgrind check errors there, too. So, should I ignore it, or should I run Valgrind myself to try and see what it is complaining about?
The other thing is that I removed the code form the Dump module, which appears to print out the tree for the user. I can go back there and mock something up is wanted.

pinkenburg:
I get the same error when trying to look at the valgrind log. valgrind is a runtime check - with the current macros your code is not executed so anything in that report is not your problem. 
It looks like you removed a class from the dump module which has been deleted. There is no problem with that, but if we have any DST where these TrkrHitTruthClusters were saved - it will become unreadable if you remove those classes (one of the dirty secrets of schema evolution - once it's written out you will have to keep the class forever which is why we really pay attention to this) 
Please go through your includes - order of "" versus <> matters. include "" first, then include <>, no exceptions
#include "TrkrCluster.h"
#include <phool/PHObject.h>
#include "TrkrClusterContainer.h"

david-stewart:
The order of order of "" versus <>  has been corrected in all trees. I haven't made any trees with TrkrHitTruthClusters in them that I am using, but I suppose it is possible that there are some that someone generated after submitting the TrkrHitTruthClusters in a prior commit and removing it with the TrkrClusters in this one; however, I think it is unlikely.

blackcathj:
Running a Jenkins maintenance, which may break the check. If fails, will restart it again later manually

osbornjd:
@david-stewart is this ready to go in? I can't remember the conclusion from the tracking meeting on whether or not we would use pt/phi/pseudorapidity or px/py/pz, but all jenkins checks pass so in principle if you think this is ready to go we can merge it. If there are still some changes/improvements you'd like to add then we can wait also - the PR isn't marked as draft so I wasn't sure.

david-stewart:
> @david-stewart is this ready to go in? I can't remember the conclusion from the tracking meeting on whether or not we would use pt/phi/pseudorapidity or px/py/pz, but all jenkins checks pass so in principle if you think this is ready to go we can merge it. If there are still some changes/improvements you'd like to add then we can wait also - the PR isn't marked as draft so I wasn't sure.

Hi Joe,
I removed it from Draft myself when I thought that it was ready to go in.
Good point on the formatting. The meeting didn't address whether pt/phi/pseudorapidity or px/py/pz is preferred. My impression from talking with Tony is that pt/phi/eta are probably the best points for doing the track matching, which is the point of the code anyhow. (Judging from `SvtxEvaluator.cc` around line 1908-1917 that both are used). As far as I know, track matching is the primary customer, so I vote for it. If there are alternative thoughts, I'd be happy to entertain them.

osbornjd:
> > @david-stewart is this ready to go in? I can't remember the conclusion from the tracking meeting on whether or not we would use pt/phi/pseudorapidity or px/py/pz, but all jenkins checks pass so in principle if you think this is ready to go we can merge it. If there are still some changes/improvements you'd like to add then we can wait also - the PR isn't marked as draft so I wasn't sure.
> 
> Hi @osbornjd , I removed it from Draft myself when I thought that it was ready to go in, but am not sure if there is a more formal procedure. Good point on the formatting. The meeting didn't address whether pt/phi/pseudorapidity or px/py/pz is preferred. My impression from talking with Tony is that pt/phi/eta are probably the best points for doing the track matching, which is the point of the code anyhow. (Judging from `SvtxEvaluator.cc` around line 1908-1917 that both are used). As far as I know, track matching is the primary customer, so I vote for it. If there are alternative thoughts, I'd be happy to entertain them.

Sounds reasonable to me - I'll take this as your approval that this is ready to go. I will merge Christof's new cluster v4 PR first and then we can do this one, since the clusterv4 has been in the pipeline for a while now.

PR #1665, opened 2022-09-18
added Birk constant

Shuonli:

## Types of changes

This PR sets the Birks' constant for the HCal scintillator material(EJ200). 
The default Birks' constant for G4_POLYSTYRENE is 0.07943 mm/MeV, while the test beam paper(https://arxiv.org/pdf/1704.01461v1.pdf) suggests a Birks' constant(0.2 mm/MeV) that makes the simulation agrees with data the best.
 
Note that this PR won't entirely fix the factor of 2 difference in the visible sampling fraction.

## TODOs (if applicable)


blackcathj:
Thanks for the contribution. This is a necessary step in completing the new HCal simulation setup. The result sampling fraction with the Birk model does fall closer to the older-generation HCal simulation with simplified geometry than that without the Birk effect: 
image

Merging....

pinkenburg:
What should be done about the inner hcal? It's the same scintillator material. I wonder if setting the birks constant for the outer hcal will propagate in time to the inner hcal. These are global settings for the material but I don't know when they become active (or are being looked at), basically if it matters that they are being set in the outer hcal and are still being used in the inner hcal. gdml is just a complete mess when it comes to materials

Shuonli:
> What should be done about the inner hcal? It's the same scintillator material. I wonder if setting the birks constant for the outer hcal will propagate in time to the inner hcal. These are global settings for the material but I don't know when they become active (or are being looked at), basically if it matters that they are being set in the outer hcal and are still being used in the inner hcal. gdml is just a complete mess when it comes to materials

Yes, right now Birks constant for the outer hcal will propagate to the inner hcal and vise versa since the scintillator material for them are the same. I think in order to set different Birks' constant we need to name the scintillator material differently in the gdml files for ihcal and ohcal...

pinkenburg:
Sorry - I was stuck at the outer hcal and overlooked the ihcal

PR #1696, opened 2022-10-07
Tracker alignment

adfrawley:

## Types of changes

This PR adds a new offline/packages directory named TrackerMillepedeAlignment. It contains a new module named MakeMilleFiles, and the class Mille (copied without change from the Millepede2 area for convenience). Mille is a very simple C++ code that is used by MakeMilleFiles to properly format the data files needed by Pede as input for doing the actual alignment. It is called by MakeMilleFiles. All of the work needed to interface sPHENIX to Millepede2 is done in MakeMilleFiles.
 
The code is not complete yet. It calculates the cluster-track residuals, and the derivatives of the residuals with respect to the alignment parameters. But it is still missing the calculation of the derivatives of the cluster-track residuals with respect to the fitted track parameters. 

None of the repository macros know about this module yet.

## TODOs (if applicable)


osbornjd (commented):
Looking through this after looking at the Acts code, I think we should consider how we want to go about actually calculating all the things that will go into Mille. Right now it looks like the jacobians are basically calculated by projecting the track state to the surface and then calculating the PCA. From this (if my understanding is correct), the derivatives are calculated by introducing a small perturbation and calculating the difference.

After reading through the Acts alignment code the last few days, many of these calculations can be done analytically with a few lines of code using the Acts track state and track fit output. The problem is we dump this info after the track fit in favor of staying with our Svtx data objects.  So doing this would require a big change in the underlying track state objects that we carry around.

hupereir:
On the other hand, SvtxTrackState is versioned. So adding information in there is doable without breaking backward compatibility. 
Of course there is the question of how much information must be added and what it will do on output file sizes

osbornjd:
> On the other hand, SvtxTrackState is versioned. So adding information in there is doable without breaking backward compatibility. Of course there is the question of how much information must be added and what it will do on output file sizes

Indeed we could just create a new version and have it only contain an object that corresponds to the Acts track state. I also don't know what that will do to the output size because the Acts track states do carry (I think) more excessive information than we usually need.

osbornjd:
Thinking some more, this may not matter after running the module which removes all the track states except those at the vertex and in the calos. It may be worth affording any slight gain in memory consumption for ease of actually analytically calculating these values. The output file size consumption may not be substantially larger since we are only talking about ~4 track states instead of ~60.

sPHENIX-GitHub-l:
I am inclined to merge this so that others can work on it. It is not complete, but it is not part of our reconstruction chain, so it will not bother anyone.

What is needed to make it functional is the addition to MakeMilleFiles of a calculation of the derivatives of the residual with respect to the track parameters. Joe is thinking about that.

There is a possibility of getting the derivatives of the alignment angles from the tracking code, instead of numerically as is done here. That would be good, but is not needed to make it functional. I would like to have both methods in place so we can check them against each other.

I will be at a workshop for the next week, and will not have a chance to work much on tracking.

Tony

PR #1722, opened 2022-11-02
Padplane readout phi

mohaas33:

## Types of changes

## Dead areas (GEM frames) for TPC padplane

Dead segments are implemented. 
Electrons are redistributed to the neighboring pads: <https://indico.bnl.gov/event/17576/contributions/69898/attachments/44007/74253/padplane_10252022.pdf> 

## TODOs (if applicable)


mohaas33:
Hello,
> 
> 1. It looks like the cluster size increases on average. Is this understood? This changes the uncertainties based on the parameterization look up, and also the residuals (likely from the size of the cluster shifting the centroid). 

The size of the pads in phi has decreased as phi range is reduced by ~4% (dead areas). So the number of pads per cluster should be higher.

> 2. The clustering efficiency drops overall, which makes sense given the new dead areas. It looks on average that it drops about ~4% per region of the TPC, which I think is also what you reported in the tracking meeting. However in some layers there is a large spike in the drop in efficiency - is this understood?

In addition to the dead areas, I have implemented a redistribution of charges. All charges above the dead area are put on the closest sensitive pads/stripes which can smear the clusters in the layers close to the dead areas and matching could fail.
And it could also depend on how the efficiency is defined, If you expect a cluster in the dead area and you treat it as a cluster in the layer of TPC it will rise to inefficiency.

> 3. There is a slight drop in the tracking efficiency and the tracking purity, which may be related to the above comments (e.g. maybe the worse cluster efficiency leads to a drop in seeding efficiency because the links are not able to be chained together). We should understand this too.

As I am adding charges to the pads which should not have those charges in the transition regions with no pads, some impurity should appear. It is hard to estimate the exact effect on the tracking.

adfrawley:
Hello Evgeny and Joe,

It seems to me that the modeling of the redistribution of charge near the dead areas is critical. This redistribution effectively puts the charge at the wrong value of phi.

Do we believe that the redistribution of charge is correct as implemented? Do we need to implement fiducial cuts on clusters near the dead areas so that they do not interfere with seeding and tracking?

Tony

mohaas33:
Hello Tony,
Right now I have put half of the charges from the dead area into the closest layer in the active area as electric field lines have to end on the GEMs.
Is it possible to apply some weights to those clusters for the fitting procedure? 
So those will be taken into account but have less contribution (be less important for the fit).
Cheers
Evgeny

pinkenburg:
Do not worry about the scan build failure

PR #1724, opened 2022-11-02
Fix error in code in storing TrkrCluster sizes for phi and time

david-stewart:

## Types of changes

While working on the TrkrCluster matching for the Embedded to Reco tracks, I found a bug in my code for the arguments of  TrkrCluster->Set{Phi,Z}Size when filling the clusters for the truth tracks. This code fixes that error.

## TODOs (if applicable)

### DIFF
--- a/simulation/g4simulation/g4tpc/TpcClusterBuilder.cc
+++ b/simulation/g4simulation/g4tpc/TpcClusterBuilder.cc
@@ -2,22 +2,54 @@
 #include <trackbase/TrkrClusterv4.h>
 #include <trackbase/TpcDefs.h>
 #include <g4detectors/PHG4TpcCylinderGeom.h>
+#include <algorithm>
 
 class TpcClusterBuilder;
+using std::cout, std::endl;
 
 TpcClusterBuilder& TpcClusterBuilder::operator+=(const TpcClusterBuilder& rhs) {
   // layer, side, and layerGeom won't change between different additions, but they
   // but need to be set by the first addition
+  if (rhs.neff_electrons==0) return *this; 
+
   layerGeom = rhs.layerGeom;
   layer     = rhs.layer;
   side      = rhs.side;
 
+  if (rhs.hasPhiBins) {
+    if (!hasPhiBins) {
+      phi_bin_lo = rhs.phi_bin_lo;
+      phi_bin_hi = rhs.phi_bin_hi;
+    } else {
+      int rhs_phi_bin_lo = rhs.phi_bin_lo;
+      int rhs_phi_bin_hi = rhs.phi_bin_hi;
 
-  if (rhs.phi_bin_lo < phi_bin_lo) phi_bin_lo = rhs.phi_bin_lo;
-  if (rhs.phi_bin_hi > phi_bin_hi) phi_bin_hi = rhs.phi_bin_hi;
+      // if this is the case, wrap the lower values to be higher
+      if ( (phi_bin_lo - rhs_phi_bin_lo) > (nphibins/2)) {
+        rhs_phi_bin_lo += nphibins;
+        rhs_phi_bin_hi += nphibins;
+      } else if ( (rhs_phi_bin_lo - phi_bin_lo) > (nphibins/2)) {
+        phi_bin_lo += nphibins;
+        phi_bin_hi += nphibins;
+      }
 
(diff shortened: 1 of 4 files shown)


simulation/g4simulation/g4tpc/TpcClusterBuilder.cc
```diff
@@ -46,16 +81,14 @@ TpcClusterBuilder::PairCluskeyCluster TpcClusterBuilder::build(MapHitsetkeyUInt&
   TrkrClusterv4* cluster = new TrkrClusterv4();
   cluster->setPosition ( 0, phi_mean );
   cluster->setPosition ( 1, time_integral / neff_electrons);
-  cluster->setPhiSize  ( phi_bin_hi  - phi_bin_lo  +1);
-  cluster->setZSize    ( time_bin_hi - time_bin_lo +1);
+  cluster->setPhiSize  ( static_cast<char>(phi_bin_hi-phi_bin_lo+1) );
```
osbornjd, line 84: Why is this static casted as a char?

david-stewart: The actual struct member is of type `char`, as is the argument in the setter function:
`void setZSize(char zsize) { m_zsize = zsize; }` (from `TrkrClusterv4.h`). 
The argument will automatically cast, so the static_cast is functionally redundant; it just reminds the user of the memory constraint (and therefore likely values) of the member being set.
If preferred, I'd be happy to remove it.

pinkenburg: Can you add a check if the value does not overflow a char? Both phi_bins are initialized to INT_MAX, if anything goes wrong - now or in the future - you will just blow past the limits of a char. Also I fear that passing the subtraction of two ints as a char will result in compiler warnings which then need to be suppressed explicitly.

osbornjd: David is right, the member variable is a char because it is the smallest memory primitive data type that we can store this in. Adding the check seems like the safest thing to do, then we can merge

david-stewart: What should be the desired result of a failed check? An error message and then default to -1, or propagate up some sort of return with failure code?
As for compiler error, I think (hope?) that the static_cast will indicate that we are deliberate in the math...

david-stewart: As a recommend I took the firs rout: it will print an error to cout, warning that the value is bad and is being set to  -1, such that the cluster *cannot* match with any other cluster.

blackcathj:
@david-stewart @osbornjd 
 We seem to have significant more `STDOUT` messages after merging this PR: 

Pythia event QA output (large to open): https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-tracking-pythiajet-qa//1071/artifact/macros/detectors/sPHENIX/G4sPHENIX_job_1.log/*view*/ 

And here is output line count (on top of verbosity ON in fun4all server), note the jump in job 1071
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-tracking-pythiajet-qa/plot/
image

pinkenburg:
The impossible happened... good that David put a check for this in

david-stewart:
Have run some short examples but haven't been able to re-create the access of the error code. Still digging...

blackcathj:
Thanks for digging! 

This job run from this macro branch: 
https://github.com/sPHENIX-Collaboration/macros/tree/QA-tracking-pythiajet
after an automatic merging with the master branch. 

The random seeds are set deterministically, which allow us to recreate each event:  `Fun4All_G4_sPHENIX.C(450,"NullInput","G4sPHENIX_job_1")`

pinkenburg:
These printout make working completely impossible. I'll comment this printing out. I would say the error shows that this is just not working as we think and that there is something seriously wrong upstream

david-stewart:
Sounds good. I looked for the source of the error for a while yesterday and was hoping to fix it them. I haven't been able to pinpoint it yet, but am not sure that it isn't some code in the way that my code is tabulating the bins instead of upstream problems. Hope to have it diagnosed some time tomorrow.

PR #1729, opened 2022-11-04
Ditch phtpctracker

pinkenburg:

## Types of changes
- [X ] Bug fix (non-breaking change which fixes an issue)

This PR removes PHTpcTracker from our build and cleans out the remnants of HelixHough

## TODOs (if applicable)


osbornjd:
We should poll the tracking group to make sure none of these are still actively used. For example, I think Hugo still uses the genfit track fitter (but not the propagator, he can correct me if I'm wrong).

pinkenburg:
Point taken - just want to see if this goes through. None of the removed code was in the Makefile, so we haven't been building any of this. If there is code still in use - we should add it back to the Makefile.am so it gets compiled and tested

osbornjd:
Okay fair - that seems likely then that nobody is using it 🤣

pinkenburg:
If it turns out I removed too many sources - we can put them back

PR #1805, opened 2023-01-22
shifting epd sector phi shift

eumaka:
PR rotates epd sectors so that the first sector starts a phi = 0. See attached image for details- The numbers [0-11] are the sector ids.
epdphishift_PR

### DIFF
--- a/simulation/g4simulation/g4epd/PHG4EPDDetector.cc
+++ b/simulation/g4simulation/g4epd/PHG4EPDDetector.cc
@@ -61,8 +61,20 @@ void PHG4EPDDetector::ConstructMe(G4LogicalVolume* world)
     for (int32_t k = 0; k < nsectors; ++k)
     {
       G4RotationMatrix* rotate = new G4RotationMatrix();
+      
+      //shift epd sectors so that first sector starts at phi = 0
+      double phi_shift = (k + 3) * 2 * M_PI / nsectors;
 
-      rotate->rotateZ(k * 2 * M_PI / nsectors);
+      if(phi_shift >= (2.0 * M_PI))
+      {
+         phi_shift -= (2.0 * M_PI);
+      }
+      else if (phi_shift < 0.0)
+      {
+         phi_shift += (2.0 * M_PI);
+      }
+
+      rotate->rotateZ(phi_shift);
 
       m_volumes.emplace(
           new G4PVPlacement( rotate, positive, volume, label, world, false, 2 * k + 0, OverlapCheck()),


sPHENIX-GitHub-l:
Dear Ejiro:

Thanks for the nice plots!

Where is the viewpoint for these plots? Is it z = 0, or |z| > 316 cm?

If it’s z=0, then I would think you would want sector number increasing in the direction of increasing phi, that is, when moving from the +x-axis towards the +y-axis. To be explicit, for the z=316 cm north “after” plot, this would mean the sector numbered as 11 would become 1, that is, sector numbers increasing in CCW direction from the z=0 viewpoint looking towards +z (standard 3D cylindrical coordinate system). 

This is tricky because of mirror symmetry about the x-y plane. Again, if these plots are made with viewpoint at z=0 , then looking at the south wheel, *if* you want sector 0 to be the mirror image of sector 0 in the north EPD, then it would sit where sector 7 currently appears in the south “after” plot, and from this viewpoint sector number would be increasing in the CW direction. 

None of this matters if you have thoroughly debugged mapping functions from EPD coordinates to global coordinates. One thing I am thinking about is working on the real detector and cabling. Again, if these plots are from the z=0 viewpoint, Sector 0 in the north is on the east side of the detector and the west side of the detector for the south EPD. It’s not my call, but I wanted you to be aware of this.

I had not thought through these mirror image problems before. I find it useful to think of the OHCal sector numbering scheme, where of course a given sector extends from +z to -z, no physical ambiguity at all! You might ask what the TPC sector numbering conventions are for the pad planes, where they have a N/S similar split similar to the EPD’s. The question then becomes whether it makes more sense to adapt the TPC convention or the HCal convention (assuming there is a difference at all).

Best regards,

Bill 

—————————————————
W.A. Zajc
I.I. Rabi Professor of Physics
Columbia University
New York, NY 10027

https://blogs.cuit.columbia.edu/waz1/
—————————————————

pinkenburg:
Dear Bill,

I don't want to pre-empt Ejiro from answering but let me give you the background for the decision on the epd phi mapping (which is probably my idea). It does differ from STAR which uses z=0 to assign phi while we use the sPHENIX coordinate system. Doing it our way just makes the geometry handling simpler (no distinction for north/south needed to return phi for a given tile). Lining it up with the calorimeter sector 0 definition then makes this more consistent (avoiding those "why does the epd sector 0 points to 90degrees while the calorimeters sector 0 sits at 0 degrees?").

In the end all of this is hidden from most of our code and the users - they will see some tile id (where we encoded the arm/phi/ring, but you need to look at it in hex to see this encoding) and ask the geometry object where this tile is located. But we do have to make sure this is done correctly which I would say Ejiros plots show. This does need to be followed through and verified further downstream but that code is in flux (with the epd reco code itself still missing)

Best regards,
Chris

eumaka:
Hi Bill, Chris,

Thanks Bill for your point and this makes a whole lot of sense. These are
from the viewpoint of z = 0.

I can make a small change:
The north wheel can be mapped keeping sector zero where it is in the after
plot with increment in sector IDs going in the counterclockwise direction
(as you are suggesting). This also meets Chris' and the calorimeter
convention of having the first sector starting at phi = 0.

However, the southwheel sector mapping will be identical to the north's
since we want to keep tiles with the same id at the same phi (agnostic to
the wheel z location).

As Chris mentioned, we are working on a geometry code that will provide
information on where the tiles are located (what you referred to as the EPD
coordinates mapping).

Best regards,
Ejiro

PR #1807, opened 2023-01-24
epd sector re-arrangement

eumaka:
PR rotates epd sectors counter clockwise with sector 0 starting at phi=0


eumaka:
The PDF explains this pictorially 
[epd_sector_phishift.pdf](https://github.com/sPHENIX-Collaboration/coresoftware/files/10490715/epd_sector_phishift.pdf)

PR #1810, opened 2023-01-25
YCM possible trick to remove memory leak and make valgrind happy

ycorrales:
This PR is just a test to remove valgrind complains in g4micromegas
## Types of changes

## TODOs (if applicable)

### DIFF
--- a/simulation/g4simulation/g4micromegas/PHG4MicromegasDetector.cc
+++ b/simulation/g4simulation/g4micromegas/PHG4MicromegasDetector.cc
@@ -100,7 +100,7 @@ void PHG4MicromegasDetector::Print(const std::string &what) const
 //_______________________________________________________________
 void PHG4MicromegasDetector::setup_tiles()
 {
-  
+
   // TODO: replace with more realistic description from latest engineering drawings
   m_tiles.clear();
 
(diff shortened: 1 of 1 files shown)


pinkenburg:
handing down the address of a local G4RotationMatrix does not work. It goes out of scope and the rotation during the actual detector construction is done using garbage values

ycorrales:
Very strange, we I used some similar approach for assembly volume and it works. Actually a possible solution is using a G4Tranform3D as in https://github.com/sPHENIX-Collaboration/coresoftware/blob/b51c2947ac504b1d419524c0fb36d0eeda269401/simulation/g4simulation/g4mvtx/PHG4MvtxDetector.cc#L197

hupereir:
Yasser, 
using G4Transform3D would fix it indeed. What G4VPlacement does in that case is that it creates the rotation internally, and takes care of deleting it, when the placement gets out of scope, I think.
 The only drawback is that one creates a new rotation for each placement, instead of having the possibility to reuse the same rotation.
This should not be a problem in our case. 
I'll switch to this construct and we will see.

hupereir:
see e.g. https://apc.u-paris.fr/~franco/g4doxy/html/classG4PVPlacement.html

ycorrales:
yes, you are right

PR #1818, opened 2023-02-01
Update PHG4EPDDetector.cc

eumaka:
This PR assigns arm = 0 as the negative wheel, i.e. the south wheel. And arm = 1 as the positive wheel, i.e. the north wheel (see attached pdf).

The reason is to match the sim to the channel mapping scheme of TowerInfoContainer: https://github.com/sPHENIX-Collaboration/coresoftware/blob/master/offline/packages/CaloBase/TowerInfoContainerv1.cc#L101

where that mapping scheme decodes arm = 0 as the south and thus, channels ranging from 0-371; and arm = 1 as the north with channels ranging from 372-743.

[epd_arm_index.pdf](https://github.com/sPHENIX-Collaboration/coresoftware/files/10563497/epd_arm_index.pdf)


pinkenburg:
The valgrind leak is from PHG4Reco.cc where the line numbers have shifted. Will be surpressed

PR #1834, opened 2023-02-17
Fixed phi and z bin number and added sampa_tbias

jizhongling:
## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work for users)
- [ ] Requiring change in macros repository (Please provide links to the macros pull request in the last section)
- [x] I am a member of [GitHub organization of sPHENIX Collaboration](https://github.com/orgs/sPHENIX-Collaboration/people), EIC, or ECCE (contact Chris Pinkenburg to join)

## What kind of change does this PR introduce? (Bug fix, feature, ...)
Function cluster_and_reset in TpcClusterBuilder copied from TpcClusterizer but missing:
if(side == 0) clusz = -clusz;
clust = clust + m_sampa_tbias;
Also, iphi and iz wrongly used a relative bin number.
This fixed all the issues.

PR #1850, opened 2023-03-01
Update the phi and z clustering locations

david-stewart:

## Types of changes
- [ ] I am a member of [GitHub organization of sPHENIX Collaboration](https://github.com/orgs/sPHENIX-Collaboration/people), EIC, or ECCE (contact Chris Pinkenburg to join)

This request undoes a few lines modified in [PR 1834](https://github.com/sPHENIX-Collaboration/coresoftware/pull/1834/files) which inadvertantly place the clusters reconstruction out of bounds and therefore don't make truth clusters for many tracks. I think this re-correction is the proper way to do things but will cast a weathered eye over it tomorrow, and would appreciate any expert opinion.

Essentially what happend in PR 1834 is the offset corrections to phi and time put the clusters out of bounds for the number of possible bins (the checks at TpcClusterBuilder.cc lines 105 and 109) and therefore removes many clusters.

I left the sampa bias correction from PR 1834 in, as well as the flipping of the cluster Z location for the other side of the TPC (which I think is correct). The logic all follows from TpcClusterizer.cc, which is the work of many experts, but the logic there is not perhaps difficult, but at least highly nested.

## TODOs (if applicable)

### DIFF
--- a/simulation/g4simulation/g4tpc/PHG4TpcElectronDrift.cc
+++ b/simulation/g4simulation/g4tpc/PHG4TpcElectronDrift.cc
@@ -583,7 +583,7 @@ int PHG4TpcElectronDrift::process_event(PHCompositeNode *topNode)
       const int sector = TpcDefs::getSectorId(node_hitsetkey);
       const int side = TpcDefs::getSide(node_hitsetkey);
 
-      if (Verbosity() > 2)
+      if (Verbosity() > 8)
         std::cout << " hitsetkey " << node_hitsetkey << " layer " << layer << " sector " << sector << " side " << side << std::endl;
       // get all of the hits from the single hitset
       TrkrHitSet::ConstRange single_hit_range = single_hitset_iter->second->getHits();
@@ -679,7 +679,7 @@ int PHG4TpcElectronDrift::process_event(PHCompositeNode *topNode)
 
   truth_clusterer->cluster_and_reset(/*argument is if to reset hitsetkey as well*/ true);
 
-  if (Verbosity() > 2)
+  if (Verbosity() > 10)
   {
     std::cout << "From PHG4TpcElectronDrift: hitsetcontainer printout at end:" << std::endl;
     // We want all hitsets for the Tpc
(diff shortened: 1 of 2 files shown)


jizhongling:
I think we should not make these changes. [L100](https://github.com/sPHENIX-Collaboration/coresoftware/blob/7bd658fae08dedc2544331bbcfd2afe293857c8c/simulation/g4simulation/g4tpc/TpcClusterBuilder.cc#L100) of TPCClusterBuilder.cc should not subtract the phioffset, as we can see from [L319](https://github.com/sPHENIX-Collaboration/coresoftware/blob/0c815d572a48505e0031bbb30a9c7d2d2ebe57bf/offline/packages/tpc/TpcClusterizer.cc#L319) from the TPCClusterizer.cc. The iphi and iz should use the absolute bin numbers instead of the relative bin numbers.

david-stewart:
> We should not revert the changes made in [PR 1834](https://github.com/sPHENIX-Collaboration/coresoftware/pull/1834/files). [L100](https://github.com/sPHENIX-Collaboration/coresoftware/blob/7bd658fae08dedc2544331bbcfd2afe293857c8c/simulation/g4simulation/g4tpc/TpcClusterBuilder.cc#L100) of TPCClusterBuilder.cc should not subtract the phioffset, as we can see from [L319](https://github.com/sPHENIX-Collaboration/coresoftware/blob/0c815d572a48505e0031bbb30a9c7d2d2ebe57bf/offline/packages/tpc/TpcClusterizer.cc#L319) of the TPCClusterizer.cc. The iphi and it should use the absolute bin numbers instead of the relative bin numbers.
> 
> However, as @david-stewart mentioned, the checks at TpcClusterBuilder.cc lines [105 and 109](https://github.com/sPHENIX-Collaboration/coresoftware/blob/7bd658fae08dedc2544331bbcfd2afe293857c8c/simulation/g4simulation/g4tpc/TpcClusterBuilder.cc#L105) indeed caused the problem. We should do the checks for the relative bin number and add back the phioffset and toffset after the checks.
> 
> So my suggestions are simply do the following at [L113](https://github.com/sPHENIX-Collaboration/coresoftware/blob/7bd658fae08dedc2544331bbcfd2afe293857c8c/simulation/g4simulation/g4tpc/TpcClusterBuilder.cc#L113): iphi += phioffset; it += toffset;

I concur. The update to the PR has this update.

This also includes minor update to mirror changes in the TpcClusterizer.cc in which the mean phi is found first and then binned to the average phi bin, instead of binning in phi first and then finding the bin. (This change in TpcClusterizer happened late January). The change in calculation result is small, but it is good to be totally consistent between the truth clusters and the reconstructed clusters.

PR #1856, opened 2023-03-06
cluster error parametrization update

bogui56:

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/QA/modules/QAG4SimulationIntt.cc
+++ b/offline/QA/modules/QAG4SimulationIntt.cc
@@ -263,16 +263,16 @@ void QAG4SimulationIntt::evaluate_clusters()
 
       double phi_error = 0;
       double z_error = 0;
-      if (m_cluster_version == 3)
-      {
-        phi_error = cluster->getRPhiError() / r_cluster;
-        z_error = cluster->getZError();
-      }
-      else
+      if (m_cluster_version == 4)
       {
         auto para_errors = _ClusErrPara.get_si_cluster_error(cluster, key);
         phi_error = sqrt(para_errors.first) / r_cluster;
         z_error = sqrt(para_errors.second);
+      }
+      else
+      {
+        phi_error = cluster->getRPhiError() / r_cluster;
+        z_error = cluster->getZError();
       }
 
       // find associated g4hits
--- a/offline/QA/modules/QAG4SimulationMicromegas.cc
+++ b/offline/QA/modules/QAG4SimulationMicromegas.cc
@@ -469,12 +469,7 @@ void QAG4SimulationMicromegas::evaluate_clusters()
       // get relevant cluster information
       double rphi_error = 0;
       double z_error = 0;
-      if (m_cluster_version == 3)
-      {
-        rphi_error = cluster->getRPhiError();
-        z_error = cluster->getZError();
-      }
-      else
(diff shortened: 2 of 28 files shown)


pinkenburg:
We cannot just add or remove data members in those version classes. This does affect the readback. Is TrackEvaluationContainerv1.h being saved in past DSTs? If so please update the version to v2

pinkenburg:
Then we are good. I really want to avoid a repeat of the disaster we had in mdc1 where a change in the tracking object completely messed up the analysis of the data

osbornjd:
The distributions are obviously nontrivially changed but the momentum resolution degrades slightly. I assume this is because the pulls on some of the cluster distributions have shrunk, e.g. the TPC cluster z pulls are significantly narrower than before.

osbornjd:
Also in reference to Chris' comment, I think the changes in the `TrkrClusterv4` will cause DST readback issues since we are using this cluster version right now. While the evaluation container is probably not a big deal since this is primarily only used by tracking experts, the cluster object change might wreak havoc

sPHENIX-GitHub-l:
I agree with Joe about the TrkrCluster changes. If what is stored changes, it has to be versioned.

Tony

bogui56:
Hi, 

regardons the storage nothing has changed, we still have a unsigned short int data member for the 
charge information. Before we put the total cluster charge in the data member, now I put the max adc in the high 8 bits and the total charge in the lower 8 bits. 
Reading back old clusters would work, but obviously the max adc would return zero.

If that warrants a new cluster version, that can be arranged. 

Does someone remember where the reference comes from? Is this still from cluster version 3 times? When did we last update this?

Cheers

  Christof

pinkenburg:
the commit adds a static unsigned int. We would have to see what the root streamer makes out of this - I would not be surprised if it stores it and then tells you the objects don't match. It is used in one place - if it is not used outside of this source I would either just hardcode this as 8 or use a #define (where one has to be careful that this is set in every source which includes this file

bogui56:
Ok, there was not supposed to be a new data member, let me check.

osbornjd:
@bogui56 just double checking but I think we decided to merge this and then switch to v5 at the macro level, correct? The parameterization at least looks like it has returned to the previous performance.

hupereir:
Hi, 
Maybe I missed something. Why does this PR come with a new version of the TrkrClusterContainer (namely TrkrClusterContainerV5) ? As far as I could tell, it is _identical_ to TrkrClusterContainerv4 except for the class name. What is the actual difference that justifies scheme evolution ? 
If there is none, I would recommand removing the class and stick to v4. 
(note that this is orthogonal to having a new version for TrkrCluster)

hupereir:
In fact, checking further, the class TrkrClusterContainerV5 is not used. All existing clusterizers are still creating instances of TrkrClusterContainerv4
So unless there is an objection, I would like to prepare a separate PR that removes this unused class. It just adds confusion ...

bogui56:
You are right, we can remove this again.

bogui56:
Yes, thats the plan. The macro pull request to move to v5 as default is out. 
The paramterization was off before because the TpcClusterizer update to write the cluster max adc
was missing from the pull request by accident. A big part of the new parameterization is the 
max arc dépendance, so now it looks mach better. 

Cheers

   Christof

PR #1862, opened 2023-03-09
Add NN clustering

jizhongling:
This PR is an implementation of NN clustering in TPC by using the libtorch.

## Types of changes

This PR includes the NN clustering in TPC by using the libtorch.

This PR inputs the TPC hits into a neural network and outputs the position of the corresponding cluster.

## TODOs (if applicable)
Improve the neural network by further training.


offline/packages/tpc/TpcClusterizer.cc
```diff
@@ -797,6 +876,48 @@ int TpcClusterizer::InitRun(PHCompositeNode *topNode)
...
+    PHIODataNode<PHObject> *TrainingHitsContainerNode =
+        new PHIODataNode<PHObject>(training_container, "TRAINING_HITSET", "PHObject");
+    DetNode->addNode(TrainingHitsContainerNode);
+  }
+
+  use_nn = _use_nn;
+  if(use_nn)
+  {
+    //const char *offline_main = std::getenv("OFFLINE_MAIN");
+    const char *offline_main = std::getenv("MYINSTALL");
```
osbornjd, line 901: It looks like the default macro crashes here, which the `MYINSTALL` environment won't be present in our daily build anyway. Do we need to have some additional libraries built with our daily build? It looks like maybe you had to install pytorch yourself, in which case we should get it into the daily build.

pinkenburg: pytorch is installed and copied to OFFLINE_MAIN in every build

osbornjd: Okay thanks for the confirmation, so this just needs to lookup offline main in the usual way

jizhongling: Sorry that I forgot to change MYINSTALL to OFFLINE_MAIN. In fact, I have done this before, which is the commented code. Hopefully, this can solve the conflicts.

jizhongling:
Sorry for my late response since I am on shift this week. I have fixed the cpp check issues. For the `QA-tracking-low-occupancy`, it seems that there is no `ffamodules/XploadInterface.h` file. I will wait for the `$OFFLINE_MAIN` to include the file or I will compile it for myself.

pinkenburg:
ffamodules/XploadInterface.h was removed months ago when we switched our cdb client to the current implementation. That sounds like an outdated macro or outdated code. This include will not come back.

jizhongling:
After getting help from @pinkenburg , I merged the branch `QA-tracking-low-occupancy` with `master` and now the `Fun4All_G4_sPHENIX.C` runs smoothly without any error with my compiled code. I pushed again to see what Jenkins say now.

osbornjd:
Okay great, the QA automatically merges master into the QA branches so they can be a little out of date sometimes. We should update them so that these CDB errors don't come up for others who want to test their code with the QA branches

osbornjd:
It looks like there is still some thread creation issue, were you able to run this with your locally compiled version?
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-tracking-low-occupancy-qa/3818/console

jizhongling:
Locally, I can run with `QA-tracking-low-occupancy` smoothly. There is no error displayed. I am current examine why it causes thread error in the Jenkins.

osbornjd:
It also fails nearly immediately to create any threads... @blackcathj do you have any ideas as to why Jenkins can't create the threads for this PR but @jizhongling is able to run it locally? Oddly the default macro also runs, which uses the tpc clusterizer in multi-threaded mode by default as well (and the NN clustering in this PR is set to false by default so it should not change any default functionality).

pinkenburg:
valgrind complains about 
 if (flagDoSD)
in offline/packages/CaloReco/BEmcRec.cc:253
I would guess flagDoSD is not initialized (or set to an initialized value)

osbornjd:
That file is completely unrelated to this PR, is there a chance something else snuck in a different PR or a suppression file got changed? The good news is that the tracking QA actually runs now so that means the pytorch issue blowing up the jenkins server is resolved, so we can actually merge this soon.

pinkenburg:
Sorry - took the lowest hanging fruit. Then it's memory leaks:
offline/packages/tpc/TpcClusterizer.cc:354
  auto training_hits = new TrainingHits;

pinkenburg:
I checked the  flagDoSD, I have no idea why valgrind flagged this - this flag is a local variable declared a few lines earlier and it is initialized to true and set to false in an if statement

pinkenburg:
We should put a flag around creating and filling those training sets. It's inside the tpc clustering, that's got to have an effect on the overall time the clustering jobs take

jizhongling:
Thanks for finding the problem. I will solve the training_hits memory leak and put a flag to it.

osbornjd:
@pinkenburg these scan build warnings should always be caught since they are behind a flag to run the NN. Is it safe to add these to the suppression file?

pinkenburg:
unlike scan build does not have a suppression file, one has to create a new reference for jenkins. A quick hack to make scan build happy is just adding assert(training_hits) before it is being used in the places it complains about. That's the easiest (and it catches a null pointer if there ever is one)

jizhongling:
Thanks for the suggestion. I will add assert(training_hits) before dereference it. However, I also saw the same warnings in others' code. Hope those alone will not make it fail.

pinkenburg:
close...
The assert has to come before this line:
training_hits->v_adc[(iphi_diff+nd)*(2*nd+1)+(it_diff+nd)] = adc;

osbornjd:
Looks like also need an assert before this line too, l535

https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/Build-ScanBuild-gcc12/1414//clang-analyzer/new/source.8970bb36-41c8-4ff9-8926-9b286cbc8d8f/#535

jizhongling:
Thanks a lot. Now I check training_hits to be non-null everywhere. I should be careful about this at the beginning.

osbornjd:
These scan build errors are fall out from the Acts update - we can merge this now

PR #1865, opened 2023-03-13
Introducing vectorized TPC `TrkrHitSet`

blackcathj:

## Types of changes

As [discussed with the tracking group ](https://indico.bnl.gov/event/18555/#1-tpc-signal-processing), there are multiple problems of using `TrkrHitSet` to represent TPC data, which uses `stl::map` of `TrkrHit *` for sparse encoding. It may fit silicon tracker well, but for TPC, it is dramatically wasteful for storage (uses 10x memory to store 2-byte ADC number) and slow for operation (`log(n)` for insertion/access, and calls allocation/deallocation). It also potentially leads to memory fragmentation with millions `new`/`delete` of `TrkrHitv2` per event.

This pull request introduce a concept to use static objects to store TPC data: 
* at `TrkrHitSet` level, `TrkrHitSetTpc` and `TrkrHitSetTpcv1` is introduced to use 2D nested `stl::vector` to store a non-zero suppressed TPC time frame, avoiding the use of `TrkrHit` for TPC
* At `TrkrHitSetContainer` level,  `TrkrHitSetContainerv2` use a `TClonesArray` backend for `TrkrHitSet` storage, which optimize for ROOT IO and AVOIDs allocation/ctor/dtor calls for `TrkrHitSetTpc`. The `stl::map` based assess interface is preserved with automatic map rebuild at DST readback.

This approach is aimed to deliver multiple advantage over the current treatment of the TPC data: 
* Reflect what TPC data attempt to represent: pad-time 2D array of ADC time-frame per layer of pads
* Fast hit access with array at `O(1)` speed
* 82MB fixed size memory for AuAu TPC hit data
* No map/TObject/allocator overhead associated with `TrkrHit`
* No memory fragmentation with repeated `new`/`delete` on heap
* It is also [easier to use with digitizer code](https://github.com/sPHENIX-Collaboration/coresoftware/commit/da9ddfea9e30672696b93442af1838c0365c06a5#diff-9d26dd1e18b714726a0505dc6bee180009859671e25ebec17153096823b633b1R647) and better fit clusterizer (which decode hits into 2D arrays anyway)

## TODOs (if applicable)

**Discuss, test, debug and optimize**

In particular, integration with the digitizer and clusterizer code. We potentially would benifit from a redefine of pad and tbin for `TrkrHitSetTpc` that represent the relative position within the `TrkrHitSet` rather than global index. @adfrawley @bogui56 

Also the TPC sector test data, which is represented as time frames, can be easily ported to this structure: 
https://github.com/sPHENIX-Collaboration/calibrations/pull/117

### DIFF
--- a/offline/packages/trackbase/TrkrHitSetContainerv2.cc
+++ b/offline/packages/trackbase/TrkrHitSetContainerv2.cc
@@ -1,0 +1,162 @@
+/**
+ * @file trackbase/TrkrHitSetContainerv2.cc
+ * @author D. McGlinchey, H. PEREIRA DA COSTA
+ * @date June 2018
+ * @brief Implementation for TrkrHitSetContainerv2
+ */
+#include "TrkrHitSetContainerv2.h"
+
+#include "TrkrDefs.h"
+#include "TrkrHitSetv1.h"
+
+#include <TClass.h>
+
+#include <cassert>
+#include <cstdlib>
+
+TrkrHitSetContainerv2::
+    TrkrHitSetContainerv2(const std::string& hitsetclass)
+  : m_hitArray(hitsetclass.c_str())
+{
+}
+
+void TrkrHitSetContainerv2::Reset()
+{
+  m_hitmap.clear();
+  //! fast clear without calling destructor
+  m_hitArray.Clear("C");
+}
+
+void TrkrHitSetContainerv2::identify(std::ostream& os) const
+{
+  syncMapArray();
+
+  os << "TrkrHitSetContainerv2 with class "
+     << m_hitArray.GetClass()->GetName()
+     << ": Number of hits: " << size() << " index map size = " << m_hitmap.size() << std::endl;
+  ConstIterator iter;
+  for (const auto& pair : m_hitmap)
+  {
+    int layer = TrkrDefs::getLayer(pair.first);
+    os << "hitsetkey " << pair.first << " layer " << layer << std::endl;
+    pair.second->identify();
+  }
+  return;
+}
+
+TrkrHitSetContainerv2::ConstIterator
+TrkrHitSetContainerv2::addHitSet(TrkrHitSet* newhit)
+{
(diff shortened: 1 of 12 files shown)


offline/packages/trackbase/TrkrHitSetContainerv2.cc
```diff
@@ -0,0 +1,162 @@
...
+  return addHitSetSpecifyKey(newhit->getHitSetKey(), newhit);
+}
+
+TrkrHitSetContainerv2::ConstIterator
+TrkrHitSetContainerv2::addHitSetSpecifyKey(const TrkrDefs::hitsetkey key, TrkrHitSet* newhit)
+{
+  std::cout << __PRETTY_FUNCTION__
+            << " : deprecated. Use findOrAddHitSet()." << std::endl;
+
+  exit(1);
```
osbornjd, line 61: I guess this won't have any effect on the silicon hit set containers because they will continue to use v1, if my understanding is correct.

offline/packages/trackbase/TrkrHitSetContainerv2.cc
```diff
@@ -0,0 +1,162 @@
...
+}
+
+void TrkrHitSetContainerv2::syncMapArray(void) const
+{
+  if (m_hitmap.size() == (size_t) m_hitArray.GetSize()) return;
+
+  if (m_hitmap.size() > 0)
+  {
+    std::cout
+        << __PRETTY_FUNCTION__ << " Error: m_hitmap and m_hitArray get out of sync, which should not happen unless DST readback. "
```
osbornjd, line 150: Is this an error that should lead us to stop processing? This function re-syncs the hitset values but I'm wondering if this case arises if there is some more global problem that may be hard to then debug

osbornjd (approved):
Looks good to me other than a few minor comments/questions, I think we should merge this and test it out

PR #1902, opened 2023-03-22
Eval tree

david-stewart:

## Types of changes
Apply PruneHits to the Mvtx hits (and therefore make too many Mvtx clusters)
- add module g4eval/FillTruthRecoMatchTree, which writes the phg4 to svtx track matches, along with clusters that do and don't match to an output TTree

## TODOs (if applicable)

### DIFF
--- a/offline/packages/mvtx/MvtxClusterizer.cc
+++ b/offline/packages/mvtx/MvtxClusterizer.cc
@@ -210,6 +210,7 @@ int MvtxClusterizer::InitRun(PHCompositeNode *topNode)
 
 int MvtxClusterizer::process_event(PHCompositeNode *topNode)
 {
+  cout << " FIXME y0 : new event in MvtxClusterizer " << endl;
   // get node containing the digitized hits
   if(!do_read_raw){
     m_hits = findNode::getClass<TrkrHitSetContainer>(topNode, "TRKR_HITSET");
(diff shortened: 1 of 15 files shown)


offline/packages/mvtx/MvtxClusterizer.cc
```diff
@@ -344,16 +344,15 @@ void MvtxClusterizer::ClusterMvtx(PHCompositeNode *topNode)
...
+  int total_clusters = 0;
       for (set<int>::iterator clusiter = cluster_ids.begin(); clusiter != cluster_ids.end(); ++clusiter)
 	{
 	  int clusid = *clusiter;
 	  auto clusrange = clusters.equal_range(clusid);
 	  
 	  if (Verbosity() > 2) cout << "Filling cluster id " << clusid << " of " << std::distance(cluster_ids.begin(),clusiter )<< endl;
 	  
-	  // make the cluster directly in the node tree
+    ++total_clusters;
```
osbornjd, line 355: This looks unnecessary, and in any case if you need the total clusters you can always request the size of the container that stores them.

osbornjd (approved):
Just one small comment that looks like it is leftover from debugging, otherwise this looks good

osbornjd:
@david-stewart is this ready to be merged?

david-stewart:
ok @osbornjd 
I have merged the parts of the Mvtx and Intt that made sense from the PR that we decided to not use. The Intt and Mvtx are using the unified TruthClusterizerBase, and both digitize and cluster. I probably should also move the truth cluster making for the TPC to a derived class of TruthClusterizerBase, but it is functionally working and will change with the new HitSet version anyway, so I have left it alone for now.
I've tested the results, and get hits in Mvtx Intt and TPC for Truth and Reco, and hit matching. I'll work through Jenkins again and then it should be good to go.

osbornjd (commented):
Okay it looks like all the CI passed from the macros+coresoftware build. I still have one more question though, maybe I'm missing something here but why are the truth clusterizers still non-Fun4All inheriting classes? They look like the follow a similar structure of needing an `init_run` type function and then they do something for each event, so I'm just trying to understand the design choice of making them helper classes instead of inheriting from SubsysReco

PR #1905, opened 2023-03-23
add proper dynamic cast to use cluster v5

bogui56:

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackreco/PHActsSiliconSeeding.cc
+++ b/offline/packages/trackreco/PHActsSiliconSeeding.cc
@@ -24,7 +24,6 @@
 #include <trackbase_historic/TrackSeed.h>
 #include <trackbase_historic/TrackSeed_v1.h>
 #include <trackbase/TrkrCluster.h>            
-#include <trackbase/TrkrClusterv5.h>            
 #include <trackbase/TrkrClusterContainer.h>
 #include <trackbase/TrkrDefs.h>
 #include <trackbase/TrkrClusterIterationMapv1.h>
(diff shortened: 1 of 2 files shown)


hupereir:
I am confused. Why is it necessary at all  ? getRPhiError and getZError are virtual methods from the base class. There should be no need to cast to the latest version. Or did the signature change for ClusterV5 ? 
This code will crash if one runs the tracking on a DST with clustersV4.
IMHO it should be better to investigate why the called method does not follow inheritance (assuming this is really the issue).

bogui56:
I am confused too, but in the latest build the getRPhiError just printed the error message you get from calling the base class method
instead of returning the implemented method from clusterv5.

If you know another way around this let me know.

Cheers

   Christof

osbornjd:
Tony found instances where there was not an if statement in the code for clusterv5 and thus the clusterv4 was being called, leading to seg faults or other undefined behavior. I think as a general statement we need to consider how this is going to work long term because it is not feasible to have a block of code like
```
if(clusterv3)
{ do something }
else if(clusterv4)
{ do something else}
else if (clusterv5)
{ do a third thing }
else if(clusterv6)
{ do a fourth thing}
...
```

in every module that needs the clusters, which is a lot of them. This just isn't scalable in the long term

bogui56:
Hi Joe, 

I agree. in the beginning this was supposed to be a tempory solution to try the perforrmance with a cluster object that has a reduced set 
of data members. Maybe its time to think about a more permanent solution. Any suggestions are more than welcome. 

Regarding the pull request adding the dynamic cast I am still confused. The situation of the getRPhiError throwing the error message 
from the base class instead of returning the cluster error from clusterv5 looks to have been temporary. The default behaviour has changed again 
today. No clue what caused this. I will withdraw the pull request.

cheers

   Christof

hupereir:
> Tony found instances where there was not an if statement in the code for clusterv5 and thus the clusterv4 was being called, leading to seg faults or other undefined behavior. I think as a general statement we need to consider how this is going to work long term because it is not feasible to have a block of code like
> 
> ```
> if(clusterv3)
> { do something }
> else if(clusterv4)
> { do something else}
> else if (clusterv5)
> { do a third thing }
> else if(clusterv6)
> { do a fourth thing}
> ...
> ```
> 
> in every module that needs the clusters, which is a lot of them. This just isn't scalable in the long term

In principle that's exactly what virtual functions are supposed to do for you, so maybe we should revisit where these switches happen and encapsulate that into virtual functions from TrkrCluster with different implementations for each function. I might be able to have a look next week.

bogui56:
For the most part we can do this with virtual functions, except for v4 where we simply dont store the error is the cluster struct. 
We could make v4 return a nan with getRPhiError and catch it and replace it with the ClusterErrorPara. But that is also not a clean solution.

Lets discuss this in the next tracking or simsoft meeting

Cheers

  Christof

PR #1910, opened 2023-03-25
Truth clusterer

david-stewart:
Add class g4tracking/TruthClusterizerBase, impl for Mvtx, Intt, and TPC

Add hooks to pruners, digitizers, and clusterers to cluster PHG4Hits
for each individual embedded track with the same code that does all of
the clustering.

Because clustering the truth tracks internally uses pointers to the
downstream modules that do the clustering, it must be initialized in the
driving Fun4All macro after all necess│ary modules have already been
added to the Fun4AllServer.

An updated Fun4All macro will be added. If put in the main Fun4All macro,
then the logic looks something like this:

```.C
  //--------------
  // SVTX tracking
  //--------------
  if(Enable::TRACKING_TRACK)
    {
      TrackingInit();
    }
  if (Enable::MVTX_CLUSTER) {
    Mvtx_Clustering();
    PHG4MvtxHitReco* hitreco = (PHG4MvtxHitReco*) se->getSubsysReco("MVTX");

    PHG4MvtxDigitizer* digitizer   = (PHG4MvtxDigitizer*) se->getSubsysReco("PHG4MvtxDigitizer");
    MvtxHitPruner*     pruner      = (MvtxHitPruner*)     se->getSubsysReco("MvtxHitPruner");
    MvtxClusterizer*   clusterizer = (MvtxClusterizer*)   se->getSubsysReco("MvtxClusterizer");
    hitreco->SetTruthClusterizer(digitizer, pruner, clusterizer, verbosity);
  }

  if (Enable::INTT_CLUSTER) {
     Intt_Clustering();

    PHG4InttHitReco* hitreco = (PHG4InttHitReco*) se->getSubsysReco("PHG4InttHitReco");

    PHG4InttDigitizer* digitizer   = (PHG4InttDigitizer*) se->getSubsysReco("PHG4InttDigitizer");
    InttClusterizer*   clusterizer = (InttClusterizer*)   se->getSubsysReco("InttClusterizer");
    hitreco->SetTruthClusterizer(digitizer, clusterizer, verbosity);
  }

  if (Enable::TPC_CLUSTER) {
    TPC_Clustering();

    PHG4TpcElectronDrift* tpc_edrift = (PHG4TpcElectronDrift*) se->getSubsysReco("PHG4TpcElectronDrift");

    PHG4TpcDigitizer* digitizer   = (PHG4TpcDigitizer*) se->getSubsysReco("PHG4TpcDigitizer");
    TpcClusterizer*   clusterizer = (TpcClusterizer*)   se->getSubsysReco("TpcClusterizer");
     tpc_edrift->SetTruthClusterizer(digitizer, clusterizer, verbosity);
  }
```

## Types of changes

## TODOs (if applicable)


david-stewart:
This PR greatly simplifies the clustering of truth track clusters. Instead of duplicating the code, it puts call methods into the actual methods that do the digitizing, cleaning, and clustering for the regular ("Svtx") tracks. It add a base class `g4tracking/TruthClusterizerBase`. This base class is implemented by a derived class for the MVTX, INTT, and TPC (`g4mvtx/PHG4MvtxTruthClusterizer`, `g4intt/PHG4InttTruthClusterizer`,  and `g4tpc/PHG4TpcTruthClusterizer`). Each implementation contains pointers to the classes which deal with the `TrkrHitSetContainer` for each system.

This is much better in terms of reusing code and maintainability. It *does* require linking a library from `packages/mvtx` to `g4mvtx`, and similar for the INTT and TPC. For purposes of library size, it would be possible to move each implementation and library to `g4tracking`, as `PHG4{Mvtx,Intt,Tpc}TruthClusterizer` is accessed in each case by a pointer in `g4{Mvtx,Intt}/PHG4{Mvtx,Int}HitReco` and `g4tpc/PHG4TpcElectronDrift`+`g4tpc/PHG4TpcPadPlaneReadout`, which by default is simply a null pointer which, if not implemented from Fun4All doesn't need the accompanying library loaded at run time.

osbornjd:
I am confused why there are so many changes to the reconstructed clusterizer modules in this PR, and in fact reading through the code I am also confused as to why there is already a flag in the MvtxClusterizer for differences in algorithms between a `TRKR_HITSET` and a `TRKR_RAWHITSET`. What are the changes necessary in the reconstructed clusterizers that have already been introduced and are introduced in this PR? Why are they necessary? Your macro snippet implies that the truth hit reconstruction requires the reconstructed clustering module? This seems really circular, or am I missing something?

david-stewart:
Good questions.
First, the `TRKR_HITSET` and `TRKR_RAWHITSET` switch is preexisting. It has nothing to do with truth track clusters, and this PR just leaves them alone to their existing functions.

The logic for this PR comes from the logic in the code for generating `TrkrCluster`'s. For Svtx tracks it goes like this (for non-raw hitsets):
* For the MVTX: 
  - `g4mvtx/PHG4MvtxHitReco` takes all PHG4 hits, and writes `TrkrHitSetContainer`s to the node tree
  - `g4mvtx/PHG4MvtxDigitizer` digitizes `TrkrHitSet`'s in the `TrkrHitSetContainer`
  - `mvtx/MvtxHitPruner` prunes the `TrkrHitSetContainer`
  - `mvtx/MvtxClusterizer` generates `TrkrClusters` from the `TrkrHitSetContainer`
* For the INTT:
  - `g4intt/PHG4InttHitReco` takes all PHG4 hits and writes `TrkrHitSetContainer`s to the node tree
  - `g4intt/PHG4MvtxDigitizer` digitizes the `TrkrHitSet`'s in the `TrkrHitSetContainer`
  - `intt/InttClusterizer` generates `TrkrClusters` from the `TrkrHitSetContainer`
* For the TPC:
  - `g4tpc/PHG4TpcElectronDrift` uses `g4tpc/PHG4TpcPadPlaneReadout` to geneate `TrkrHitSetContainer`
  - `g4tpc/PHG4TpcDigitizer` digitizes `TrkrHitSet`'s in the `TrkrHitSetContainer`
  - `tpc/TpcClusterizer` generate `TrkrClusters` from the `TrkrHitSetContainer`

For the truth tracks, this process needs to proceed for a `TrkrHitSetContainer` generated with the `PHG4Hit`s resulting from each individual truth track. This means either 
  1. Writing a variable number of `TrkrHitSetContainer`'s to the node tree (one for each truth track) and then having the down stream modules recognize and loop over those containers in their `::process_event` functions.
  2. Writing a module that sits inside of the `TrkrHitSetContainer` collects the `TrkrHitSetContainer` from the set of `PHG4Hit`'s for each truth track, and then run the logic of all the downstream pruner, digitizers, and clusterizers modules in place and just put the `TrkrClusters` and `TrkrTruthTrackContainer` information on the node tree.
  3.  Do the same as 2, but instead of reproducing the pruner, digitizer, and clusterizer, code, just add hooks to the pruner, digitizer, and clusterer modules to have a function that can run their logic on a provided `TrkrHitSetContainer` and fill a specific `TrkrClusterContainer`.

Currently the coresoftware does option 2, but imperfectly. It misses the pruner and digitization entirely, makes some assumptions about user parameters selected in the clustering (and will ignore any others selected at the Fun4All level), and it was reported just this morning that there is potentially some trouble with the reimplemented Mvtx custering code/logic. It  has a larger code surface area to make mistakes and/or inconsistencies when any of the digitizer, pruner, or clustering modules are fixed and/or updated and then the follow-on corrected must be made (if it is even recognized that it should be made) in the duplicated logic of the truth clusters (which would be organized differently anyhow, making the copy-over work not entirely trivial).  

This PR changes the situation to path 3. A common `g4tracking/TruthClusterizerBase` class is written that collects `PHG4Hit`s from the `TrkrHitSetContainer` generated, keeps track of the `TrkrTruthTrack` and `TrkrTruthTrackContainer` filling. This module merely requires a down-stream specification for the process to clusterize the `TrkrHitSetContainer`; these are  `g4{mvtx,intt,tpc}/PHG4{Mvtx,Intt,Tpc}TruthClusterizer` respectively. The logic says that if a user is wants `TrkrClusters` for the truth tracks, then they will also want clustering done for the Svtx Tracks, and must therefor also selected the correct parameters, and initialize the pruner, digitizer, and clusterer modules. Therefore, for the truth clusters, the user passes pointers to these modules to the truth clusterizers. These modules are used to cluster the `TrkrHitSetContainer` as they are generated for each truth track in the Mvtx, Intt, and Tpc at the time that it is created. The clustering itself is down by the same modules doing the Svtx clusters, so it uses all the same parameters, code, libraries, etc... I think this is the right path to take. It forces the user parameters (i.e. those that can be set from Fun4All) in the clustering to be the same for truth and Svtx tracks. It removes a lot of code duplication, and and therefore increases code reliability because clustering changes will be modified in only one place by the expert and automatically apply to both truth and svtx clustering.

The running logic them becomes this:
For Svtx the clustering is the same. If truth tracks are to be clustered the following happens:
* For the MVTX: 
    In the Fun4All maco, `PHG4MvtHitReco`, `PHG4MvtxDitigizer`, `MVtxHitPruner`, and `MvtxClusterizer` modules are instantiated and placed into the Fun4AllServer. Then pointers to the instances of  `PHG4MvtxDitigizer`, `MVtxHitPruner`, and `MvtxClusterizer` are passed to `PHG4MvtxHitReco` using the `PHG4MvtxHitReco::SetTruthClusterizer` function (which also tells the `PHG4MvtxHitReco` to cluster truth tracks). Then,
  - `g4mvtx/PHG4MvtxHitReco` takes all PHG4 hits, and writes `TrkrHitSetContainer`s to the node tree. As the same time, the truth clusterizer watches each `PHG4Hit`; for each `TrkrHitSet`'s resulting from a `PHG4Hit` from a truth track, it places it into a separate `TrkrHitSetContainer` which is not located on the node tree. When it is has finished filling the `TkrHitSetContainer` for a given truth track, it passes it to the appropriate functions in `PHG4MvtxDigitizer`, `MvtxHitPruner`, and `MvxtClusterizer`. It gets back the generated `TrkrClusters` and fill the `TrkrClusterContainer` named  "TRKR_TRUTHCLUSTERCONTAINER" on the node tree. It also flushes its local`TrkrHitSetContainer` to be used for the next truth track (if any). This will occur for each truth track in each event (i.e. multiple times for each `PHG4MvtxHitReco::process_event` function call).   
  - `g4mvtx/PHG4MvtxDigitizer` is called by Fun4All and digitizes `TrkrHitSet`'s in the `TrkrHitSetContainer` for Svtx tracks (same as always)
  - `mvtx/MvtxHitPruner` is called by Fun4All and prunes the `TrkrHitSetContainer` for Svtx tracks (same as always)
  - `mvtx/MvtxClusterizer` is called by Fun4All and generates `TrkrClusters` from the `TrkrHitSetContainer` for Svtx tracks (same as always)
The process is exactly analogous for Intt, and Tpc clusters.

Hope this answers the question, but happy to iterate over remaining questions!

david-stewart:
ok -- per the tracking meeting today, we are not going to pursue this method. We will break out the functionality separately for the trkr truth clusters and not use pointers to subprocesses on the node tree.

PR #1916, opened 2023-03-29
Re-adjusting the epd geometry code for sPHENIX

eumaka:
PR registers a streamlined epd geometry code for sPHENIX and fills in the geometry in the EPD reconstruction code per Chris P's request (following a similar setup as the calorimeters).


pinkenburg:
Not sure if this is still up to date - the scan build error is serious - you are returning a value of an array which just went out of scope. Not sure why scan build does not complain about the others. Declare the arrays static const and try again

eumaka:
Closing this, replacing with another.

PR #1919, opened 2023-03-30
Replace std::vector with TClonesArray in TrkrClusterContainer

tasheng:

## Types of changes

Switch from using `TrkrClusterContainerv4` to `TrkrClusterContainerv5`. Instead of using `std::vector` to store the clusters in the container, it uses `TClonesArray` for better I/O performance and disk space usage. 

## TODOs (if applicable)


hupereir:
Hello, I have added some comments on the code, hoping this helps. 

Hugo

offline/packages/trackbase/TrkrClusterContainerv5.cc
```diff
@@ -0,0 +1,293 @@
...
+     * print error message and exit otherwise
+     */
+
+    auto cluster = (TrkrCluster*) clus_vector->At(index);
+    std::cout << "get null" << std::endl;
+    if (!cluster)
+      {
+        // clus_vector[index] = newclus;
+        TrkrClusterv4* cls = (TrkrClusterv4*) clus_vector->ConstructedAt(index);
+        cls->CopyFromv4(static_cast<TrkrClusterv4*>(newclus));
```
hupereir, line 121: In addition, this will also leak "newclus" the code expects that the ClusterContainer takes ownership of newclus (as is the case in TrkrClusterContainerV3), whereas here, a Clone/deep copy of new clus is made and stored. As a result, "newclus" is never deleted. 
For this to work you would need to insert newclus directly in the TClonesArray (which might not work if it is not the right version), without deep copy involved ...

tasheng: Looking at the documentation, there is no way to insert an element into a TClonesArray without calling the constructor and making a deep copy. I can destruct `newclus` after it is copied into the array so that it doesn't leak.

offline/packages/trackbase/TrkrClusterContainerv5.h
```diff
@@ -0,0 +1,68 @@
...
+
+  /// temporary map
+  /**
+   * the map is transient. It must not be written to the output.
+   * To do this one adds //! after the declaration
+   * see https://root.cern.ch/root/htmldoc/guides/users-guide/InputOutput.html for details
+   */
+  Map m_tmpmap;  //! transient. The temporary map does not get written to the output
+
+  int m_cluster_version = 5;
```
hupereir, line 63: You cannot have m_cluster_version as a member of the container class. What it otherwise mean is that every time the cluster version is updated, the old cluster containers will not be readible anymore, which, in turn would in fact require that you update also the container to a new version. (essentially, with this you tie the container version to the cluster version, which is something we would really like to avoid)

offline/packages/trackbase/TrkrClusterContainerv5.cc
```diff
@@ -0,0 +1,328 @@
...
+     * print error message and exit otherwise
+     */
+
+    auto cluster = (TrkrCluster*) (*clus_vector)[index];
+    std::cout << "get null" << std::endl;
+    if (!cluster)
+    {
+      if (m_cluster_version == 4)
+      {
+        cls = (TrkrClusterv4*) clus_vector->ConstructedAt(index);
```
hupereir, line 137: This wont work once switching to the base class name for the TClonesArray. 
You need to use: 
cls = new(clus_vector[index]) TrkrClusterv5();
which, I think does the right thing.
Also, you do not need static_cast when calling ::CopyFrom. This one takes the base class as argument. 
Finally, since you should not use m_cluster_version, as discussed in separate comment, what you should better do is create an alias at the top of the file: like
using TrkrCluster_impl = TrkrClusterv5;
and here, with not if statement, just use: 
cls = new(clus_vector[index]) TrkrCluster_impl();
Note that you should be able to change the alias at the top of the file to any latest version of the TrkrCluster class without breaking compatibility.
** this is not tested. Just my recollection of some code used in Phenix back in the days **

offline/packages/trackbase/TrkrClusterContainerv5.cc
```diff
@@ -0,0 +1,328 @@
...
+      }
+    delete newclus;
+  }
+  else
+  {
+    // if index exceeds the vector size, resize cluster to the right size with nullptr, and assign
+    // clus_vector.resize(index + 1, nullptr);
+    // clus_vector[index] = newclus;
+    if (m_cluster_version == 4)
+    {
```
hupereir, line 177: same comments as above. 
Also there is no resize of the TClonesArray here, and thus this block is identical to the previous one. I would then simplify the if statements. (provided that this actually works ...)

hupereir (commented):
Hello, a few more comments on how to deal with versionned TrkrCluster objects, without having to update the container version. If I remember correctly this is how we were doing it in PHENIX days. I have not tested the changes ...

offline/packages/trackbase/TrkrClusterContainerv5.h
hupereir: mmm. I take it back, after discussing with Jin, Chris: the TClonesArray needs to know the name of the implemented class, (which you can also get from TrkrCluster_impl->GetName() by the way). 
In any case this would need careful  testing when reading back old files with previous TrkrCluster version.

pinkenburg:
The QA seems to indicate that the tracks are gone

blackcathj:
This has been pending and diverged from `master` branch. Closing for now

PR #1920, opened 2023-03-30
add base class functions

osbornjd:

## Types of changes

This PR adds some functions to the TrkrCluster base class which are in v5 but not in the base class currently. Having this will prevent any casting gymnastics to occur

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackbase/TrkrCluster.h
+++ b/offline/packages/trackbase/TrkrCluster.h
@@ -70,7 +70,7 @@ class TrkrCluster : public PHObject
   virtual void setEdge(char) {}
   virtual void setTime(const float) {}
   virtual float getTime() const { return NAN;}
-
+  virtual char getSize() const {return std::numeric_limits<char>::max(); }
   
 
   //
--- a/offline/packages/trackbase/TrkrCluster.h
+++ b/offline/packages/trackbase/TrkrCluster.h
@@ -62,9 +62,16 @@ class TrkrCluster : public PHObject
   //
   virtual void setAdc(unsigned int) {}
   virtual unsigned int getAdc() const { return UINT_MAX; }
-
+  virtual void setMaxAdc(uint16_t) {}
+  virtual unsigned int getMaxAdc() {return UINT_MAX; }
+  virtual char getOverlap() { return std::numeric_limits<char>::max(); }
+  virtual void setOverlap(char) {}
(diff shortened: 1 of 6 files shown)


osbornjd:
@pinkenburg I just want to make sure before merging this that these changes will not affect our DST readback capabilities.

pinkenburg (approved):
looks good - shouldn't cause problems (but anyone not recompiling their own code will see very interesting behavior)

hupereir:
@osbornjd 
Maybe before merging, I noticed that the TrkrClusterv5::CopyFrom method is very incomplete
I does not include copying any of the new members of the class (mostly because there was no base class accessors !)
Can you add them ? Alternatively I can do that quickly in a separate PR

osbornjd:
Thanks for pointing it out, I will add them to this PR since it is trying to clean up some of the derived classes

hupereir:
Also, the TrkrClusterv5::CopyFrom( TrkrClusterv5* source) method should go
It is irrelevant if the copyFrom( base class) is complete.

hupereir:
And I just noticed that the same is true for Clusterv4.
This is really a mess. We should pay more attention when increasing versions

osbornjd:
> Also, the TrkrClusterv5::CopyFrom( TrkrClusterv5* source) method should go It is irrelevant if the copyFrom( base class) is complete.

can you elaborate on this point? The base class implementation is just a blank empty function. Shouldn't this mean it should be left in?

hupereir:
> 

What I mean is that TrkrClusterv5::CopyFrom( TrkrCluster* ) should be enough
If you call all the base class methods in there, it will copy the right thing if what is passed as argument is actually a TrkrClusterv5, and thus will do the exact same thing as TrkrClusterv5::CopyFrom( TrkrClusterv5* )
(and it will also work for v6, v7, etc)
So the second method is superfluous, once the first one is fixed. 
Makes sense ?

osbornjd:
> 

I'm assuming one of those was supposed to be a `CopyFrom(const TrkrCluster&)` rather than `CopyFrom(TrkrCluster*)`. But I think I get the main point, basically they do the same thing

osbornjd:
@hupereir let me know if the last commit addresses all your points, I think it does.

hupereir:
> @hupereir let me know if the last commit addresses all your points, I think it does.

Looks good yes ! (as far as I can tell). Thanks !

PR #1939, opened 2023-04-06
Bbc vertex

pinkenburg:

## Types of changes
- [X ] New feature (non-breaking change which adds functionality)

This PR adds the new Bbc Vertex to our global vertices. Major rewrite - now vertices are added via an enum which dtermines the order of precedence, not just the order in which vertices are added which allows now to have multiple modues adding vertices independently. Still needs a lot of work - the Bbc currently produces two vertex objects, the global reco for the svtx vertex needs the old bbc vertex map which is not provided anymore. Also no consistency between old smeared t0 and bbc reconstructed t0 (~8ns offset)  

## TODOs (if applicable)

### DIFF
--- a/simulation/g4simulation/g4bbc/BbcSimReco.cc
+++ b/simulation/g4simulation/g4bbc/BbcSimReco.cc
@@ -116,22 +116,20 @@ namespace BBCINFO
   };
   */
 
-}  // namespace BBCINFO
-
-using namespace BBCINFO;
+}
 
 //____________________________________
 BbcSimReco::BbcSimReco(const std::string &name)
   : SubsysReco(name)
   , _tres(0.05)
 {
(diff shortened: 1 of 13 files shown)


osbornjd:
Does the bbc reco actually fill the `BbcOut` container with nonzero values? I tried to interface this with some of our tracking code but only got `-99999` for the vertex position from the `BbcOut` container

pinkenburg:
It does - tested it. But it doesn't always succeed, in which case you'll find -9999 (which is something we ought to change). But likely you did not read the corresponding DST_BBC_G4HIT file which contains them. You are probably better off using the globalvertex map once this PR is done. We have conflicting bbc output objects - I am not sure if BbcOut will be the one which will get chosen

pinkenburg:
The compiler checks are done, this code is not in any macro. I'll merge this to get it into the 10:24 daily build

PR #1946, opened 2023-04-07
Adding pointers protection

antoniocosilva:

## Types of changes

Adding protection against null pointers (towerinfo and towergeom)

### DIFF
--- a/offline/packages/CaloReco/RawClusterBuilderTopo.cc
+++ b/offline/packages/CaloReco/RawClusterBuilderTopo.cc
@@ -468,12 +468,20 @@ int RawClusterBuilderTopo::process_event(PHCompositeNode *topNode)
     for(unsigned int iEM = 0; iEM < towerinfosEM->size(); iEM++)
     {
       towerInfo = towerinfosEM->get_tower_at_channel(iEM);
+      if(!towerInfo)
+      {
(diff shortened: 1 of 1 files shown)


pinkenburg:
We have to be careful with just protecting the code from crashing. The first check protect against missing towers which may be a possibility in the future, but right now this is an indication that something is severely wrong somewhere.
The second check is for the geometry pointer. If this one is null we are in severe trouble and you would just drop acceptance without realizing. There should be at least a printout.
In general - I do prefer code to crash and burn rather than to continue silently when fatal errors are encountered and the results are likely tainted

antoniocosilva:
As far as I could see, the towers are always there and not crashing anything. I will add a print out there.
As for the geometry pointers, I agree that this could be a serious issue. At this moment, I'm still not sure why the pointer would not be found. I will ask Tim if he saw something like this happening too.

antoniocosilva:
I think I found the issue. I will cross-check it with Tim.

antoniocosilva:
Hi Cris, Tim made a PR to fix the bug that was making the geometry objects to crash the code. I think I agree with you that sometimes it's better if the code crash and burn so that we know that something is not quite right, so we should probably not include these protections and cancel this PR.

PR #1963, opened 2023-04-14
Adding Functionality for Laser Stepping Over Angles in ROOT file

luisval:
Files Affected:
simulation/g4simulation/g4tpc/PHG4TpcDirectLaser.h
simulation/g4simulation/g4tpc/PHG4TpcDirectLaser.cc
simulation/g4simulation/g4tpc/theta_phi.root

Changes:
- Added functionality to open a .root file which contains the theta and phi angles required to hit all the "pixels" of the central membrane.
- Pull NTuple of angles from root file
- Provides NTuple angles to the laser to be fired

TODO:
- suppress std::cout statements in AimToPatternStep_File function in the .cc
- delete .root file when we can add it to cvfms 
- general cleanup


hupereir:
Question: shouldn't the root file (considered as binary) got into our $CALIBRATION_MAIN, rather than be committed in the code ?

chughes90:
Hey Hugo,

For sure it should. We did not know how to do this so Luis did this a temporary fix. If you read the TODO part of the pull request comments it says:

>     * delete .root file when we can add it to cvfms

cvfms is where the $CALIBRATION_MAIN is I think?

In any case, we only meant it as a temporary thing right now - to be fixed when we put the root file in cvfms.

Is this unacceptable?

pinkenburg:
This might make the root file part of the git history which we will drag around forever and we should avoid bloat. The root file is small enough - please put this root file into our calibrations repo https://github.com/sPHENIX-Collaboration/calibrations and make a PR (just like this one here). Our build copies everything under $CALIBRATIONSROOT.
Then make a new branch, commit only the PHG4TpcDirectLaser files and make a fresh PR. I am closing this one

PR #1972, opened 2023-04-18
Associate tracks to BBC vertex

osbornjd:

## Types of changes

This PR addresses events where there is only 1 track per event, thus a properly defined `SvtxVertex` cannot be determined. In this situation, the track is propagated back to (0,0,0) to get the track parameters at the beamline. Then it is associated to the BBC vertex, if any is found. If there is no BBC vertex, the track is vertex-less. It improves the DCA_z of 1 track events to just under 1 cm, which is dominated by the BBC vertex reconstruction resolution. 

## TODOs (if applicable)


pinkenburg:
This will likely break or run in some weird way with our standard macro. In our standard macro the BbcVertexFastSimReco is enabled which smears the true vertex with the Bbc resolution. It fills the BbcVertexMap, not BbcOut. We should really sort this out before using the Bbc Vertex anywhere

osbornjd:
It only uses the BBC info if it is available on the node tree and checks explicitly - if the info is not there, it isn't used and only fills the an edge case of 1 track events. If you would prefer, I can write into the code for now to also check for the smeared BBC vertex in the event that the real reconstructed vertex is not available

osbornjd:
@pinkenburg if you are okay with merging this I will go ahead and do it, it passes the jenkins checks and I don't expect it to crash anything because if the BBC vertex (or BbcOut map, for that matter) is not found the single track event is just not assigned a vertex. Previously it was assigned vertex (0,0,0) which is just wrong. This edge case is called only for events with 1 track where a track vertex is unable to be created.

PR #1996, opened 2023-05-04
read from cdb if file is available

osbornjd:
Read the fieldmap from the CDB if it is available, per Chris' suggestion.

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackreco/MakeActsGeometry.cc
+++ b/offline/packages/trackreco/MakeActsGeometry.cc
@@ -35,6 +35,8 @@
 #include <fun4all/Fun4AllReturnCodes.h>
 #include <fun4all/Fun4AllServer.h>
 
+#include <ffamodules/CDBInterface.h>
+
 #include <phool/PHCompositeNode.h>
 #include <phool/PHDataNode.h>
 #include <phool/PHNode.h>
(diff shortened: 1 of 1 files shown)


pinkenburg:
That crash is a bug in the code. The CDB returns a fully qualified path, but the code prepends it with the path to $CALIBRATIONSROOT:
 char *calibrationsroot = getenv("CALIBRATIONROOT");
      m_magField = "sphenix3dtrackingmapxyz.root";
      
      m_magField = CDBInterface::instance()->getUrl("FIELDMAPTRACKING");
    
      if (calibrationsroot != nullptr)
      {
	m_magField = std::string(calibrationsroot) + std::string("/Field/Map/") + m_magField;
      }
That then results in
 /var/lib/jenkins/workspace/sPHENIX/Build-Master-gcc12@4/build/new/install.1/share/calibrations/Field/Map//cvmfs/sphenix.sdcc.bnl.gov/calibrations/sphnxpro/cdb/FIELDMAPTRACKING/8e/4d/8e4d6c3b1660540a658da3a275af2bde_sphenix3dtrackingmapxyz.root
being handed down as fieldmap. It look like MakeActsGeometry() then throws an exception if the file is not found

osbornjd:
Actually, we're not  going away from the CDB so is there any reason to leave the calibrations root stuff in there and not just read from the CDB by default?

pinkenburg:
That would be fine by me. I think initially there was an idea to keep the ability to run without cdb in some containers but we should cut that cord. Otherwise we just have backdoors which can bite us

osbornjd:
Ah I see, that makes sense, as I think I recall the singularity container doesn't have access to the CDB. I can leave it in just as a backup.

PR #2002, opened 2023-05-05
first version of MVTX unpacker

KvapilJ:
first version of MVTX data unpacker, currently no online_distributions calls are implemented since there is no decoder, there is hardcoded fill for the DST: 1 hitsetkey with 10 hitsets

## Types of changes

first version of MVTX data unpacker

## TODOs (if applicable)

implement actual online_distributions calls once we have decoder

### DIFF
--- a/offline/packages/mvtx/Makefile.am
+++ b/offline/packages/mvtx/Makefile.am
@@ -20,6 +20,7 @@ AM_LDFLAGS = \
 pkginclude_HEADERS = \
   MvtxClusterizer.h \
   MvtxHitPruner.h \
+  MvtxRawDataDecoder.h \
   CylinderGeom_Mvtx.h \
   SegmentationAlpide.h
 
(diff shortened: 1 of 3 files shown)


pinkenburg (changes requested):
Please follow our conventions for include files:
https://wiki.sphenix.bnl.gov/index.php/Codingconventions#Include_Files
this is important and will bite us if this is being ignored

KvapilJ:
not sure that was the requested change...changed <> to ""

pinkenburg:
https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html#Directory-Options

> not sure that was the requested change...changed <> to ""
There is a subtle but important difference between #include <> and #include ""
https://gcc.gnu.org/onlinedocs/gcc/Directory-Options.html#Directory-Options

-->1.For the quote form of the include directive, the directory of the current file is searched first. 

We use this feature to be able to put our modified include files in our own installation area ahead of the generally installed ones even if those includes are included by another include. Takes a while to wrap one's head around but it does the trick. If we don't follow our include file convention, this might break and you might deal with shared libraries which were build using contradicting include files which is endless fun (completely unpredictable and weird results)

pinkenburg (approved):

PR #2027, opened 2023-05-19
adding new EPD event plane reco

eumaka:
PR adds module to extract raw psi_n and phi weighted psi_n for the EPD. MBD will be added next once its geometry is ready.


eumaka:
This is becoming a hydra, closing this. Will be replaced.

PR #2058, opened 2023-06-01
Tpc geometry fix

hupereir:
This pull request moves all the Tpc Geometry initialization relevant to ACTS (via MakeActsGeometry) from PHG4TpcElectronDrift to PHG4TpcDetector. This allows to have the TPC geometry properly initialized when running e.g. on PRDFs, some thing that the current implementation prevents, since PHG4TpcElectronDrift runs only for simulations. 

There is still some duplicated code between PHG4TpcDetector and PHG4TpcPadPlaneReadout, because the later uses some internally built constants to handle dead areas between regions and sectors. Ultimately this should be changed to use only what's stored in PHG4TpcCylinderGeom. This will be the subject of future commits. 

I have double checked that the content of PHG4TpcCylinderGeom is identical between before and after the change.

## Types of changes

## TODOs (if applicable)


osbornjd:
There seems to be some issues in R1 of the TPC, did some parameter get overlooked?

https://nbviewer.jupyter.org/github/sPHENIX-Collaboration/QA-gallery/blob/jenkins-sPHENIX-test-tracking-low-occupancy-qa-3698-test-tracking_Event500_Sum30/QA-Tpc.ipynb

Or does this need to be run in conjunction with https://github.com/sPHENIX-Collaboration/macros/pull/646?

simulation/g4simulation/g4tpc/PHG4TpcPadPlane.cc
```diff
@@ -20,6 +20,9 @@ PHG4TpcPadPlane::PHG4TpcPadPlane(const std::string &name)
 
 int PHG4TpcPadPlane::InitRun(PHCompositeNode *topNode)
 {
+  
+  std::cout << "PHG4TpcPadPlane::InitRun" << std::endl;
```
osbornjd, line 24: Extra print statement

osbornjd (commented):
Looks fine to me other than the issues in R1 and a few leftover print statements, maybe there is just some parameter that is not being loaded properly

hupereir:
> There seems to be some issues in R1 of the TPC, did some parameter get overlooked?
> 
> https://nbviewer.jupyter.org/github/sPHENIX-Collaboration/QA-gallery/blob/jenkins-sPHENIX-test-tracking-low-occupancy-qa-3698-test-tracking_Event500_Sum30/QA-Tpc.ipynb
> 
> Or does this need to be run in conjunction with [sPHENIX-Collaboration/macros#646](https://github.com/sPHENIX-Collaboration/macros/pull/646)?

My understanding is that it should run without, but I might be wrong. 
Some debugging is in order, definitly. (so far I have checked only that the cylinderGeom object were identical, not the result of PHG4TpcElectronDrift. I'll do just that today. 

Thanks for going through the code !

hupereir:
All right. That should fix that. Let's see.

osbornjd:
Interesting that the other regions didn't also bomb out given their max radii were missing also...

PR #2073, opened 2023-06-06
cleaning up event plane determination module 

eumaka:
PR for the event plane determination for sPHENIX detectors; cleaned up and brought to the 21st century.


eumaka:
Replacing with updated request

PR #2097, opened 2023-06-19
fix: DCA calculation and duplication

osbornjd:
This removes a lot of duplicated DCA calculation code and consolidates it to use a single function in track analysis utilities. It also corrects a bug Tony found that rotates the vector clockwise rather than counter clockwise to make the y component of the 3D DCA 0.

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackbase_historic/TrackAnalysisUtils.cc
+++ b/offline/packages/trackbase_historic/TrackAnalysisUtils.cc
@@ -1,15 +1,14 @@
 #include "TrackAnalysisUtils.h"
 #include "SvtxTrack.h"
-#include <globalvertex/GlobalVertex.h>
 
 namespace TrackAnalysisUtils
 {
 
 TrackAnalysisUtils::DCAPair get_dca(SvtxTrack* track,
-				    GlobalVertex* gvertex)
+				    Acts::Vector3& vertex)
 {
   TrackAnalysisUtils::DCAPair pair;
-  if (!track or !gvertex)
+  if (!track)
   {
     std::cout << "You passed a nullptr, can't calculate DCA" << std::endl;
     return pair;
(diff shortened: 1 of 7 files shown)


osbornjd:
I moved the global vertex package compilation to earlier in the build, before the core tracking libraries, so that trackbase_historic can depend on the global vertex objects.

osbornjd:
Okay... global vertex depends on trackbase_historic too. Coming up with a different strategy...

offline/packages/trackbase_historic/TrackAnalysisUtils.cc
```diff
@@ -36,7 +34,7 @@ TrackAnalysisUtils::DCAPair TrackAnalysisUtils::get_dca(SvtxTrack* track,
 
   Acts::Vector3 r = mom.cross(Acts::Vector3(0., 0., 1.));
   float phi = atan2(r(1), r(0));
-
+  phi*= -1;
```
osbornjd, line 37: @blackcathj Tony found a bug in the DCA calculation during the rotation that left a nonzero y component after the rotation, which is fixed by flipping the azimuthal direction of the rotation to clockwise. This now gives the resulting DCA y component equal to 0 which changes the DCA performance, making it somewhat worse as the QA shows. This is probably why there was a very sharp peak at DCA_xy~0 prior to this fix, which can be seen in the reference histograms. I will wait to merge until you comment further

adfrawley: This commit gives the correct dca3dxy. Reese and I looked at it carefully. The previous code yielded a value that was only a fraction of the track-vertex residual in the plane of the perigee surface. The dca3dxy must go up when that is fixed.

PR #2143, opened 2023-07-14
Updates to the jet reconstruction:

david-stewart:
Can fill JetContainer(s) instead of JetMap(s)
  - requires that Jetv2 be used - Jetv2 internally has vector<float> that contain all additional data (jet area, zg, etc...) - This additiona data is indexed and access from the JetContainer
  - the Jetv2 are stored in a TClonesArray, instead of std::map<unsigned int, Jet*>

I will present this first to the Jet Substructure group before making the pull request. There is still a lot of work to come, but this is a good, working point, to stop and get feedback for the downstream users and other experts.

## Types of changes
       Updates to the FastJetAlgo, JetAlgo, and JetReco to use a new JetContainer in place of a JetMap.

- Use of TClonesArray(s) will help with execution instead of the std::map in use in JetMap classes
- Internal to each Jetv2 there is:
    -  vector<float> for the properties instead of a map<Jet::PROPERTY,float> in Jetv1
    - vector<pair<Jet::SRC,unsigned int>> in place of std::multimap<Jet::SRC, unsgiend int> in Jetv1

## TODOs (if applicable)


osbornjd:
This is my fault I was tooling around with some things in the package list. Push an empty commit to retrigger Jenkins

david-stewart:
Hi @pinkenburg , I have tried to address your concerns and remove the vestigal code that had been left over from the sorting functionality. Is there anything more that you will to be vetted from your side before merging this? Thanks, Dave

offline/packages/jetbase/Jetv2.h
pinkenburg, line 1: please order the methods/data members so there is one public and one private section. Currently this has multiple private/public declarations. This is legal C++ but this makes this hard to understand

pinkenburg:
I gave the feedback in MM:
https://chat.sdcc.bnl.gov/sphenix/channels/jet-structure-tg

The include file ordering is mostly not correct:
https://wiki.sphenix.bnl.gov/index.php/Codingconventions#Include_Files
using the wrong ordering in include files will create problems if you have your own installation area with modified files

Then in Jetv2.h :
mutable bool _is_sorted { false }; // if the _comp_ids (component ids) are sorted
I thought sorting is done in a helper class. We shouldn't carry flags into our DSTs

The Jetv2.h is done but the proper include file ordering (and use of "" vs <>) is not.

david-stewart:
ok sounds good. I didn't mention in my comment, but I had fixed the include statements with the previous commit. They all match the documentation that you sent.

I have now also rearranged the private/public methods with this commit

offline/packages/jetbase/ClusterJetInput.h
pinkenburg, line 1: Why does this include exist?

offline/packages/jetbase/ClusterJetInput.cc
pinkenburg, line 1: Why does this file exist?

offline/packages/jetbase/FastJetAlgo.h
pinkenburg, line 1: use #include <fastjet/...>

offline/packages/jetbase/FastJetOptions.cc
pinkenburg, line 1: leave empty lines between groups of include
run clang-format -style=file -i

offline/packages/jetbase/FastJetOptions.h
pinkenburg, line 1: Please do not use "_" as variable name, this will confuse people reading this code

offline/packages/jetbase/Jet.h
pinkenburg, line 1: why is mv1 and mv2 which are JetV dependent implemented in the base class?
they have the same signatures (string, ostream), and presumably do the same, why not a generic mv and implement the real method in the versioned class?

offline/packages/jetbase/JetContainer.h
pinkenburg, line 1: use climits instead of limits.h

offline/packages/jetbase/JetContainerv1.cc
pinkenburg, line 1: use int i = 0 instead of int i {0}

offline/packages/jetbase/TowerJetInput.h
pinkenburg, line 1: move calobase include ahead of iostream

pinkenburg (changes requested):
clang-format -style=file -i 
needs to be applied to all modified sources.

pinkenburg:
just noticed, it seems to affect the e- jet qa:
[QA-calorimetric-jet for e- at p_T=4GeV](https://nbviewer.jupyter.org/github/sPHENIX-Collaboration/QA-gallery/blob/jenkins-sPHENIX-test-calo-single-qa-5107-e-_pT4_Sum14/QA-calorimetric-jet.ipynb) : combined Chi2/nDoF = 332.222 / 42, and combined p-Value = 8.6305e-47
All other PR's are fine, so it is likely related to this one

offline/packages/jetbase/FastJetOptions.cc
david-stewart: done

offline/packages/jetbase/Jet.h
david-stewart: I have updated them to a single depmsg which uses two static strings for the differentiation between version.

offline/packages/jetbase/JetContainer.h
david-stewart: done

offline/packages/jetbase/JetContainerv1.cc
david-stewart: ok, changed in all places in this file.
But why? int i{0} is correct, and at least as far as the venerable and opinionated Bjarne Stroustrup seems to say, it is probably preferable because is it acceptable in _all_ contexts, see for instance: https://www.stroustrup.com/list-issues-2.pdf

offline/packages/jetbase/TowerJetInput.h
david-stewart: done

offline/packages/jetbase/FastJetAlgo.cc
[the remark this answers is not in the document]
david-stewart: ok, think I got them all

offline/packages/jetbase/ClusterJetInput.h
david-stewart: Not sure. Git blame says that all I did was add include guards in the *.h file; otherwise it all predates me. Should I remove it?

offline/packages/jetbase/FastJetAlgo.h
david-stewart: done

offline/packages/jetbase/FastJetOptions.h
david-stewart: ok, done

pinkenburg:
The calorimeter jet QA for e- is very different from the reference, it looks like some histos aren't filled with the code in this PR. This needs to be understood and fixed

osbornjd:
@david-stewart The QA code can be found [here](https://github.com/sPHENIX-Collaboration/coresoftware/blob/master/offline/QA/modules/QAG4SimulationCalorimeter.cc). In that directory you can also see the other QA modules (e.g. tracking etc.). You can compile these locally and run them in your macro by setting the flag `Enable::QA=true`, which will dump all of the information from these modules into a TFile that you can look through to identify what is going haywire.

david-stewart:
Hey @pinkenburg and @osbornjd ,
I started looking at this yesterday. There is a failure somewhere, presumably in the QA, where things die because a Jetv1 is getting called with a method defined only for Jetv2. This is pretty weird. I haven't found the location yet, but hopefully will be an easy fix.

osbornjd:
All the QA modules are under this area:
https://github.com/sPHENIX-Collaboration/coresoftware/tree/master/offline/QA/modules
So you can check there for issues

pinkenburg:
The calorimetric jet for e- still shows a drastic change because of empty historgrams (I assume). I converted the PR to draft so it doesn't get merged accidentally

osbornjd:
I restarted your build manually after Jenkins crashed yesterday and it looks like there are still some nontrivial changes to one of the calo QAs, these should be checked if they are understood
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-calo-single-qa/5415/

david-stewart:
> I restarted your build manually after Jenkins crashed yesterday and it looks like there are still some nontrivial changes to one of the calo QAs, these should be checked if they are understood https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-calo-single-qa/5415/

Thanks -- I am running things manually and trying to pin down the differences, using the QA macro indicated by @blackcathj. When running a single event, the JetMap's have identical output, but something is happening with multiple events. Still trying to understand why.

pinkenburg:
It now passes the calorimeter jet qa for e- without modifications in jenkins, good to go

PR #2180, opened 2023-08-10
adding fractional pad and time bin access to TPC geo and TrkrNtuplizer

bogui56:

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/simulation/g4simulation/g4detectors/PHG4TpcCylinderGeom.h
+++ b/simulation/g4simulation/g4detectors/PHG4TpcCylinderGeom.h
@@ -49,7 +49,9 @@ class PHG4TpcCylinderGeom : public PHG4CylinderGeom
   virtual int get_zbin(const double z) const;
   virtual int get_phibin(const double phi, int side = 0) const;
   virtual int get_phibin_new(const double phi) const;
-  
+
+  virtual float get_pad_float(const double phi, int side = 0) const;
+  virtual float get_tbin_float(const double z) const;
   virtual int find_phibin(const double phi, int side = 0) const;
 
   void set_layer(const int i) override { layer = i; }
--- a/offline/packages/TrackingDiagnostics/TrkrNtuplizer.cc
+++ b/offline/packages/TrackingDiagnostics/TrkrNtuplizer.cc
@@ -78,6 +78,7 @@ TrkrNtuplizer::TrkrNtuplizer(const string& /*name*/, const string& filename, con
   , _ntp_vertex(nullptr)
   , _ntp_hit(nullptr)
   , _ntp_cluster(nullptr)
+  , _ntp_clus_trk(nullptr)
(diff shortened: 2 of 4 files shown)


simulation/g4simulation/g4detectors/PHG4TpcCylinderGeom.h
```diff
@@ -49,7 +49,9 @@ class PHG4TpcCylinderGeom : public PHG4CylinderGeom
   virtual int get_zbin(const double z) const;
   virtual int get_phibin(const double phi, int side = 0) const;
   virtual int get_phibin_new(const double phi) const;
-  
+
+  virtual float get_pad_float(const double phi, int side = 0) const;
```
osbornjd, line 53: @pinkenburg will adding methods to our geometry objects screw up DST readback? These aren't versioned like the data model objects so I assume if you try to access these functions using already produced DSTs there will be some issue. This probably doesn't matter for this particular case because these provide expert only information so it is unlikely anyone would be trying to use these who isn't producing their own DSTs, but I figured I'd ask first

bogui56:
Hi Joe,

for what its worth, this just adds utility functions for coordinate conversion.
I don't think this should cause compatibility issues.

Cheers

  Christof

On 10. Aug 2023, at 18:54, Joe Osborn ***@***.***> wrote:



@osbornjd commented on this pull request.

In simulation/g4simulation/g4detectors/PHG4TpcCylinderGeom.h<https://github.com/sPHENIX-Collaboration/coresoftware/pull/2180#discussion_r1290422859>:

> @@ -49,7 +49,9 @@ class PHG4TpcCylinderGeom : public PHG4CylinderGeom
   virtual int get_zbin(const double z) const;
   virtual int get_phibin(const double phi, int side = 0) const;
   virtual int get_phibin_new(const double phi) const;
-
+
+  virtual float get_pad_float(const double phi, int side = 0) const;


@pinkenburg<https://github.com/pinkenburg> will adding methods to our geometry objects screw up DST readback? These aren't versioned like the data model objects so I assume if you try to access these functions using already produced DSTs there will be some issue. This probably doesn't matter for this particular case because these provide expert only information so it is unlikely anyone would be trying to use these who isn't producing their own DSTs, but I figured I'd ask first

hupereir:
As far as I know, adding methods only (not members, not static members) to a persistent class does not break reading back the older versions ...

hupereir:
In fact we do that all the time to the base class of our versionned classes.

osbornjd:
Good point, but the versioned classes are somewhat insulated from this for exactly the reason that you stated - the base class always has a default method so that if you call a function that doesn't exist in a versioned implemented class it can always default to the base class. Anyway, it was more for my own curiosity... I agree it seems like it would be okay since no new members are being added

pinkenburg:
There is no problem adding methods

PR #2197, opened 2023-08-21
Update eta range of flow in HepMCFlowAfterBurner.cc

eumaka:
Extend the eta range in which the flow after burner is applied.


eumaka:
just noticed this PR has already been placed by @pinkenburg

PR #2239, opened 2023-09-26
Added method to InttMapping to get certain raw fields from new InttRa…

josephbertaux:
…wHit object

## Types of changes

## TODOs (if applicable)


josephbertaux:
Added Intt's "Event Header" (InttEventInfoContainer) to trackbase subdir. Unfortunately it sits on top of a couple commits to the intt subdir, but I was planning on merging them eventually anyway. Those changes are desireable/compatible/compilable, but there will be more to follow.

It compiles but I haven't messed with incorporating it into the unpacker workflow to properly test it. I might need some hand-holding with that.

pinkenburg:
The InttEventInfo class is needlessly complicated and the scheme will break the first time another version will be added since you rely on definitions in the base class. v1 contains a single map: map<unint64_t, uint64_t>, what should be stored here? Please implement something along the lines of the other info classes.

josephbertaux:
What I was hoping to store is a map from event number to full BCO. The idea is one of my unpackers would find (and add if not found) a node with an InttEventInfoContainer (similar to how it looks for a "TRKR_HITSET" node), and append to the map of the pointer to InttEventInfoContainer.

Ideally the InttEventInfoContainer would be a branch in a TTree on disk that's written at the end of the Fun4All macro, but I'm not sure what I need to do on the back end to make sure it works this way--is it enough to inherit from PHObject?

I can get rid of the struct in the base class and have more general get/set methods (e.g., float& get_f_val(std::string const&) or similar). That way each version can have its own map type but a similar IO. I'll need to pick out a key type, b/c the virtual signatures in base class depend on that, but at least the value type and i/o handling can be versioned.

Let me know what you think and I'll make a new PR

pinkenburg:
Please implement an InttEventInfo Container along the other tracking detectors. But first coordinate with the tracking group what is actually needed from their side

PR #2277, opened 2023-10-13
Update to Cluster Position Dependent Correction

ahodges21:

This update adds CDB functionality to the residual correction of the position dependent correction.

## Types of changes

## TODOs (if applicable)


offline/packages/CaloReco/RawClusterPositionCorrection.cc
pinkenburg, line 1: This check is dangerous:
if (calibdir[0] == '/')
If there is no entry the cdb will return an empty string. But if we in the future decide to use local files a check for "/" will fail even for a valid calibration

ahodges21: ROOT has the TFile implementation IsZombie(), you think that's a better implementation?

osbornjd:
The scan build update errors are from the Acts update, please ignore

pinkenburg:
Allways one more thing - there is a memory leak by not deleting the cdbhistos after use (either once the historgam has been copied from it or in the dtor)

bseidlit:
This is looking great! But are we sure we don't want to interpolate between the bins or something? Also do we want to get the energy boundaries for where the residual correction is aplied from the histo, then all we need to do is upload a new histo to extend the range?

ahodges21:
> This is looking great! But are we sure we don't want to interpolate between the bins or something? 

I'd say absolutely, but I wanted to get what we used for QM up and running first before upgrading the performance. Right now the deviations between bins is super small, so I don't think it's a priority to change. 

Also do we want to get the energy boundaries for where the residual correction is aplied from the histo, then all we need to do is upload a new histo to extend the range?

You mean instead of doing "if(cluster energy > blah && cluster energy < blah)" just go by something like Hist -> GetBinEdge(lastBin)? We can certainly implement that.

PR #2343, opened 2023-11-09
New MvtxEventInfov2

cdean-github:

## Types of changes

I just noticed that the MVTX event header can only store strobe BCOs when there is an associated L1 trigger BCO seen in the event. I made a new version that inherits from v1 so the strobe BCO is saved for every event, irregardless of whether there is a L1 BCO

## TODOs (if applicable)

### DIFF
--- a/offline/framework/ffarawobjects/Makefile.am
+++ b/offline/framework/ffarawobjects/Makefile.am
@@ -91,7 +91,6 @@ libffarawobjects_la_SOURCES = \
   MicromegasRawHitv1.cc \
   MvtxRawHitContainerv1.cc \
   MvtxRawHitv1.cc \
-  MvtxRawEvtHeader.cc \
   MvtxRawEvtHeaderv1.cc \
   TpcRawHitContainerv1.cc \
   TpcRawHitv1.cc
--- a/offline/framework/ffarawobjects/MvtxRawEvtHeader.h
+++ b/offline/framework/ffarawobjects/MvtxRawEvtHeader.h
@@ -20,7 +20,13 @@ public:
   virtual void AddFeeId(const std::set<uint16_t>&) { return; }
   virtual void AddL1Trg(const std::set<uint64_t>&) { return; }
 
+  virtual std::set<uint16_t>& getMvtxFeeIdSet() { return dummySet16; };
+  virtual std::set<uint64_t>& getMvtxLvL1BCO() { return dummySet64; };
+
 private:
+  std::set<uint16_t> dummySet16;
+  std::set<uint64_t> dummySet64;
+
   ClassDefOverride(MvtxRawEvtHeader, 1)
 };
 
--- a/offline/framework/ffarawobjects/MvtxRawEvtHeaderv1.h
+++ b/offline/framework/ffarawobjects/MvtxRawEvtHeaderv1.h
@@ -5,6 +5,8 @@
 
 #include "MvtxRawEvtHeader.h"
 
+#include <phool/PHObject.h>
+
 #include <cstdint>
 #include <set>
 
@@ -13,7 +15,7 @@ class  MvtxRawEvtHeaderv1: public MvtxRawEvtHeader
 
 public:
   MvtxRawEvtHeaderv1() = default;
-  virtual ~MvtxRawEvtHeaderv1() = default;
+  ~MvtxRawEvtHeaderv1() = default;
 
   ///Clear Event
   void Reset() override;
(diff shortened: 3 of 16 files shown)


osbornjd:
What is the reason for having it inherit from the `v1` instead of just the base class directly? If new versions are needed will you be forced to inherit from `v2` and so on, and would that be problematic if e.g. you wanted a v3 that didn't need functionality in the v1?

cdean-github:
> What is the reason for having it inherit from the `v1` instead of just the base class directly? If new versions are needed will you be forced to inherit from `v2` and so on, and would that be problematic if e.g. you wanted a v3 that didn't need functionality in the v1?

This was just to add the strobe BCO information while keeping all the v1 info that we need. I thought it would be better to inherit from v1 rather than copy the code over but I can make a v2 that inherits just from the base class

pinkenburg (commented):
In most cases a base class pointer should be used instead of a version. E.g. getClass should always use the base class
mvtx_event_header = findNode::getClass<MvtxEventInfov1>(topNode, "MVTXEVENTHEADER");
should be
mvtx_event_header = findNode::getClass<MvtxEventInfo>(topNode, "MVTXEVENTHEADER");
same for datamembers
MvtxRawEvtHeaderv1* mvtx_raw_event_header = nullptr;
should be
MvtxRawEvtHeader* mvtx_raw_event_header = nullptr;
The only place where you need a version is where the object is actually created
mvtx_event_header = new MvtxEventInfov1();
everywhere else using a version is redundant (lucky) or hurting (e.g. getClass will not return a v2 object if you explicitly ask for a v1)
the class def overrides are essential - you cannot comment them out
//ClassDefOverride(MvtxEventInfov1, 1)
often this causes trouble when the override keyword is not used (properly) in the class. Then that needs to be fixed

pinkenburg:
I converted this PR to draft. The mvtx group sprinkles hardcoded versions all over the place and methods from the versioned classes are not propagated to the base class. There is only one place where you are allowed to use the version - when the object is created. Everywhere else, use the base class. If that doesn't work - there is something wrong in your code which has to be fixed.
As it turns out the new MvtxEventInfo is not properly versioned at all. I gave up at that point - the current PR will not pass jenkins.

cdean-github:
> I converted this PR to draft. The mvtx group sprinkles hardcoded versions all over the place and methods from the versioned classes are not propagated to the base class. There is only one place where you are allowed to use the version - when the object is created. Everywhere else, use the base class. If that doesn't work - there is something wrong in your code which has to be fixed. As it turns out the new MvtxEventInfo is not properly versioned at all. I gave up at that point - the current PR will not pass jenkins.

@pinkenburg @osbornjd  I'm working on the changes and found an issue. There is a change to use ClusHitsVerbose instead of ClusHitsVerbosev1 in MvtxCusterizer (I didn't write either of these classes). I understand using the base class when we call objects but this kicked up an error. I checked doxygen and ```addPhiHit``` indeed doesn't exist in the base class. Is it OK for me to add this function to the base class or will this brick everything?
```bash
../MvtxClusterizer.cc: In member function 'void MvtxClusterizer::ClusterMvtx(PHCompositeNode*)':
../MvtxClusterizer.cc:439:51: error: 'class ClusHitsVerbose' has no member named 'addPhiHit'
  439 |         for (auto& hit : m_phi) mClusHitsVerbose->addPhiHit    (hit.first, (float)hit.second);
      |                                                   ^~~~~~~~~
../MvtxClusterizer.cc:440:51: error: 'class ClusHitsVerbose' has no member named 'addZHit'
  440 |         for (auto& hit : m_z)   mClusHitsVerbose->addZHit      (hit.first, (float)hit.second);
      |                                                   ^~~~~~~
../MvtxClusterizer.cc:441:27: error: 'class ClusHitsVerbose' has no member named 'push_hits'
  441 |         mClusHitsVerbose->push_hits(ckey);
      |                           ^~~~~~~~~
make[1]: *** [MvtxClusterizer.lo] Error 1
make[1]: Leaving directory `/gpfs/mnt/gpfs02/sphenix/user/cdean/software/coresoftware/offline/packages/mvtx/build'
make: *** [install] Error 2
```

I also have a second question, if there are methods in the base class which should in principle never need to be altered in a derived class, do I still need o implement them in the derived class? What I mean is in the MvtxEventHeader base class I have some generic functions for adding and reading sets of strings, floats and ints. The logic should change for these basic setters and getters so I didn't implement them in the derived classes. (In principle, we know the info we need in the header so they could also be removed but this loses the ability to use them later).

The header looks like:
```c++
  void set_stringval(const std::string & /*name*/, const std::string & /*ival*/);
  std::string get_stringval(const std::string & /*name*/) const;

  virtual void set_number_HB(const int /*ival*/) {};
  virtual int get_number_HB() const { return 0; };
  ```

PR #2344, opened 2023-11-09
Add ability to get bunch crossing independent of INTT in pp

adfrawley:

## Types of changes

This PR introduces redundancy in the determination of the bunch crossing during extended-readout pp running. It will prevent loss of tracking acceptance if a part of the INTT is not working. The changes are:

1) Estimate the bunch crossing in PHSiliconTPCTrackMatcher using the z offset between (otherwise) matched silicon and tpc seeds. This is accurate to +/- 2 bunch crossings.
2) Make SvtxTrackSeed_v2 and add the bunch crossing estimate to it.
3) In PHActsTrkFitter, fit any track that does not have a valid INTT bunch crossing by postulating bunch crossing values of (crossing estimate +/- 2) when calculating the TPC z positions. The best chisq/ndf is well defined, and this is chosen.

It works well, finding the correct bunch crossing in essentially all cases. It does produce about 5-10% additional false positives, those will have to be dealt with using analysis cuts. 

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackbase_historic/Makefile.am
+++ b/offline/packages/trackbase_historic/Makefile.am
@@ -24,6 +24,7 @@ pkginclude_HEADERS = \
   TrackSeed.h \
   TrackSeed_v1.h \
   SvtxTrackSeed_v1.h \
+  SvtxTrackSeed_v2.h \
   TrackSeed_FastSim_v1.h \
   TrackSeedContainer.h \
   TrackSeedContainer_v1.h \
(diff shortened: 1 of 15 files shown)


adfrawley:
There is something I am missing here. cpp-check claims that the copy constructor for the new SvtxTrackSeed_v2 class does not assign the member variables of the class. Looking at the full report, I see the same message is there for the existing SvtxTrackSeed_v1 class. Both classes explicitly set the member variable values from the input seed. Does anyone understand this?

pinkenburg:
There is nothing wrong with the code - cppcheck does not go down the call stack. We do the right thing in the CopyFrom method which is called from the copy ctor but cppcheck ignores this. We had a suppression for this but cppcheck changed the warning type, so the suppression didn't do anything. I updated this here and all other places (so we should see a nice drop in cppcheck warnings)

PR #2405, opened 2023-12-12
Distortion modified transforms

adfrawley:

## Types of changes

This PR modifies the way that TPC crossing and distortion corrections are applied in PHActsTrkFitter.

Previously this was done by applying the distortion corrections to the ideal global position, projecting the corrected cluster global position back to the surface, then transforming the global position to local coordinates on the surface. 

Now the correction to the global position is calculated and made into an affine transform, which is then used to modify the ideal geometry transform for that cluster. The cluster local coordinates are not changed. The modified transform is placed in a temporary copy of the transformation container that is made for each track, a new GeometryContext is set equal to the temporary transformation container, and Acts uses that when it fits the track. 

This is more in line with the Acts philosophy that the measurement is local to the surface and all corrections are applied to the surface position. This makes the crossing and distortion correction like a track-by-track alignment correction.

Tested with Pythia pileup events, where the crossing correction can be 10's of cm. Seems to work well. Let's see what the QA looks like.

Not tested yet with distortion corrections, but these are small compared with crossing corrections. This change will be applied also to PHCosmicsTrkFitter and PHActsGSF, once we evaluate it for a while.

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackbase/ActsGeometry.cc
+++ b/offline/packages/trackbase/ActsGeometry.cc
@@ -172,3 +172,28 @@ Acts::Transform3 ActsGeometry::makeAffineTransform(Acts::Vector3 rot, Acts::Vect
 
   return actsAffine;
 }
+
+Acts::Vector2 ActsGeometry::getLocalCoords(TrkrDefs::cluskey key, TrkrCluster* cluster)
+{
+  Acts::Vector2 local;
+
+  const auto trkrid = TrkrDefs::getTrkrId(key);
+  if(trkrid == TrkrDefs::tpcId)
+    {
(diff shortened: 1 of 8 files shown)


osbornjd:
This failed in the jupyter notebook, not sure what happened here
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-calo-single-qa/5590/console

osbornjd (approved):
This looks good to me. As a side note, we should come up with a helper module which does the source link translation since the code is largely the same between all of the fitting modules. That would reduce the amount of duplicated code and then make these types of changes easier as they only have to happen in one place

PR #2432, opened 2024-01-09
Add fastjet::contrib::ConstituentSubtractor to FastJetAlgo

david-stewart:
Allow the user to add a reference for a probe into a FastJetAlgo instance The user sets the phi and eta range of the probe (as well as pT value or range). The probe is subsequently clustered with the jet particles, and the JetContainer is fed the probes location, phi, eta, pt values. The use can then look at the jets clustered with the probe (and could, of course, compare to other jets clustered without the probe). The utility is for jet background fluctuation studies. I'll talk to the JetStructure group is see if we do, or do not, want to merge. 

(Note: It does add some memory to every JetContainer, name one bool and an array<float,3> for the probe size, such that JetContainer_v1's size is modified. I suspect if we do add it, way might end up adding it into a JetContainerv2, or maybe separately put the JetProbe on the node tree along with an appropriate source tag under the Jet::SRC namespace, which may be a more elegant way of doing this. In any case, it is good to have here for documentation for a study that I am running.) 

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/packages/jetbase/FastJetAlgo.cc
+++ b/offline/packages/jetbase/FastJetAlgo.cc
@@ -12,14 +12,18 @@
 // fastjet includes
 #include <fastjet/AreaDefinition.hh>
 #include <fastjet/ClusterSequenceArea.hh>
+#include <fastjet/Selector.hh>
 #include <fastjet/tools/BackgroundEstimatorBase.hh>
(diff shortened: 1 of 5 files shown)


pinkenburg:
The jenkins failure is a calibration mismatch issue, not from this PR. It is fixed in the master branch

PR #2452, opened 2024-01-23
add MBD event plane recentering histograms

eumaka:
Recentering histograms are filled by the module during first run. On second run, these histograms are used for MBD event plane calibration. Currently uses the MBD total charge instead of centrality.


eumaka:
Output for changes added:
[correction-hists-output-for-psi2.pdf](https://github.com/sPHENIX-Collaboration/coresoftware/files/14028055/correction-hists-output-for-psi2.pdf)
[raw-and-recentered_psi_2.pdf](https://github.com/sPHENIX-Collaboration/coresoftware/files/14028056/raw-and-recentered_psi_2.pdf)

PR #2455, opened 2024-01-24
MVTX Hot Pixel Masking Implementation with CDBTree

MYOMAO:
[comment]: This pull request is made to implement the hot pixel masking for MVTX. The JSON file of the MVTX hot pixel maps has been converted to CDBTree format and is added to the database:   std::string database = CDBInterface::instance()->getUrl("MVTX_HotPixelMap");  //This is specifically for MVTX Hot Pixels

I have checked the preliminary validation on the noisy pixel map. All MVTX pixels passing the masking are not in the hot pixel map. Also, I checked several MVTX chips. The masked pixels are all in the MVTX hot pixel map.

The details on the study can be found at my presentation in sPHENIX software meeting on 01/23/2024: https://indico.bnl.gov/event/21806/contributions/85935/attachments/52243/89357/sPHENIX%20EvtGen%20HF%20Topical%20Meeting%2001-23-2024.pdf

## Types of changes

## TODOs (if applicable)


cdean-github:
Hi Zhaozhong, 

Thanks for the PR. I have some questions, comments and suggestions:

I think you can see that the reason for the clang and gcc is a whitespace in Makefile.am

The valgrind failure is due to creating a new object without calling a delete after it was used:. The code is in the unpacker but wasn't added by you. I don't know why valgrind didn't error on this before but could you add a delete after the object has finished its use? The line you're looking for is:
```c++
hit = new TrkrHitv2;
```

You have a map to hold the noisy pixels but I'm a bit confused by it. The keyed value is a pair of HitSetKeys and HitkKeys but the mapped value always seems to be 1? If I understand correctly, you create a new key for every hot pixel but what is the purpose of the mapped value? If it has no purpose then wouldn't it be better to use an std::set of std::pairs or a map where the key is the HitSetKey and the mapped value is a vector of HitKeys for that HitSetKey?
```c++
  HotPixelMap.insert({std::make_pair(HotPixelHitKey,HotHitKey),1});
```

I see you're including a JSON parser but I thought we got rid of reading JSON files? Is this supposed to have been removed?
```c++
#include <nlohmann/json.hpp>
...
using json = nlohmann::json;
```

You have:
```c++
using namespace std;
```
but I was under the impression that this is not allowed by our coding conventions. Chris or Joe can correct me if I'm wrong but it should be removed.

I also see there is a ```cout``` statement that is not contained within a verbosity check. This means it will print for every new run. I don't think it matters what verbosity level it's under but it shouldn't print by default.

There also seems to be a lot of random line breaks, tabs and code commented out. This should be cleaned and removed appropriately. For example:
```c++

  cout << "NPixel = " << NPixel << endl;


  for(int i = 0; i < NPixel; i++){

	  int Layer = cdbttree->GetSingleIntValue(std::string("layer_") + std::to_string(i));
	  int Stave = cdbttree->GetSingleIntValue(std::string("stave_") + std::to_string(i));
	  int Chip = cdbttree->GetSingleIntValue(std::string("chip_") + std::to_string(i));
	  int Col = cdbttree->GetSingleIntValue(std::string("col_") + std::to_string(i));
	  int Row = cdbttree->GetSingleIntValue(std::string("row_") + std::to_string(i));

	 // cout << "Hot Pixel: " << i << "   Layer: " << Layer << "   Stave: " << Stave << "   Chip: " << Chip << "   Col: " << Col << "   Row: " << Row << endl;

	  TrkrDefs::hitsetkey HotPixelHitKey = MvtxDefs::genHitSetKey(Layer, Stave, Chip, 0);     	
	  TrkrDefs::hitkey HotHitKey = MvtxDefs::genHitKey(Col,Row);
	  HotPixelMap.insert({std::make_pair(HotPixelHitKey,HotHitKey),1});

  }


  return Fun4AllReturnCodes::EVENT_OK;
  ```
  
  I think with these alterations, the PR should pass Jenkins and line up with the coding conventions

**Comment by osbornjd** [comment]
Interesting, github has a bug where when updating a PR on master it sometimes shows already existing commits in the upstream origin master branch as a part of the PR. Despite a lot of information online about this, the only way to fix it is to swap the target branch temporarily ..
https://github.com/orgs/community/discussions/39031

**Comment by cdean-github** [comment]
@osbornjd I was looking at the line where the noisy pixel is generated as a hit set. The last value here is 0 which is the strobe information. Do we not need to generate a noisy pixel hit key for all possible strobe values or else this ill only mask pixels that are synched with the current BCO?
```c++
	const TrkrDefs::hitsetkey hitsetkeymask = MvtxDefs::genHitSetKey(layer, stave, chip, 0); 
```

osbornjd:
I was thinking about this too... I think we do need to generate noisy pixels for every strobe, so that we are sure that the same pixel is masked in every strobe window in the offline. That seems really redundant as we'd have to iterate through all possibilities for the strobe and check that the hitsetkey is not masked (5 bits for the strobe window -> 32 possible strobes, so for each hit iterating 32 times and masking each strobe is quite a lot). Or the hitsetkey for each 32 possible strobe has to be stored in the map, and then that is also a redundancy. Perhaps we can reduce this based on whatever the strobe window length is and corresponding streaming window is (e.g. for a 5 mus strobe window in normal beam conditions we would only expect hitsetkeys up to +/-1 strobe offset from the nominal trigger in 7 mus streaming).

cdean-github:
> I was thinking about this too... I think we do need to generate noisy pixels for every strobe, so that we are sure that the same pixel is masked in every strobe window in the offline. That seems really redundant as we'd have to iterate through all possibilities for the strobe and check that the hitsetkey is not masked (5 bits for the strobe window -> 32 possible strobes, so for each hit iterating 32 times and masking each strobe is quite a lot). Or the hitsetkey for each 32 possible strobe has to be stored in the map, and then that is also a redundancy. Perhaps we can reduce this based on whatever the strobe window length is and corresponding streaming window is (e.g. for a 5 mus strobe window in normal beam conditions we would only expect hitsetkeys up to +/-1 strobe offset from the nominal trigger in 7 mus streaming).

Here's a thought to kill two bird with one stone. If we switch from an std::map where the key is the HitSetKey and the HitKey to an std::set where the key is the layer, stave and chip, and the mapped value is a vector of pairs, where the pair is the row and column then you now have a pixel mask independent of the strobe information (we know it's a constant map so that's fine).

Now, you already calculated the strobe index:
```c++
    uint64_t bcodiff = gl1bco - strobe;
    double timeElapsed = bcodiff * 0.106; // 106 ns rhic clock
    int index = std::floor(timeElapsed / m_strobeWidth);
```
so if you compare the layer, stave and chip from the HitSetKey of the actual hit then you don't need the strobe information. You then just use the strobe index as needed

cdean-github:
Wait, ignore my thoughts. Zhaozhong already took care of this as he regenerates a HitSetKey with a strobe value of 0 as the pixel mask is independent of the strobe so this should naturally mask the pixels for each strobe

MYOMAO:
OK I will work on this too. In the meantime, I will update CDBTree file using channel

osbornjd:
I cleaned up the memory leak and a few other things, let's see if it passes Jenkins now and then we can merge this.

MYOMAO:
I have also updated the input hot MVTX pixel map with a new CDBTree format using channels so that the size of the root file is 50 kB instead of 4.5 MB. I will make another commit later on!

PR #2471, opened 2024-02-01
Add ability to enter a "probe" into the jet tree to cluster with jets

david-stewart:
Example:
  In Fun4All.C macro, add it to the tree:

  JetProbeMaker* probe_maker = new JetProbeMaker();
  probe_maker->set_pt(35.);  // defaults to 30. GeV
  probe_maker->set_abs_eta(0.8); // defaults to 0.7
  // etc...
  se->registerSubsystem(probe_maker);

  Then cluster the proble with jet inputs with JetProbeInput() which
  knows where to find the Probe JetContainer (always containing a single
      jet) on teh node tree:

  JetReco *itertowerreco = new JetReco();
  itertowerreco->add_input(new JetProbeInput()); // add the probe to be clustered with the jets
  itertowerreco->add_input(new TowerJetInput(Jet::CEMC_TOWERINFO_SUB1));
  itertowerreco->add_input(new TowerJetInput(Jet::HCALIN_TOWERINFO_SUB1));
  itertowerreco->add_input(new TowerJetInput(Jet::HCALOUT_TOWERINFO_SUB1));

  Then in the analysis, can also use the JetProbeInput to fetch the
  probe values like:

  JetProbeInput probe(topNode);
  float probe_pt  = probe.pt; // put the probe
  float probe_eta = probe.eta;
  float probe_phi = probe.phi;

  Final note:
  The probe is put on the node tree under a JetContainer named
  ``JetProbeContainer'', and in each event it adds a single
  Jet, with the pT, phi, and eta kinematics randomly chosen
  but the user supplied ranges (defaults to pT=30GeV, any
      azumuth, and phi uniform in |eta|<0.7.
  The generated jet has a single component of type Jet::JET_PROBE, and index of 0.

## Types of changes

## TODOs (if applicable)


pinkenburg:
you can ignore the clang-tidy failure

offline/packages/jetbase/JetProbeMaker.cc
pinkenburg, line 1: We try to stay away from TRandom3, please use the gsl random number generator. The setting of the seed to PHRandomSeed() is missing, results produced here will not be reproducible

david-stewart: OK, I have made the changes, but I am missing something in the `Makefile.am` to get to to link to the proper libraries to make the calls to `gsl_rng_set`, `gsl_rng_uniform` and `gsj_rng_mt19937`.
Is it worth it to add a new dependency? If the extra linkage weight isn't worth it, can I still use `PHRandomSeed()` to set the ROOT TRandom3?

david-stewart: ok! I have looked at the Makefile.am for g4mvtx, but I cannot seem to see the difference that adds the gsl libraries for the linker

pinkenburg: likely gsl is resolved by its dependency on other libraries which themselves include a dependency on libgsl. You can run an ldd on the libg4mvtx.so to verify that. It doesn't really matter how the mvtx resolves this - if the test linker tells you that you need to add libgsl or others you have to do that

david-stewart: That makes sense. I'm just not sure where the libraries to link to are located, and thought that g4mvtx would have a hint because they use them

pinkenburg: ahh - okay. The 3rd party software installations are copied to $OFFLINE_MAIN, so you'll find those under $OFFLINE_MAIN/lib which is already in your Makefile.am

david-stewart:
OK, I have used this new version of the code and the results are as expected. It doesn't use TRandom3, but rather the gsl libraries. From my end, it is ready to merge.

PR #2473, opened 2024-02-02
Implementing plane shifting 

eumaka:
PR implements Equation (6) of arxiv:nucl-ex/9805001. Some runs require recentering in addition to shifting. In particular some of the zero field runs. See plot for implementation result.


eumaka:
Earlier comment should say: some runs require shifting in addition to recentering. Recentering comes first. Can this be merged so we get it into production @pinkenburg @blackcathj @jdosbo

pinkenburg:
Was the EventPlaneInfo saved on any DST previously. The change in EventPlaneInfov1 will render those DSTs unreadable

eumaka:
It was not previously stored in a DST

PR #2488, opened 2024-02-14
INTT Z clustering fix

petermic:

Enables nontrivial INTT Z clustering by switching from a hard-coded cluster Z size of 1 to the value calculated in the clusterizer. Under default running conditions, Z clustering is switched off, and there is no change to that pathway.

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/packages/intt/InttClusterizer.cc
+++ b/offline/packages/intt/InttClusterizer.cc
@@ -547,8 +547,8 @@ void InttClusterizer::ClusterLadderCells(PHCompositeNode* topNode)
       {
         phierror *= scalefactors_phi[2];
       }
-      // z error. All clusters have a z-size of 1.
-      const float zerror = length * invsqrt12;
+      // z error.
+      const float zerror = zbins.size() * length * invsqrt12;
 
       double cluslocaly = NAN;
       double cluslocalz = NAN;
@@ -572,7 +572,7 @@ void InttClusterizer::ClusterLadderCells(PHCompositeNode* topNode)
       clus->setPhiError(phierror);
       clus->setZError(zerror);
       clus->setPhiSize(phibins.size());
-      clus->setZSize(1);
+      clus->setZSize(zbins.size());
       // All silicon surfaces have a 1-1 map to hitsetkey.
       // So set subsurface key to 0
       clus->setSubSurfKey(0);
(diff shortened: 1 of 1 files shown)


osbornjd:
This seems fine to me, is this needed for the dNdeta analysis? How do you plan to actually do the z clustering?

petermic:
> This seems fine to me, is this needed for the dNdeta analysis? How do you plan to actually do the z clustering?

The issue was encountered while working on the dNdEta analysis; as part of investigating some irregularities with INTT clusters in data, it was found that the _make_z_clustering switch in the clusterizer doesn't actually do what it says it does. This fixes that.

There are actually no changes to the Z clustering algorithm itself -- it uses an adjacency graph in a manner basically identical to the MVTX (where the _make_z_clustering switch controls whether adjacency in the Z direction adds an edge to the graph). The only change is that the result of the Z clustering is no longer ignored when making clusters if the _make_z_clustering switch is turned on.

In any case, to explain why this is a draft PR, this is still pending review and approval from INTT experts.

osbornjd:
Right, I see there are no changes to the clustering itself - just was curious to know the use case of this. In any case, if an INTT expert would like to review it then we can wait on that

ChengWeiShih:
Hello, 

Here is Cheng-Wei. 
I just want to confirm, I remember there is a switch boolean that the users can decide whether the cluster z size is fixed at 1 or not. An I correct ?  

Another thing is, if the clustering algorithm in the z direction is the same as that of MVTX, which means the diagonal adjacent is also accepted. I think INTT cluster may need a more tight requirement.

petermic:
> Hello,
> 
> Here is Cheng-Wei. I just want to confirm, I remember there is a switch boolean that the users can decide whether the cluster z size is fixed at 1 or not. An I correct ?
> 
> Another thing is, if the clustering algorithm in the z direction is the same as that of MVTX, which means the diagonal adjacent is also accepted. I think INTT cluster may need a more tight requirement.

Yes, that's what _make_z_clustering does. It's a `std::vector<bool>` that's filled in the macro `common/Trkr_Clustering.C` with the settings for each layer. (As such, there are no default values for this parameter within the module itself -- it must always be initialized after construction.) The default version of `common/Trkr_Clustering.C` has this parameter set to `false` for all four layers. 

As for the definition of adjacency, I'm in the middle of investigating whether it's an issue or not in the INTT; that's a slightly larger change that might be better off in a separate PR.

ChengWeiShih:
Hello Michael, thanks for the explanation. It looks good to me!

PR #2496, opened 2024-02-16
G4mvtx float to double

pinkenburg:

## Types of changes
- [ X] Bug fix (non-breaking change which fixes an issue)

The mvtx g4 description was a bad mix of floats and doubles which then leads to math issues (compiling without optimizer gives different results from enabling the optimizer). The current code gives identical results for both cases but the results will be different from previous runs

## TODOs (if applicable)


pinkenburg:
Qualitatively the qa histos are fairly identical. Given that going consistently to doubles handling gives the same results with optimized and unoptimized code, this version is more accurate than the previous versions

osbornjd:
Do these changes/fixes potentially affect the material map or distribution in any way in the mvtx volumes? If so this would require that we update the Acts material map.

pinkenburg:
There are minor changes - what in the end made a difference was the results of asin and acos, so the placement of the support structures is slightly different but the actual effect is somewhere in the later digits. The root cause is that when using floats with (C) math functions the argument gets converted to double and the calculation is done in double precision. When "using namespace std" or the std:: version this conversion is not done and it is strictly float (which is faster but less accurate). Now everything is using double, so these numerical inconsistencies are gone. It's just that even the tiniest change in a g4 volume location leads to a quickly snowballing change in the g4hits

PR #2512, opened 2024-02-26
corrected ADC

HataMisaki:
corrected ADC
- because all ADC value was 0 before.


osbornjd:
I sent you an invitation to join the collaboration github page, please accept it and then we can run the CI on this

PR #2564, opened 2024-03-15
add vertex info the event plane reco

eumaka:
Talked to Chris, we need to produce the event plane reco now. Ok :) This adds some additions from a local version- the z-vertex cuts. This is already ok for mass production. Keep in mind that the corrections will also be done as a function of z-vertex (testing in progress for optimal width, run by run variations etc...).


pinkenburg:
instead of TString::Form()
Form("tprof_mean_cos_south_mbd_order_%d", order)
use (until std::format is available)
#include <boost/format.hpp>

boost::str(boost::format("tprof_mean_cos_south_mbd_order_%d") %order).c_str()

PR #2567, opened 2024-03-16
Fix boost::geometry::index intersect call

david-stewart:
There was a small bug in PHCA seeding in which the call to find neighboring clusters across the periodic phi boundaries were not found. This is fixed. The effect is quit small, as it is relatively rare to have the phi search across the boundary.

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackreco/PHCASeeding.cc
+++ b/offline/packages/trackreco/PHCASeeding.cc
@@ -224,11 +224,21 @@ Acts::Vector3 PHCASeeding::getGlobalPosition(TrkrDefs::cluskey key, TrkrCluster*
 
 void PHCASeeding::QueryTree(const bgi::rtree<pointKey, bgi::quadratic<16>> &rtree, double phimin, double etamin, double lmin, double phimax, double etamax, double lmax, std::vector<pointKey> &returned_values) const
 {
-  double phimin_2pi = phimin;
-  double phimax_2pi = phimax;
-  if (phimin < 0) phimin_2pi = 2*M_PI+phimin;
-  if (phimax > 2*M_PI) phimax_2pi = phimax-2*M_PI;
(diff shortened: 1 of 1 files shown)


osbornjd (commented):
Is this what is responsible for this small drop in efficiency at the azimuthal periodic boundary that we have seen before? As a part of this PR can you make a plot of the efficiency as a function of phi for before and after? Maybe this was communicated in the tracking meeting and I missed it...

pinkenburg:
The clang-tidy errors are fixed in Ejiros PR (which is why they show up here), no problem for this PR

osbornjd:
By the way I'm not objecting to this as it does look like a bug, would just like to try to tie it to something we know exists already... You can see it in single particle events so it should be easy to test

david-stewart:
> Is this what is responsible for this small drop in efficiency at the azimuthal periodic boundary that we have seen before? As a part of this PR can you make a plot of the efficiency as a function of phi for before and after? Maybe this was communicated in the tracking meeting and I missed it...

Hi @osbornjd, this is a good idea and question. I haven't actually presented this anywhere else. I have just done the internal "proof" of putting in printout statements of how many clusters the Boost rtree found whenever we adjusted the phi range for the periodic boundaries. In my small test, there were never any neighbors found when the phi window was adjusted the previous way, but there were often some clusters with the fix. So with that confidence, I submitted.

For the study, I'll probably take the route of least resistance and get the macros from Derek and see what the result on efficiency is. I haven't yet found why the CA seeder is skipping obvious tracks, and want to spend most of my time there.

osbornjd:
It should be easy to test - just run single pions with this change and with the default build. I don't think any special macros are needed, and it should be a relatively quick check.

david-stewart:
The fix works as expected. These are just the distribution of phi in G4_tracks (10 different 10 GeV pi- particles per event) with a default TTree->Draw() command. The PR fixes the dip around phi=0.

[PR_effect_on_phi.pdf](https://github.com/sPHENIX-Collaboration/coresoftware/files/14723906/PR_effect_on_phi.pdf)

PR #2604, opened 2024-04-08
Phtpcresiduals update

hupereir:
This consolidates PHTPCResiduals to get more robust distortion correction maps. 

- extrapolate from any track state rather than only the one at beginning
- check from not-a-number in matrix calculation. one could also track down the reasons for nan in the first place, but since any single occurence, no matter how rare, breaks the whole matrix calculation (and all processed past events), better keep the check in place.

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/packages/tpccalib/PHTpcResiduals.cc
+++ b/offline/packages/tpccalib/PHTpcResiduals.cc
@@ -49,9 +49,9 @@
 #include <iostream>
 #include <sstream>
 
-namespace 
+namespace
 {
-  
+
   // square
   template<class T> inline constexpr T square( const T& x ) { return x*x; }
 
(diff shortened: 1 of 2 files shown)


hupereir:
The clang error is unrelated.
The clang-tidy warning are missing parenthesis after if statement, which running clang-tidy locally cannot find. 
(clang-tidy --checks="-*,hicpp-braces-around-statements" PHTpcResiduals.cc, returns nothing.)
They are harmless anyway (they are all single line if statement)
This should go in.

osbornjd:
the magic clang-tidy fix command should put brackets around those, I'm not sure why it doesn't... in any case I think we should put brackets around them even though it doesn't change functionality, for the purpose of maintainability.

hupereir:
> the magic clang-tidy fix command should put brackets around those, I'm not sure why it doesn't... in any case I think we should put brackets around them even though it doesn't change functionality, for the purpose of maintainability.
In general, I agree. However: without the ability to 'fix' this automatically with the command above, doing it manually is very tedious: 

- first because there are 8 instance already, most of them unrelated to the actual change ... 
- first because clang-tidy might found more on the second pass (has happened in the past),

I don't think this should prevent this fix to be merged.

PR #2608, opened 2024-04-09
(INTT) Lots of changes/cleanup/some fixes to InttCombinedRawDataDecoder

josephbertaux:
* Fixed time bucket computation in InttCombinedRawDataDecoder
* Introduced general handles for configuring classes that should be loaded from CDB
* Added a base class for these classes (InttLoadable)
* Removed unnecessary versioning from classes that will not be written directly to ROOT files
* Some renaming and implementation cleanup
* Additionally changed Makefile.am so the ONLINE case uses some of the new classes (InttMap, InttBadMap, InttFeeMap, InttLoadable, InttSurveyMap)
    * If the ONLINE build fails, you can reduce that case to what it was before (only InttFelixMap and InttMapping) and I'll fix it and change it later in a separate PR

## Types of changes

## TODOs (if applicable)


pinkenburg:
here is a little trick how to make an empty commit to trigger jenkins without having to change a thing:
https://wiki.sphenix.bnl.gov/index.php/Git#empty_commit_to_trigger_jenkins

josephbertaux:
--allow-empty is pretty useful, ty

josephbertaux:
I'm trying to reproduce the error locally, but the main Fun4All_TrkrHitSet_Unpacker.C uses
```C++
#include <Trkr_Clustering.C>
```
as opposed to `#include "../relative/path/to/Trkr_Clustering.C"`, so running the macro locally uses something from the newer build and not my modified local copy. I can modify it and do a `git ignore` on it, but is there a way to set up my environment so it will look for my local copy by default? (I've already tried the local modification and have been able to reproduce the state Jenkins reaches--I'll either edit it so it matches or stash and switch branches and delete my macros feature branch)

I'd rather not accidentally commit something that shouldn't be committed, or fool myself thinking I've updated my fork doing a git/fetch/merge and it's still ignored, and if for whatever reason I actually need to edit it, I'd have to know how to do this anyway.

josephbertaux:
I think I may have figured it out--in the build Jenkins is using, the payloads I'm trying to get by default aren't in the build's CDB.

I added default behavior to `InttCombinedRawDataDcoder` look up `INTT_BCOMAP`, `INTT_HotMap`, `INTT_DACMAP`, and `InttFeeMap` during its `InitRun` method.

With some print statement debugging, I've realized `CDBInterface::getUrl("INTT_BCOMAP")` returns `""` when the macro runs. I can see it in the CDB now, but what you added may not be accessible with the timestamp the macro uses.

I'll add additional guard clauses to my base class to catch this and exit, and change it so it doesn't look up these classes by default. However, I do need to be able to find some version of `InttFeeMap` for any timestamp (otherwise we won't be able to associate our hardware readouts with our sensors).

Once I add these additional guard statements, how can I flag/throw/exit so being unable to find things in the CDB is detectable at the Jenkins level? Is `gSystem->Exit(1);exit(1);` sufficient, or should I throw an exception? B/c while some things aren't needed for every run, other things might be and it's good to know if a PR or new build breaks that access.

offline/packages/intt/InttLoadable.h
pinkenburg, line 1: abstract classes have caused problems in the past when loading them in root macros. Given that this code will only be used in the intt software - the enforcement mechanism is not really needed

josephbertaux: I'll change it back. Also, when I ran the build from macros, everything looked good except the valgrind check--I don't dynamically allocate anything in a loop except for new TrkrHits, but that part of the code has been unchanged for months.

Is the memory leak something I need to try to fix before this gets merged, or could that be due to another part of the build?

pinkenburg: looking at the last valgrind run - the job crashes because of a macro problem, so any memory leak messages are meaningless

PR #2611, opened 2024-04-10
fix: remove unused functions from TrackSeed_v2

osbornjd:

## Types of changes

Removes leftover verbosity from some previous PR

## TODOs (if applicable)


hupereir:
Joe,
My understanding is that Tony added those on purpose, to make sure that the old interface to get_px and get_py is not used any more, and that the new one is used instead, with internal m_phi azimuthal angle.
Using both interface might lead to inconsistencies (different px, py) in the code, and between how distortion corrections are accounted for. 
I would leave the printout unchanged. (or crash the code)

hupereir:
See: https://github.com/sPHENIX-Collaboration/coresoftware/pull/2606,
which I would not have figured out without the printout.

pinkenburg:
In my understanding this renders old DSTs useless, so backward compatibility is not an issue here. How about just removing the old methods? Then the compiler takes care of flagging "illegal" uses.

hupereir:
> In my understanding this renders old DSTs useless, so backward compatibility is not an issue here. How about just removing the old methods? Then the compiler takes care of flagging "illegal" uses.

- indeed: new reconstruction code cannot handled old trackseed_v1 any more (because of not re-calculating phi on the fly any more, which was done incorrectly anyway)
- you can still read and reconstruct old DSTs using past ana build though.
- no objection against removing the old methods, but I'd leave that to Joe/Tony

hupereir:
Side remark: in general, I would advise against having algorithm (such as circle_fit, phi fit) methods inside an interface class (TracksSeed). I'd recommend moving those to an helper class instead that takes the same arguments (+ the trackseed) to do the work. This will simplify scheme evolution.

adfrawley:
Sorry, I missed this exchange because I was lecturing.

Yes, I added the printout to v2 to alert when the old methods (which get the global position for the default geometry cluster position, instead of being given corrected global positions) were being called for v2.

I was under the impression that v1 would still work because the methods in v1 were not modified.

Is the suggestion to remove the old methods from all versions then?

Tony

osbornjd:
I think the suggestion is to remove the old methods in v2, which I can do

offline/packages/trackbase_historic/TrackSeed_v2.cc
```diff
@@ -364,7 +362,6 @@ void TrackSeed_v2::circleFitByTaubin(TrkrClusterContainer* clusters,
                                      uint8_t startLayer,
                                      uint8_t endLayer)
 {
-  std::cout << "Call to old circle fit" << std::endl;
```
osbornjd, line 367: Okay, I'm now confused. Tony's PR did not introduce any new functionality for this wrapper to call the circle fit function in TrackFitUtils, but there is a print statement here to an old circle fit call. I assume this is old because this function does not consider the distortions, it just calls the usual `getGlobalPosition` function. What was this functionality replaced by? @adfrawley

sPHENIX-GitHub-l:
That functionality goes away. There is no undistorted data, so that circle fit is not applicable. The method that takes a list of global positions is used for everything.

Tony

osbornjd:
Okay, so the expected API is now that when calling the circle fit you have already determined the global positions beforehand and applied any distortion corrections you need. Is this centralized in some way, i.e. so that we don't rely on applying the distortion corrections in each module each time we need to call a circle fit?

osbornjd:
Looking more carefully it appears that is not the case, so I'm confused about how this is expected to be used. Example: in the KFProp module the global position key map is filled [here](https://github.com/sPHENIX-Collaboration/coresoftware/blob/cb345f9d95f03b5f0b9d39e8dd824c4b824c2204/offline/packages/trackreco/PHSimpleKFProp.cc#L466-L469) and then called to circle fit [here](https://github.com/sPHENIX-Collaboration/coresoftware/blob/cb345f9d95f03b5f0b9d39e8dd824c4b824c2204/offline/packages/trackreco/PHSimpleKFProp.cc#L282), but there are no distortion corrections applied. It looks like in your PR you just changed all of the `TrackSeed_v1` to `TrackSeed_v2` without actually changing any of the calls to the circle fitting - is my reading correct? Was that left for further development?

adfrawley:
I thought about setting it up so that whenever the circle fit is called it updates the persistent phi value.  The problem is that the circle fit is called in PHCASeeding in situations where we do not want to set phi.

There are only a couple of places in the code where the persistent phi value needs to be set.  It is OK as it is, I think.

Tony


adfrawley:
Hi Joe,

I added new methods in v2 to return anything that depends on phi (such as pT) from the persistent stored phi. value

The initial seed determination in KFProp is done without distortion corrections.  It has to be, we do not know the bunch crossing for the clusters. The seed is updated after distortion corrections, if in triggered mode. If in pp mode, the seed fit is updated  later in the PrelimDistortionCorrection module after a refit of the (preliminary) corrected cluster positions using the ALICE KF.

Maybe we should talk offline.

Tony

osbornjd:
Okay, I removed the old functions from `TrackSeed_v2` and updated the trackreco code that depended on those accordingly. Should be good to go now

offline/packages/trackbase_historic/TrackSeed_FastSim_v2.cc
```diff
@@ -46,8 +46,7 @@ void TrackSeed_FastSim_v2::identify(std::ostream& os) const
   return;
 }
 
-float TrackSeed_FastSim_v2::get_phi(TrkrClusterContainer* /*clusters*/,
-                                    ActsGeometry* /*tGeometry*/) const
```
hupereir, line 50: @osbornjd I don't think this works.
get_phi without arguments overloads the accessor to m_phi from TrackSeed_v2. Which is not what we want.
We want to calculate phi 'once' (or twice), assign it via set_phi, to m_phi member variable, and then keep using this one via subsequent calls to get_phi (no arguments).

hupereir: We should carefully think this through before doing any more commits to the TrackSeed class familly I think, which is quite a mess as it is now ...

osbornjd:
I misread and thought that the phi was actually set in the truth track seeding modules. I left the track sim objects as is and just fixed the calls to px and py which now call the base class

hupereir:
> I misread and thought that the phi was actually set in the truth track seeding modules. I left the track sim objects as is and just fixed the calls to px and py which now call the base class

Yep. Looks good, it seems. Thanks !

osbornjd:
Somehow I thought this PR was merged but it was not, and I inadvertently force pushed (lesson learned, always double check before you force push)

adfrawley:
In the tracking QA, the INTT efficiency is substantially lower than the reference here. Is that from this PR, or was it already that way?

Tony

osbornjd:
It's from the merging of the intt survey geometry. The track momentum resolution is a bit worse because there are fewer intt measurements on each track due to (I suspect) the reference radii changing significantly. It's on my to-do list to dig into, hence I have left the QA reference alone so that we do not lose track of what the previous performance was. Jin just told me the reference root files are stored so I may reset it and just refer to the old reference file while I'm working on this.

PR #2619, opened 2024-04-13
MVTX standalone event combining

cdean-github:

## Types of changes

When we run the event combiner for the MVTX and no other detectors, all hits are deleted. This is due to a check on the reference BCO and the BCO range to look for hits in. All values are unsigned ints and the reference BCO defaults to 0. When we subtract the range from the reference, we wrap around to almost the max value of and unsigned 64-bit int.

Also, in the unpacker, there's a check for the GL1 so we can figure out the MVTX time bucket. There's no GL1 in standalone mode so flags were put in to avoid this

## TODOs (if applicable)

### DIFF
--- a/offline/framework/fun4allraw/Fun4AllStreamingInputManager.cc
+++ b/offline/framework/fun4allraw/Fun4AllStreamingInputManager.cc
@@ -685,16 +685,19 @@ int Fun4AllStreamingInputManager::FillMvtx()
   {
     select_crossings += m_MvtxRawHitMap.begin()->first;
   }
+
+  uint64_t ref_bco_minus_range = m_RefBCO < m_mvtx_bco_range ? 0 : m_RefBCO - m_mvtx_bco_range;
   if (Verbosity() > 2)
   {
     std::cout << "select MVTX crossings"
-              << " from 0x" << std::hex << m_RefBCO - m_mvtx_bco_range
+              << " from 0x" << std::hex << ref_bco_minus_range
               << " to 0x" << select_crossings - m_mvtx_bco_range
               << std::dec << std::endl;
   }
   // m_MvtxRawHitMap.empty() does not need to be checked here, FillMvtxPool returns non zero
   // if this map is empty which is handled above
-  while (m_MvtxRawHitMap.begin()->first < m_RefBCO - m_mvtx_bco_range)
+  //All three values used in the while loop evaluation are unsigned ints. If m_RefBCO is < m_mvtx_bco_range then we will overflow and delete all hits
+  while (m_MvtxRawHitMap.begin()->first < ref_bco_minus_range)
   {
     if (Verbosity() > 2)
     {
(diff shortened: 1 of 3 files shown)


osbornjd (commented):
I think setting the gl1 bco and more importantly the bco difference to 0 when running standalone is fine. That is the same as setting the bco difference to that of the triggered crossing which, in standalone running, is effectively the case. This will give you the "default" set of hitsetkeys, so I think this is what you want.

cdean-github:
> I think setting the gl1 bco and more importantly the bco difference to 0 when running standalone is fine. That is the same as setting the bco difference to that of the triggered crossing which, in standalone running, is effectively the case. This will give you the "default" set of hitsetkeys, so I think this is what you want.

Hi Joe, 

Thanks. I assume this is in reference to the comment I left in the code? I also assumed that this is the case we want, and I think Yasser had a similar opinion to me. I wanted to leave it there maybe for discussion between us all, maybe there's a case I'm not considering where we would want to have the time bucket info.

I also need to think about how we want to handle the strobe window information. 89 us is what we use normally for cosmics but we took data 2 nights ago at 10 us and 5 us. The best idea is probably to have some CDB info on strobe length or add it to the raw event header information

pinkenburg:
I have no problems with the standalone but it should be possible to figure this out via the missing GL1. Using a flag just makes this error prone. It's guaranteed that somebody will either forgot to set it or set it when running with the gl1 which might then cause hard to recognize event alignment problems.
If the strobe information is in the raw data and needed later we might want to put this info under the RUN node. Then it stays with the processed data and modules which need this info can pick it up from the dst (we can still put it into the cdb as part of the raw data processing if we want to)

cdean-github:
> I have no problems with the standalone but it should be possible to figure this out via the missing GL1. Using a flag just makes this error prone. It's guaranteed that somebody will either forgot to set it or set it when running with the gl1 which might then cause hard to recognize event alignment problems. If the strobe information is in the raw data and needed later we might want to put this info under the RUN node. Then it stays with the processed data and modules which need this info can pick it up from the dst (we can still put it into the cdb as part of the raw data processing if we want to)

OK, I can think about this a bit more but off the top of my head, I can alter the initial check for the GL1 run node, and instead of aborting event if it doesnt exist then I can set the time bucket to 0. It will have the same effect and I can put a very high verbosity flag to tell the user that they're running with no GL1 info

pinkenburg:
Sounds good, Thanks

osbornjd:
> 
> Hi Joe,
> 
> Thanks. I assume this is in reference to the comment I left in the code? I also assumed that this is the case we want, and I think Yasser had a similar opinion to me. I wanted to leave it there maybe for discussion between us all, maybe there's a case I'm not considering where we would want to have the time bucket info.

Yes sorry, I should have clarified that this was in reference to the comment in the code.

> I also need to think about how we want to handle the strobe window information. 89 us is what we use normally for cosmics but we took data 2 nights ago at 10 us and 5 us. The best idea is probably to have some CDB info on strobe length or add it to the raw event header information

I agree we will need this somewhere. Is it something that can be stored in the DAQ db?

PR #2659, opened 2024-04-23
Laser clusterizer and matching

bkimelman:

## Types of changes

New clustering for diffuse laser flashes using rtree structure, this allows clustering to be performed across pad rows since laser hits should span at least two pad rows.
Additional changes to matching of reco diffuse laser clusters to truth pattern (will be updated to match with survey data shortly). This new method implements a new global R matching scheme that does a nearest-neighbor matching for each reco radial peak, then shifts everything over in each direction and finds the minimum total residual between all peaks in a given shift

## TODOs (if applicable)


offline/packages/tpc/LaserClusterizer.h
```diff
@@ -60,7 +60,7 @@ class LaserClusterizer : public SubsysReco
   //void remove_hits(std::vector<pointKeyLaser> &clusHits,  bgi::rtree<pointKeyLaser, bgi::quadratic<16> > &rtree, std::multimap <unsigned int, std::pair<TrkrDefs::hitkey,TrkrDefs::hitsetkey>> &adcMap, std::multimap <unsigned int, float*> &adcCoords);
   void remove_hits(std::vector<pointKeyLaser> &clusHits,  bgi::rtree<pointKeyLaser, bgi::quadratic<16> > &rtree, std::multimap <unsigned int, std::pair<std::pair<TrkrDefs::hitkey,TrkrDefs::hitsetkey>, std::array<int, 3>>> &adcMap);
 
-  void set_debug_name(string name){ m_debugFileName = name; }
+  void set_debug_name(const std::string name){ m_debugFileName = name; }
```
hupereir, line 63: should be (const std::string& [...]) to avoid superfluous string copy

offline/packages/tpc/LaserClusterizer.cc
```diff
@@ -336,7 +345,7 @@ int LaserClusterizer::process_event(PHCompositeNode *topNode)
 	    float x = r * cos(phi);
 	    float y = r * sin(phi);
 	    float z = m_tdriftmax * m_tGeometry->get_drift_velocity() - zdriftlength;
-	    if (side == 0){
+	    if (side == 1){
```
hupereir, line 348: I am a bit worried about these sign flips appearing a multiple places in the code. Could we not have that in a centralize place (for instance as some inline function in TPCDefs ? ) (cc @blackcathj, @bogui56 )
Also I think it would make it much easier for readabilaty and future maintenance to use an enumeration for the side, defined in TPCDefs also, such as: 
Enum 
{
SideNorth == 0,
SideSouth == 1
};
(or the opposite), and use this instead of 0 or 1 ?

hupereir (commented):
Hi Benjamin, 
two comments on your latest commit, following private discussion from yesterday. 
I hope this helps, 
Hugo

offline/packages/tpc/LaserClusterizer.cc
bkimelman: This is a fair point and could help resolve some of the inconsistencies we've been seeing. I'll leave the centralization to Jin and Christof though

offline/packages/tpc/LaserClusterizer.h
bkimelman: Good point, will fix

PR #2693, opened 2024-05-01
reconstruct SMD, update sEPD masking

eumaka:
Changes:

For ZDC-SMD: reconstruct all 52 channels
For sEPD: Mask channels not instrumented using updated 2024 map

### DIFF
--- a/offline/packages/CaloBase/TowerInfoContainerv1.cc
+++ b/offline/packages/CaloBase/TowerInfoContainerv1.cc
@@ -31,7 +31,7 @@ TowerInfoContainerv1::TowerInfoContainerv1(DETECTOR detec)
   }
   else if (_detector == DETECTOR::ZDC)
   {
-    nchannels = 16;
+    nchannels = 52;
   }
   _clones = new TClonesArray("TowerInfov1", nchannels);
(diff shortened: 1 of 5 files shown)


pinkenburg:
The reading for 2023 zdc data in the Fun4All_Year1.C macro fails

eumaka:
If you are running it over 2023 data it will fail at calibration state: 
/home/phnxbld/sPHENIX/gcc-12.1.0/ana/source/coresoftware/offline/database/cdbobjects/CDBTTree.cc:691:  Could not find channel 16 for zdc_calib in float calibrations

Remember what's in the CDB, is a file with 16 channels

eumaka:
I just made another calibration file with 52 channels (will be 0 for the extra channels (SMD+Veto)), and ran a private DST. It works fine for the 2023 data. Once the new calibration file is in the CDB, it should be good to go.

pinkenburg:
The problem is that we need to run a towerbuilder over the 2023 data. The current Fun4All_Year1.C macro crashes - that is a problem which has to be fixed. It might mean that this macro needs a change but whatever we do - we have to preserve the ability to process run 2023 data

PR #2697, opened 2024-05-01
Add general tracks-in-jets QA module

ruse-traveler:
This PR adds a QA module which accumulates histograms for tracks and their constituent clusters and hits both in jets and inclusively. The code is designed to be flexible with respect to inputs (it will work for jets made from tracks, particle flow elements, calorimeter towers/clusters, or any combination of them) and easy to extend with respect to outputs (additional histograms are easily generated for different populations of jets, tracks, etc.).

Furthermore, the module can be run both as a plain `SubsysReco` with output being written to a specified file, or with histograms being handled by the QA Histogram Manager.

## Types of changes
- [X] New feature (non-breaking change which adds functionality)

This PR adds a new feature, and the module was developed as part of the Jet Structure QA workfest. A macro illustrating usage of the module and a README with further information about the structure of the code can be found in the analysis repo at:

https://github.com/sPHENIX-Collaboration/analysis/tree/master/JS-Jet/TrackJetQAMaker

## TODOs (if applicable)

The exact suite of histograms and which populations of jets, tracks, clusters, and hits to be written out will need to be decided on by the Jet Structure and Tracking Software groups.

There are also some clear areas for improvement (most of which are commented on in the code):
 - Matching non-constituent tracks to jets is currently done geometrically based on the specified resolution parameter, but this should be done either using the actual jet area (which isn't necessarily pi*R_{jet}^2) or by matching tracks to constituent  calorimeter towers/clusters;
 - Raw pointers are used in several places which should be smart pointers;
 - And additional debugging output should be added to the components outside of the top-level Fun4All module.

### DIFF
--- a/offline/QA/Jet/Makefile.am
+++ b/offline/QA/Jet/Makefile.am
@@ -39,11 +39,13 @@ libjetqa_la_SOURCES = \
   TrksInJetQAJetManager.cc
 
 libjetqa_la_LIBADD = \
+  -L$(libdir) \
+  -L$(OFFLINE_MAIN)/lib \
   -lcalo_io \
-  -lcalotrigger_io \
   -lfun4all \
   -lphool \
   -lg4dst \
+  -lg4eval \
   -lSubsysReco \
   -lqautils
 
--- a/offline/QA/Jet/Makefile.am
+++ b/offline/QA/Jet/Makefile.am
@@ -14,14 +14,36 @@ AM_CPPFLAGS = \
   -isystem$(ROOTSYS)/include
 
 pkginclude_HEADERS = \
+  TrksInJetQA.h \
+  TrksInJetQAHist.h \
+  TrksInJetQATypes.h \
+  TrksInJetQAConfig.h \
+  TrksInJetQABaseFiller.h \
+  TrksInJetQAInJetFiller.h \
+  TrksInJetQAInclusiveFiller.h \
+  TrksInJetQABaseManager.h \
+  TrksInJetQAHitManager.h \
+  TrksInJetQAClustManager.h \
+  TrksInJetQATrkManager.h \
+  TrksInJetQAJetManager.h
 
 libjetqa_la_SOURCES = \
+  TrksInJetQA.cc \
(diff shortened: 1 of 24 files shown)


osbornjd:
I added this new directory to our package list so on your next push the CI should actually test it

ruse-traveler:
> I added this new directory to our package list so on your next push the CI should actually test it

Thanks! Are the cpp-check failures related to that?

osbornjd:
The cppcheck failures should be real as that just runs over whatever code is there, no build process required

ruse-traveler:
Gotcha! Let me check where those are coming from then...

pinkenburg:
It's the dependency on libg4dst which crashes the build. libg4dst is a convenience library which allows us to just resolve everything which is on our DSTs (and is therefore an overkill for modules which use a subset). It is build last and it is not available when this package is linked. It doesn't show up if you build it using an existing build (which contains libg4dst). Remove it from youe Makefile.am and see which symbols are not resolved and add their (_io) libraries.

ruse-traveler:
> It's the dependency on libg4dst which crashes the build. libg4dst is a convenience library which allows us to just resolve everything which is on our DSTs (and is therefore an overkill for modules which use a subset). It is build last and it is not available when this package is linked. It doesn't show up if you build it using an existing build (which contains libg4dst). Remove it from youe Makefile.am and see which symbols are not resolved and add their (_io) libraries.

Gotcha! I'll get that change implemented!

pinkenburg:
This should go through jenkins now. One remark - please take our rule about include file ordering in our coding conventions https://wiki.sphenix.bnl.gov/index.php/Codingconventions very seriously. They were the other way round which will bite you.
Also the include files have a lot of unnecessary includes which can be replaced by forward declarations (or just removed). Doing this will speed up compilation and reduce the number of sources which are affected by changes in include files

ruse-traveler:
Apologies! That's a bad habit I haven't broken yet 😓 
I'll be more conscious of include's for the next PR (and fix these when I get a chance).

PR #2751, opened 2024-05-15
Update PHCASeeding with more efficient code to execute same algorithm.

david-stewart:
The code revises PHCASeeding for efficiency of execution. The result is that PHCASeeding now runs ~12-30% faster for generating the bilinks (the most time consuming part of the code). Some changes are:
* The boost trees are now 2D instead of 3D
* The cached map is now a std::unordered_map instead of a std::map
* `using` statements are now scoped internal to the namespace of the PHCASeeding class (before they were set in the global namespace)
* The keyLinks now are just std::pair<TrkrDefs::cluskey,TrkrDefs::cluskey> instead of the std::array<coordKey,2> with coordkey=std::pair<std::array<float,3>,TrkrDefs::cluskey. This is because in the code, as implemented, the global position of the links was always taken from the cached cluster positions, instead of the internal std::arrayu<float,3> anyway
etc...

In better news, the code that follows the bilinks is actually on the order of ~10% or less of the time it takes to generate the bilinks. This means that if we want, we *could* make that part of the code more sophisticated and still only consume part of the time we have already saved with this PR by making the bilink generation more efficient.

While this all works, it seems likely that for our application, we could actually forgo some of the actual algorithm, and just pass key-chains directly to the `PHSimpleKFPProp` without so much logic inside of PHCASeeding itself. If we really want to decrease time, we could also probably not even measure the bilinks in the inner ~15 rows of the TPC, as all the tracks should be seeded from the external rows anyway. The change to the code would be negligible (just adjusting the loop).

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackreco/PHCASeeding.cc
+++ b/offline/packages/trackreco/PHCASeeding.cc
@@ -74,6 +74,13 @@
 #define LogWarning(exp) \
   if (Verbosity() > 0) std::cout << "WARNING: " << __FILE__ << ": " << __LINE__ << ": " << exp
 
+#if defined(_PHCASEEDING_TIMER_OUT_)
+#define _PHCASEEDING_PRINT_TIME(timer,statement) timer->stop(); \
+  std::cout << " _PHCASEEDING_PRINT_TIME: Time to " << statement << ": " << timer->elapsed()/1000 << " s" << std::endl;
+#else
+#define _PHCASEEDING_PRINT_TIME(timer, statement) (void) 0
+#endif
+
 /* #define FIXME(str) std::cout <<" FIXME: " << str << std::endl; */
 
 //#define _DEBUG_
@@ -401,21 +408,15 @@ int PHCASeeding::FindSeedsWithMerger(const PHCASeeding::PositionMap& globalPosit
   t_seed->restart();
 
   keyLinksPerLayer bilinks = CreateBiLinks(globalPositions, ckeys);
-  if (Verbosity() > 0) {
-    t_makebilinks->stop();
-    std::cout << "Time to make bilinks: " << t_makebilinks->elapsed() / 1000 << " s" << std::endl;
-  }
+  _PHCASEEDING_PRINT_TIME(t_makebilinks, "init and make bilinks");
   // std::pair<ckeylinks_per_layer, ckeylinks_per_layer> = CreateLinks(fromPointKey(allClusters), globalPositions); // put the bilinks finder into the end of create links --> don't have to save all the rows
   // std::vector<std::vector<keylink>> biLinks = FindBiLinks(links.first, links.second);
   
 
-  if (Verbosity() > 0) { t_makeseeds->restart(); }
+  t_makeseeds->restart();
   std::vector<keyList> trackSeedKeyLists = FollowBiLinks(bilinks, globalPositions);
-  if (Verbosity() > 0 )
-  {
-    t_makeseeds->stop();
-    std::cout << "Time to make seeds: " << t_makeseeds->elapsed() / 1000 << " s" << std::endl;
-  }
+  _PHCASEEDING_PRINT_TIME(t_makeseeds, "make seeds");
+
   std::vector<TrackSeed_v2> seeds = RemoveBadClusters(trackSeedKeyLists, globalPositions);
 
   publishSeeds(seeds);
--- a/offline/packages/trackreco/PHCASeeding.cc
+++ b/offline/packages/trackreco/PHCASeeding.cc
@@ -74,6 +74,8 @@
 #define LogWarning(exp) \
   if (Verbosity() > 0) std::cout << "WARNING: " << __FILE__ << ": " << __LINE__ << ": " << exp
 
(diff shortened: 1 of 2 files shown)


osbornjd:
This dramatically changes the QA, in some positive and some negative ways. Can we do some more tests before merging this? 

1. Can you run an actual timing test to see what the change is? Or maybe you already did this?
2. Do we understand why the purity of the tracks goes down and the efficiency overall increases? This is probably correlated but I meant do we understand it from an algorithmic standpoint - i.e. what changes in the logic of the seeder lead to this behavior.

david-stewart:
> This dramatically changes the QA, in some positive and some negative ways. Can we do some more tests before merging this?
> 
>     1. Can you run an actual timing test to see what the change is? Or maybe you already did this?
> 
>     2. Do we understand why the purity of the tracks goes down and the efficiency overall increases? This is probably correlated but I meant do we understand it from an algorithmic standpoint - i.e. what changes in the logic of the seeder lead to this behavior.

I did run some tests. I turns out that I am seeing some (to me) anomalous behavior from the `std::binary_search` function which is finding the heads of the linked chains. When run with `std::any_of` I believe that the results are identical. I'll swap that bit back, and it should give totally identical results. Let's just wait and see what the new push does.

osbornjd:
It's not necessarily bad, the efficiency is improved nontrivially. I just want to make sure we understand why

david-stewart:
I did run a local test

> It's not necessarily bad, the efficiency is improved nontrivially. I just want to make sure we understand why

I did run local tests. The results are identical up to the generation of the bilinks (which is the time intensive part of the code). I also made a small update to making the seeds from the bilinks, which resulted in far too many seeds (it basically considered every 6 link chain as a seed, and passed too many seeds to the KFilter).  I'll undo that last small change, and we should see the results go back to identical in seeds generated as well.

image

david-stewart:
The new code results in identical code as the old. I found the error in my `std::binary_search` implementation. There is a modest additional change which I wish to add before merging. I'll run times again, but a test of ten events had the following statistics using essentially this version. The first set is the times for the bilink generation from the master branch to this branch (in seconds), and their ratio changes. The second set the seed track generation. This is only 10 events, and may have been influenced by network speed, etc... However, the fact that the seeding time are increasing and decreasing (essentially randomly) gives support to the fact that the more expensive rTree to bilink generation uniform time drop (which are the bulk of the time) are not random. A better metric would be the ratio of the PHSimpleKFProp time per event.

```
make bilinks times
 0.057 -> 0.047 : delta -0.010 ratio -0.179
 0.044 -> 0.037 : delta -0.007 ratio -0.153
 0.015 -> 0.010 : delta -0.005 ratio -0.326
 0.043 -> 0.037 : delta -0.007 ratio -0.154
 0.045 -> 0.035 : delta -0.010 ratio -0.222
 0.070 -> 0.057 : delta -0.013 ratio -0.189
 0.041 -> 0.035 : delta -0.006 ratio -0.155
 0.017 -> 0.015 : delta -0.002 ratio -0.123
 0.047 -> 0.039 : delta -0.008 ratio -0.168
 0.038 -> 0.030 : delta -0.008 ratio -0.223
make seeds times
 0.001 -> 0.001 : delta 0.000 ratio 0.057
 0.001 -> 0.002 : delta 0.001 ratio 1.598
 0.000 -> 0.000 : delta -0.000 ratio -0.321
 0.001 -> 0.001 : delta 0.001 ratio 0.666
 0.001 -> 0.001 : delta 0.000 ratio 0.121
 0.001 -> 0.003 : delta 0.001 ratio 1.372
 0.001 -> 0.002 : delta 0.001 ratio 1.742
 0.000 -> 0.000 : delta -0.000 ratio -0.346
 0.001 -> 0.001 : delta 0.000 ratio 0.008
 0.001 -> 0.001 : delta -0.000 ratio -0.321
 ```

osbornjd:
Okay nice, thanks for the numbers. If you are satisfied we can merge this, I'll wait for your approval

david-stewart:
ok -- I don't want to make a project out of it (there are probably more important things to do) but there is one last set of things that I want to try today which will simplify the second half of the code, and may make the logic around chaining the bilinks and selecting which chains to make into seeds easier. (It's a toss up if it is will actually make it shorter, but it is around 2-3% of the PHCASeeding run time (excluding the check on bad chains which actually isn't doing anything in the current implementation) so it probably doesn't matter speed-wise either way. I'll try this morning and test it. If it works, I'll push it, and we can make a single merge. Otherwise, we can merge this one.

osbornjd:
@david-stewart are we good to merge this now?

david-stewart:
> @david-stewart are we good to merge this now?

Yes. Clang-tide is complaining about the pre-processor macro programing. If I have to it later, that's fine and I will. It just seemed like a zero-cost way to identify and keep the QA in-side of it.
I also have a lot more understanding of the tradeoffs for the original choices in the programming. I thought I could dramamtically improve them, but all the edge cases kept pushing me back closer to what was already there.

PR #2770, opened 2024-05-23
Micromegas full event assembly

hupereir:
This is a new attempt at full event assembly for TPOT, taking advantage of the new FEE firmware. 
It uses the gtm bco information, projected back to the 20 bits FEE BCO to decide which waveform belongs to which trigger. 
Works much better than what we had before. 

I am aware that Jin is working on a similar code for the TPC, but having two independantly written codes to do the same here: it will help debug the many foreseen error conditions from the high trigger rate, zero suppressed data we will soon have to assemble. 

Once everything is ironed out, we will converge on a unique version applicable to both the TPC and TPOT. 

In the meanwhile this PR will allow to properly assemble the recent high trigger rate, non-zero suppressed data taken with TPOT

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/framework/fun4allraw/SingleMicromegasPoolInput.cc
+++ b/offline/framework/fun4allraw/SingleMicromegasPoolInput.cc
@@ -87,7 +87,7 @@ unsigned int SingleMicromegasPoolInput::bco_matching_information_t::get_predicte
   static constexpr double multiplier = 4.2629164;
 
   // get lvl1 bco difference with proper rollover accounting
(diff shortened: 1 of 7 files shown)


hupereir:
Side note: I'm still evaluating, debugging this. But this is already much better than what we had before. 
Side note two: for this to work stabiliy we need the right multipliers between the fee clock frequency and the gtm clock frequency. Current value has been estimated experimentally :)
Will update when we get the infor from John K.

PR #2819, opened 2024-06-05
Added pedestal RMS and ZS RMS histograms

kr2580:

## Types of changes

Added pedestal RMS and zero-suppressed RMS histograms in iphi and ieta.

### DIFF
--- a/offline/QA/Calorimeters/CaloValid.cc
+++ b/offline/QA/Calorimeters/CaloValid.cc
@@ -5,7 +5,7 @@
 #include <calobase/RawClusterContainer.h>
 #include <calobase/RawClusterUtility.h>
 #include <calobase/RawTowerGeomContainer.h>
-#include <calobase/TowerInfov2.h>
+#include <calobase/TowerInfo.h>
 #include <calobase/TowerInfoContainer.h>
 
 #include <fun4all/Fun4AllHistoManager.h>
@@ -40,14 +40,10 @@
 #include <string>
 #include <utility>  // for pair
 
-CaloValid::CaloValid(const std::string& name, const std::string &filename) // int nevents)
+CaloValid::CaloValid(const std::string& name)
   : SubsysReco(name)
   , detector("HCALIN")
-  , m_outputFileName(filename)
 {
-  OutputNtupleFile = nullptr;
-  OutputFileName = filename;
-  //_nevents = nevents;
 }
 
 CaloValid::~CaloValid() = default;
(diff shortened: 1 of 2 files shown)


offline/QA/Calorimeters/CaloValid.cc
```diff
@@ -5,7 +5,7 @@
 #include <calobase/RawClusterContainer.h>
 #include <calobase/RawClusterUtility.h>
 #include <calobase/RawTowerGeomContainer.h>
-#include <calobase/TowerInfo.h>
+#include <calobase/TowerInfov2.h>
```
osbornjd, line 8: We should only include the base class, the versioned classes should never need to be included in analysis modules. If they are needed then something has gone wrong

offline/QA/Calorimeters/CaloValid.cc
```diff
@@ -54,9 +58,45 @@ int CaloValid::Init(PHCompositeNode* /*unused*/)
   {
     std::cout << "In CaloValid::Init" << std::endl;
   }
+  std::cout << "CaloValid::Init started..." << std::endl;
+  OutputNtupleFile = new TFile(OutputFileName.c_str(),"RECREATE");
```
osbornjd, line 62: You should not use a new file, you should use the histomanager will deal with all the histogram memory management for you like is done in `createHistos`

offline/QA/Calorimeters/CaloValid.cc
```diff
@@ -54,9 +58,45 @@ int CaloValid::Init(PHCompositeNode* /*unused*/)
...
 
   createHistos();
+  char hname[99];
+  char hnameE[99];
 
+//---------EMCal--------//
+{
+  int size = 128*192;
+  for(int channel = 0; channel < size; channel++){
+    sprintf(hname,"h_cemc_channel_pedestal_%d",channel);
```
osbornjd, line 73: see `createHistos()` to use boost instead of sprintf

offline/QA/Calorimeters/CaloValid.cc
```diff
@@ -671,11 +678,96 @@ int CaloValid::process_towers(PHCompositeNode* topNode)
...
+        float ped_rms = h_ohcal_channel_pedestal[channel]->GetRMS();
+        h_CaloValid_ohcal_etaphi_pedRMS->Fill(ieta, iphi, ped_rms);
+        MirrorHistogram(h_ohcal_channel_energy[channel]);
+	double rmsZS = h_ohcal_channel_energy[channel]->GetRMS();
+	h_CaloValid_ohcal_etaphi_ZSpedRMS->Fill(ieta, iphi, rmsZS);
+      }
+
+   }
+ }
+  QAHistManagerDef::saveQARootFile(OutputFileName);
```
osbornjd, line 755: This is not needed, this should be called in your macro. The production needs to work this way so that file names can be specified as they already are in the production scripts

offline/QA/Calorimeters/CaloValid.h
```diff
@@ -48,10 +48,25 @@ class CaloValid : public SubsysReco
...
+  void MirrorHistogram(TH1* histogram);
   std::string getHistoPrefix() const;
   bool m_debug{0};
   std::string detector;
+  TFile* OutputNtupleFile;
+  std::string m_outputFileName;
+  std::string OutputFileName;
 
+  TH1F* h_cemc_channel_pedestal[128*192];
+  TH1F* h_ihcal_channel_pedestal[32*48];
```
osbornjd, line 60: These should not be needed to be owned by the CaloValid module if they are properly registered with the QAUtil HistoManager, which will do the memory management for you

osbornjd (changes requested):

offline/QA/Calorimeters/CaloValid.h
kr2580: These are temporary histograms, should I still use the histogram manager? I do not want them saved to the output file

pinkenburg: If you don't register the histograms with the qa, they won't be saved. Means don't use the histo manager for those

kr2580: Just to make sure, I can leave these histograms like this?

pinkenburg:
sorry - pushed the wrong button. Yes the histos are fine. I just modified the include file - we try to use base class pointers (TH1*, TH2*), the type is set when the histogram is created with TH2 *h2 = new TH2F.. This allows us to change easily the type if needed (e.g. going from floats to doubles TH2F to TH2D) in a single place - just like out versioned classes

PR #2847, opened 2024-06-13
Towerinfov4 slimmed down tower objects for high lumi running 

bseidlit:

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/packages/CaloBase/TowerInfov4.h
+++ b/offline/packages/CaloBase/TowerInfov4.h
@@ -4,7 +4,7 @@
 #include "TowerInfo.h"
 #include "TowerInfov1.h"
 
-class TowerInfov4 : public TowerInfov1
+class TowerInfov4 : public TowerInfo
 {
  public:
   TowerInfov4() {}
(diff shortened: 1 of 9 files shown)


offline/packages/CaloBase/TowerInfov4.h
pinkenburg, line 1: Please remove TMath, no need to use it here (log is part of c++ cmath) and follow our include file conventions:
https://wiki.sphenix.bnl.gov/index.php?title=Codingconventions#Include_Files

pinkenburg:
The clang error is from another PR, if you merge this with the current master branch it'll be fixed

bseidlit:
@pinkenburg  I switched to std log, it compiled and it ran a test.  Do you think we can just merge this so others can test the new objects?

pinkenburg:
We probably need one more change - the memory layout of the CaloTowerv4 is inefficient. It inherits from CaloTowerv1 which uses
 private:
  short _time = 0;
  float _energy = 0;
which should have never been committed this way (it leaves a 16 bit hole in the middle of the data). Now CaloTowerv4 adds to this and it's layout is
 private:
  short _time = 0;
  float _energy = 0;
 uint8_t _chi2 = 0;
  uint8_t _status = 0;
adding another 16 bit hole at the end. We should drop the inheritance from v1 and use

private:
 float _energy = 0;
short _time = 0;
uint8_t _chi2 = 0;
 uint8_t _status = 0;

which now lines this nicely up along 32 bit boundaries (or event 64 bit)
and maybe drop the _'s from the variable names which are just painful when T->Drawing() interactively

bseidlit:
@pinkenburg  I made the switches.  The resulting DST was the exact same size, not sure if that was expected but OK.

pinkenburg:
I didn't expect the size to change but for the cpu I was thinking that it's easier for memory access by the cpu if data is aligned (they typically fetch 32 or 64 bit words in one shot independent of the "real" size). I don't know if this has a real effect but it's free - so why not try? In terms of storage - root stores every variable separately in a "basket" which is then compressed, for that the memory layout doesn't matter.

bseidlit:
OK, yeah Ii was wondering if it was something like that.  Yeah no sweat, certainly free and an easy change. I also never understood the leading underscore (no ones got time for that!!)

PR #2883, opened 2024-06-20
ZDC north/south energy

eumaka:
create module to extract zdc calibrated energy sums in north and south


eumaka:
Need to add forgotten sqrt for r calculation, but will wait for this to go through first so as not to re-trigger Jenkins

eumaka:
ah nvm, it turns out I already did :)

pinkenburg:
You committed a lot of binary files which don't belong here (I assume your editor generated them?). Please remove them from this PR

eumaka:
yep, thanks to vscode

eumaka:
thanks Jin for killing the extra checks and sorry Chris for the expensive checks :)

offline/packages/zdcinfo/Zdcinfov1.h
pinkenburg, line 1: We should initialize the energy to NaN. {} initializes to zero which is a possible value

PR #2884, opened 2024-06-20
fix: suppress verbosity

osbornjd:
@hupereir any objection to suppressing this with a verbosity call? it can fill up logfiles

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/framework/fun4allraw/SingleMicromegasPoolInput.cc
+++ b/offline/framework/fun4allraw/SingleMicromegasPoolInput.cc
@@ -186,7 +186,10 @@ void SingleMicromegasPoolInput::FillPool(const unsigned int /*nbclks*/)
       // drop packet if not found
       if (!bco_matching_information.is_verified())
       {
-        std::cout << "SingleMicromegasPoolInput::FillPool - bco_matching not verified, dropping packet" << std::endl;
+        if(Verbosity())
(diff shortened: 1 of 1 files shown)


hupereir:
@osbornjd This message tells you that the TPOT data is unuseable essentially. What can happen is that you get it a few times before finally a proper reference is found and the run can be reconstructed.
If it fills the log, it means that the entire run is unuseable for TPOT and TPOT should then not be included in the reconstruction. (essentially all runs before 45288 ) 

So all in all, I would rather not hide this message behind verbosity because  this is really a problem. (and if the data is reconstructible, the message should not appear, or only a few times).

What I can do is set a limit to the number of times it happens and then abort, if you prefer.

osbornjd:
> What I can do is set a limit to the number of times it happens and then abort, if you prefer.

Sure, that would be nice as if the run is not reconstructible for TPOT then we should just not continue on. This would also suggest we should separate the event combining production into separate jobs though, as a run that is bad for TPOT may not be for some other subsystem. Perhaps we should punt this to when some further development on that front happens.

hupereir:
could you point me to one or several run numbers for which this message has been flooding the logs ?

osbornjd:
Takao reported this, let me discuss with him

PR #2889, opened 2024-06-21
New feature to identify and recover bit flip waveforms from within template fit method of CaloWaveformFitting 

emclaughlin2:
New option to identify and recover bit flips of 2^11, 2^12 and 2^13 from calorimeter waveforms allowing for towers with known intermittent hot tower behavior to be recovered. 
Option can be accessed by turning on bit flip recovery option in CaloTowerBuilder class which propagates to the fitter.
New member variables of TowerInfov2 and TowerInfov4 are included as well to give information as to whether a tower has been recovered using this recovery method (variable type float in v2 and bool in v4). 

Additional older changes to interface between run database and fun4all.

## Types of changes

New feature to allow for waveforms of reoccurring bit flip towers to be recovered.

## TODOs (if applicable)

### DIFF
--- a/offline/database/rundb/CaloTemp.cc
+++ b/offline/database/rundb/CaloTemp.cc
@@ -48,6 +48,13 @@ CaloTemp::CaloTemp(const std::string& name, const std::string& filename)
   , outfilename(filename)
   , temphist(true)
 {
+  runnumber = -9999;
+  second = -9999;
+  minute = -9999;
+  hour = -9999;
+  day = -9999;
+  month = -9999;
+  year = -9999;
 }
 
 CaloTemp::~CaloTemp()
--- a/offline/database/rundb/CaloTemp.cc
+++ b/offline/database/rundb/CaloTemp.cc
@@ -134,7 +134,19 @@ int CaloTemp::InitRun(PHCompositeNode* topNode)
     std::cout << "runtime " << runtime << std::endl;
   }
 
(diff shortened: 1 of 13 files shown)


sPHENIX-GitHub-l:
Sorry, what is it that has been saved? This code generates the local to global transformations used to project readout cluster positions to the global frame. It is completely independent of anything written to the data files. That was the point. I am testing it on existing data files without any problems.

The only thing that is saved anywhere is the alignment constants file that is read in by the AlignmentTransformation module. I could, if we are worried about it, modify the code to detect if it sees 6 or 9 parameters and set the rotation parameters to zero so the code does not crash. But then it would get wrong answers, so I would have it abort with an error message.  Actually, that is what I should have done. I will change it.

Tony

pinkenburg:
Hi Tony, 
this is a PR for calorimeters
Chris

adfrawley:
BTW, I should have realized that Jenkins would fail because the default alignment parameters file has only 6 parameters.

I will change the code so that it detects the wrong number of parameters and sets the additional three to zero for now so it can be tested. It can be changed later to abort if a 6-parameter file is read in.

Tony

bseidlit:
Opps, yeah this is wrong.  There is an 8 bit status which not all the bits are used.  When I say "add a status bit" I unfortunately mean "start using an unused status bit".  Currently the 6th bit is not being used.  So follow the example of like 
  void set_isZS(bool isZS) override { set_status_bit(5, isZS); }
  bool get_isZS() const override { return get_status_bit(5); }

PR #2893, opened 2024-06-22
Prelim alignment

adfrawley:

## Types of changes
- [ x] Breaking change (fix or feature that would cause existing functionality to not work for users)

This PR adds an alignment rotation in the global frame into the Acts local to global transformation. The alignment rotation occurs after the Acts transform and before the alignment translation. The purpose is to allow for rotations of layers around the detector axis. It was specifically added to correct the relative rotation of the MVTX layers with respect to each other. 

This is a breaking change because it increases the number of alignment parameters from 6 to 9. The alignment parameters file in the database will need to be changed at the same time that this is merged. 

There will be a later PR for calibrations/TrackerMillepedeAlignment/makeAlignmentParams.C so that it provides nine parameter alignment files.

## TODOs (if applicable)


osbornjd:
Do we really want to put this in to the repo? I worry that this is going to complicate things long term by having more parameters needed for all layers when this is really only relevant for the mvtx. If we are confident that the layer rotations are needed, which the plots show they are, then why don't we just implement this info the geometry? Are we really going to have to make this change more than once?

osbornjd:
I guess the idea is that we would do that when we also understand the 5 mm shift WRT the beam pipe?

sPHENIX-GitHub-l:
Hi Joe,

I don't think it complicates things especially. We can fix any parameters that we don't want to vary when fine tuning the alignment. And I seriously doubt that the MVTX is the only place where we will need the additional rotation. Also, the parameters I used were eye-balled from the uncorrected residuals. They will get better as we fix other alignment issues.

For analyzing data, there is no difference at all between changing the reference geometry and changing the alignment parameters. Either way, we are just changing a local-global transform of the readout parameters. Until we are ready to do simulations with more realistic detector mass distributions, putting what we learn into the alignment constants as we go seems to me to be best.

We do have to be careful when merging this. The database default alignment file needs to be updated to 9 parameters or the reconstruction will break.

As we learn about things we need to update the file in the database. But we should discuss and agree on what we put into the default alignment file every time we update it.

Tony

PR #2912, opened 2024-06-26
Avoid Name Clashes and Add Functionality to Select Triggers in Jet QA

ruse-traveler:
This PR makes two major updates to the Jet QA:

1. Certain inputs and parameters have been made user-configurable in order to avoid name clashes across modules and histograms, and to allow for the same module to be run on different jet populations in parallel;
2. And functionality has been added to all modules to allow the user to select on a specific trigger by specifying the desired trigger's index in the GL1 trigger vector.

## Types of changes

This PR introduces the new features described above. There have been modest changes to the API for certain modules (e.g. adding additional arguments to ctor's), and so users should update their downstream macros accordingly.

### DIFF
--- a/offline/QA/Jet/ConstituentsinJets.cc
+++ b/offline/QA/Jet/ConstituentsinJets.cc
@@ -32,6 +32,7 @@
 #include <TH2.h>
 #include <TH1.h>
 
+#include <algorithm>
 #include <cmath>
 #include <iostream>
 #include <string>
@@ -44,6 +45,7 @@
 
 ConstituentsinJets::ConstituentsinJets(const std::string &moduleName, const std::string &recojetname, const std::string &towBkgdName, const std::string &histTag)
   : SubsysReco(moduleName)
+  , m_moduleName(moduleName)
   , m_recoJetName(recojetname)
   , m_towBkgdName(towBkgdName)
   , m_histTag(histTag)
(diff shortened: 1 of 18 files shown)


osbornjd (changes requested):
This is somewhat nitpicky, but we have a ton of QA histograms being produced now (good). Many of them go into the same file based on whatever the production job is, so if there is not a clear naming scheme to the histogram names it might be hard to track down where they came from for someone who is looking to understand what actually went into filling the histogram. Most (probably not all, which would be nice to unify) of the QA is using a unified scheme that looks something like `h_ModuleName_HistogramName` - can we implement that here?

osbornjd (commented):
One other comment - `JetQADefs` looks more like a super useful `TriggerDefs` object that would be used in many different analyses. @danjlis or @pinkenburg is there any plan for something like this, and if not lets generalize this so that we can have one centralized place for all things offline trigger decision

ruse-traveler:
> This is somewhat nitpicky, but we have a ton of QA histograms being produced now (good). Many of them go into the same file based on whatever the production job is, so if there is not a clear naming scheme to the histogram names it might be hard to track down where they came from for someone who is looking to understand what actually went into filling the histogram. Most (probably not all, which would be nice to unify) of the QA is using a unified scheme that looks something like `h_ModuleName_HistogramName` - can we implement that here?

Sure! That will be straightforward to implement!

osbornjd:
I guess we can leave the JetQADefs module to a future PR - it would be good to generalize this though once we get a more standardized offline trigger object

PR #2944, opened 2024-07-03
update sEPD phibin for north arm

eumaka:
Mirror phi bin in north


eumaka:
appears to fail for reasons other than the commit

PR #2954, opened 2024-07-05
fix: add flag end run in INTT pooling

osbornjd:
In discussion with @mpurschke and @hahahachiya , there are INTT pooling job where the job abruptly ends due to a mismatched run number. Martin traced this down and it appears that the end run event is filled and data after this is corrupted, due to some interplay between the INTT RCDAQ plugin and RCDAQ itself. Until further upstream changes in the firmware can be made to handle this, this PR puts a flag in that if the `ENDRUNEVENT` is found the INTT pooler exits gracefully.

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/framework/fun4allraw/SingleInttPoolInput.cc
+++ b/offline/framework/fun4allraw/SingleInttPoolInput.cc
@@ -90,6 +90,7 @@ void SingleInttPoolInput::FillPool(const unsigned int /*unused*/)
       {
         std::cout << "End run flag for INTT found, remaining INTT data is corrupted" << std::endl;
         delete evt;
+        AllDone(1);
         return;
       }
       delete evt;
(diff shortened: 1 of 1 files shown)


offline/framework/fun4allraw/SingleInttPoolInput.cc
pinkenburg, line 1: can you add the setting of the all done flag by this:
AllDone(1);
This signals upstream that there is nothing more to be expected from this input manager and the combiner will not ask again for more events. Otherwise it might request more events in the next iteration which will then likely come back with the invalid run number

osbornjd: Happy to do this, I tested on the cosmic run you provided and what happened was that the input manager received the signal of "no more events to process because the hit map is empty" so it finishes processing gracefully. I can add this as it is probably clearer in the code

osbornjd:
Merging this to start a new build

PR #2979, opened 2024-07-10
adding event number for gl1-emcal alignment tracking

bseidlit:

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/QA/Global/GlobalQA.cc
+++ b/offline/QA/Global/GlobalQA.cc
@@ -28,6 +28,7 @@
 #include <phool/phool.h>  // for PHWHERE
 
 #include <ffarawobjects/Gl1Packet.h>
+#include <ffaobjects/EventHeader.h>
 
 #include <TH1.h>
 #include <TH2.h>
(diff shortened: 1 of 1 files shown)


offline/QA/Global/GlobalQA.cc
```diff
@@ -85,7 +86,22 @@ int GlobalQA::process_towers(PHCompositeNode *topNode)
...
+    }
+  }
+  else{
+    std::cout << "GlobalQA::process_event()  No event header" << std::endl;
+  }
+
+  //--------------------------- MBD vertex------------------------------//
   MbdVertexMap *mbdmap = findNode::getClass<MbdVertexMap>(topNode, "MbdVertexMap");
   MbdVertex *bvertex = nullptr;
   float mbd_zvtx = -999;
```
pinkenburg, line 107: We should use nan as invalid initialisation. Comparing floats with == is just bad practice and should be avoided. std::isfinite() works. There is not much of a point filling a histogram with goes to -300 with -999, if the idea is that these counts end up in the overflow bin - this can be set directly with SetBinContent()

offline/QA/Global/GlobalQA.cc
```diff
@@ -102,11 +118,11 @@ int GlobalQA::process_towers(PHCompositeNode *topNode)
   }
   h_GlobalQA_mbd_zvtx->Fill(mbd_zvtx);
   h_GlobalQA_mbd_zvtx_wide->Fill(mbd_zvtx);
-  if (mbd_zvtx == -999) 
+  if (mbd_zvtx == -999)
   {
     h_GlobalQA_mbd_zvtxq->Fill(0);
```
pinkenburg, line 123: probably better to use SetBinContent() instead of relying on histogram binning to get this right

PR #3004, opened 2024-07-22
add setter for TPC readout range

bogui56:

## Types of changes

## TODOs (if applicable)


osbornjd (commented):
Why do we need a setter for this when the number of z bins is set in the geometry? This is just adding another variable for us that we have to keep track of, no?

sPHENIX-GitHub-l:
Outside the simulation world this range isn't set anywhere. 
Also we want to be able to custom set this in the clusterizer so we can set it to a different value to cut out the laser flashes. 

We should probably put an explicitt setter in the geometry as well for the reco world.

osbornjd:
>We should probably put an explicitt setter in the geometry as well for the reco world.

That is my confusion:
```
NTBins = (unsigned short) layergeom->get_zbins();
```
this is obtaining the number of time bins from the geometry, is it not? The layergeometry has nothing to do with simulation, it is the geometry object

PR #3085, opened 2024-08-21
add prelim ZDC z-vertex determination 

eumaka:
This PR adds a zdc z-vertex determination using a  energy weighted average of the zdc times. Raw signal extraction is still the "FAST" option. After a discussion with the cold QCD team (Itaru et al.), this initial determination will do for now for the cold QCD analysis needs.


pinkenburg:
we cannot add data members to versioned classes once they were saved. Please bump up the version

offline/packages/zdcinfo/ZdcReco.h
```diff
@@ -44,13 +44,18 @@ class ZdcReco : public SubsysReco {
   std::vector<float> vzdctime;
   float smd_adc[32] = {0.0f};
   float smd_pos[4] = {0.0f};
-  float radius_south{0.};
-  float radius_north{0.};
+  float _radius_south{0.};
+  float _radius_north{0.};
+  const double _t{17.623}; //convert to ns
+  const double _c{29.9792}; //speed of light in cm/ns
```
pinkenburg, line 50: Please use c from gsl include (depending on units, e.g. gsl/gsl_const_cgs.h)

PR #3094, opened 2024-08-26
change trackmap name & lepton PDGID

yuxdPKU:

## Types of changes

Two changes in this PR
1. In KFParticle_DST, we search for SvtxTrack by using track id which is the index in that track map. We should search in the input track map, rather than "SvtxTrackMap", because we can use track map different from SvtxTrackMap from default tracking software.
2. lepton PDG ID convention is different from charged hadrons. For example, pi+ --> 211, pi- --> -211, e+ --> -11, e- --> 11. I add a if statement to time an extra minus sign for lepton including e+/-, mu+/-, tau+/-.

## TODOs (if applicable)

### DIFF
--- a/offline/packages/KFParticle_sPHENIX/KFParticle_DST.cc
+++ b/offline/packages/KFParticle_sPHENIX/KFParticle_DST.cc
@@ -183,7 +183,7 @@ void KFParticle_DST::fillParticleNode_Track(PHCompositeNode* topNode, const KFPa
     }
   }
 
-  SvtxTrackMap* originalTrackMap = findNode::getClass<SvtxTrackMap>(topNode, "SvtxTrackMap");
+  SvtxTrackMap* originalTrackMap = findNode::getClass<SvtxTrackMap>(topNode, m_origin_track_map_node_name.c_str());
   KFParticle* daughterArray = &daughters[0];
 
   for (unsigned int k = 0; k < daughters.size(); ++k)
--- a/offline/packages/KFParticle_sPHENIX/KFParticle_DST.h
+++ b/offline/packages/KFParticle_sPHENIX/KFParticle_DST.h
@@ -46,6 +46,7 @@ class KFParticle_DST
   bool m_write_track_container = true;
   bool m_write_particle_container = true;
   std::string m_container_name;
+  std::string m_origin_track_map_node_name = "SvtxTrackMap";
 
  private:
   SvtxTrackMap* m_recoTrackMap = nullptr;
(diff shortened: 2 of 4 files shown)


osbornjd:
@cdean-github any objection to this?

cdean-github:
> @cdean-github any objection to this?

I don't have a problem with the code. However, [there is missing instance of setting the lepton ID here](https://github.com/sPHENIX-Collaboration/coresoftware/blob/master/offline/packages/KFParticle_sPHENIX/KFParticle_eventReconstruction.cc#L428) which should be changed

My objection before to setting the SvtxTrackMap name to something other than the default in the DST was it could hide bugs in other peoples code. I thought we should always maintain the track map indices for appropriate mappings and if someone doesn't do this, then my code will return a different track. By changing this, we could mask problems like that which would cause confusion downstream and may involve lost time trawling through modules.

I guess my summary is, Xudong's code is correct. Im happy for it to be merged after the last PDG ID change is made, as long as we're OK with the potential issue with the track map association between modules

yuxdPKU:
Thanks Cameron. Let me also explain it again. My starting point is from photon conversion analysis. I want to apply matching cut on tracks which can not be applied by using KFParticle interface. Therefore, i need to use a analysis module to select tracks and then insert into a new track map. Let me call it **SvtxTrackMap_new**. The same track in **SvtxTrackMap** and **SvtxTrackMap_new** has different id, because get_id() function just return the order in trackmap. So it has problem that i use id of track in **SvtxTrackMap_new** and search in **SvtxTrackMap**. It probably return a wrong track. If we do not use setTrackMapNodeName, the default trackmap is still **SvtxTrackMap**.

yuxdPKU:
I mean Track-Calo matching cut, like |dphi|, |dz| cut

yuxdPKU:
For my understanding, it is non-trivial to do that. Maybe @cdean-github can comment more about that. Apply kinematic cut on track, like pt, p, quality cut is feasible, because it can be accessed from SvtxTrack object. However, if we apply some more complicated cuts, like track matching with other subsystem, it is non-trivial.
As far as i know, before this PR, KFParticle_sPHENIX has already allowed to use customized (SvtxTrackMap*) that is produced by HFTrackEfficiency to select tracks matched by truth information.
This PR just want to make input trackmap, and the trackmap we used to search for tracks consistent. So for my understanding, it is fine.

osbornjd:
Okay, I misunderstood. There are some basic track quality cuts in KFP but this requires something more complex since we are matching the tracks to the calorimeters first. Let's merge it

PR #3097, opened 2024-08-26
Add a small module to filter TPC seeds for:

david-stewart:
1. minimum number of clusters
2. require clusters in multiple radial segements
3. min pt
4. max pt
6. min eta 7 max eta

The default behavior is to make no cuts at all, and the user can use either the arguments in the constructor or separate setters. All it does is delete tracks don't pass cuts.
Currently acts only on TPC segements (not TPC + Silicon).

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackreco/TpcSeedFilter.cc
+++ b/offline/packages/trackreco/TpcSeedFilter.cc
@@ -1,0 +1,110 @@
+#include "TpcSeedFilter.h"
+
+#include <phool/PHCompositeNode.h>
+#include <phool/PHDataNode.h>
+#include <phool/PHNode.h>
+#include <phool/getClass.h>
+
+#include <trackbase_historic/SvtxTrack.h>
+#include <trackbase_historic/SvtxTrackMap.h>
+
+#include <fun4all/Fun4AllReturnCodes.h>
+
+int TpcSeedFilter::process_event(PHCompositeNode* topNode)
+{
+  if (topNode == nullptr)
+  {
+    return Fun4AllReturnCodes::ABORTRUN;
+  }
+  if (Verbosity() > 1000)
+  {
+    topNode->print(); 
+  }
+
+  // Get the map of all the input seeds
+  SvtxTrackMap* seeds = findNode::getClass<SvtxTrackMap>(topNode, "SvtxTrackMap");
+  if (!seeds)
+  {
+    std::cout << "Could not locate SvtxTrackMap node when running "
+              << "\"TpcSeedFilter\" module." << std::endl;
+    return Fun4AllReturnCodes::ABORTEVENT;
+  }
+
+  // iterate through the seeds and remove the ones that fail the cuts
+  for (SvtxTrackMap::Iter track_pair = seeds->begin();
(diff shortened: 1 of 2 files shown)


david-stewart:
Hey @jdosbo,
Beam is down, so just wrapped up this very small module. See the notes on the application. Note sure if we prefer to put it in a different modules. It is currently in draft form, but should work to plug into a script.

Some immediate questions regarding interface and use:
1. It is hardcoded to pick up the node names "SvtxTrackMap"
2. It works over all TPC tracks (independent of status to silicon).

osbornjd (changes requested):
Right now this operates on tracks, and the intent was for it to operate on seeds (e.g. right after the PHSimpleKFProp module is run, for example). Can you change this to operate on the TpcTrackSeedContainer instead of the SvtxTrackMap?

david-stewart:
Hey all, I tested this small module in the new `Fun4All_TrackAnalysis.C` that @osbornjd wrote. This module goes through the node `TpcTrackSeedContainer` and calls `seed_container->erase(i_seed)` for every `i_seed` that fails the cuts. When this happens, the track seed container deletes the track contained and sets the pointer to equal a null ptr in the internal data, which is stored in a `vector<TrackSeed*>`. The effect is that the users of the `TpcTrackSeedContainer` need to check for null values.

When used in Fun4All, the `TrackResiduals` sub-module is seg-faulting. Still working on finding the place to fix it

david-stewart:
Hi all (@osbornjd, @hupereir, ...) , I tested this small module in the new Fun4All_TrackAnalysis.C that @osbornjd wrote. This module goes through the node TpcTrackSeedContainer and calls seed_container->erase(i_seed) for every i_seed that fails the cuts. When this happens, the track seed container deletes the track contained and sets the pointer to equal a null ptr in the internal data, which is stored in a vector<TrackSeed*>. The effect is that the users of the TpcTrackSeedContainer need to check for null values.
 
 In any case, I am having significant trouble finding where this filtering in the new module is causing the downstream code to seg-fault. (Currently I have narrowed it down to the TrackResiduals::fillEventTree function, but something must being written to out of bounds memory, because the seg fault isn't occurring in a logical place (i.e. in line `TrkrHitSetContainer::ConstRange tpc_hitsets = hitmap->getHitSets(TrkrDefs::TrkrId::tpcId);`
 
Thoughts?
 
 Thanks

osbornjd:
I would pull down the master branch, or test with the new build. There was a bug in track residuals that required a fix

osbornjd (approved):
Looks good to me, I'm fine with this if you are @david-stewart

david-stewart:
> Looks good to me, I'm fine with this if you are @david-stewart

Hi @osbornjd , I think it works fine for purpose, and it does remove tracks. However, I didn't succeed last night in not getting seg faults in other down-stream modules when there are `nullptr` in `TpcTrackSeedContainer`. Maybe when the updated build goes through it'll be good.
I'm out until Monday, but would be happy to test more, but perhaps it's best to just push this now so that it can get used until then and I can modify latter if need be

PR #3155, opened 2024-09-22
add sEPD calibrated towers

eumaka:
PR add sEPD calibrated towers and makes geometry


pinkenburg:
It looks like you have too many library dependencies in the Makefile.am. In EpdReco you use calobase includes which are resolved by libcalo_io but in the Makefile.am you also have globalvertex_io and trackbase_historic_io. I don't think those are needed. Could you check?

PR #3160, opened 2024-09-25
sEPD reco updates and event plane updates for sEPD

eumaka:
PR updates sEPD for new calibration file
Updates the event plane code for sEPD phi mapping
Makes event plane code light weight (only produces raw Q vectors for production), corrections and histograms filling are decoupled.


pinkenburg:
the data members of Eventplaneinfov1 were changed. Please verify that it has never been saved in our DSTs. Otherwise you have to introduce v2. I make this PR a draft until this is sorted out

blackcathj:
That seems a failed git check out on Jenkins. Let me try to restart it

eumaka:
@pinkenburg no it was never said in our DST

PR #3166, opened 2024-10-01
MVTX modified geometry (coresoftware)

hrjheng:

## Types of changes

- Modify MVTX GEANT geometry based on the alignment parameters (i.e "localAlignmentParamFile.txt")
- A new PHG4MvtxMisalignment class that calculates the global displacements of MVTX
- MakeActsGeometry: fix matching between the ActsSurface and TrkrDef::hitsetkey to account for the global shift of MVTX

Software and simulation meeting on 20241001: https://indico.bnl.gov/event/24940/#14-mvtx-geometry-update

## TODOs (if applicable)

- macros: https://github.com/sPHENIX-Collaboration/macros/pull/956
- calibrations: https://github.com/sPHENIX-Collaboration/calibrations/pull/158


osbornjd:
Looking at these clang-tidy issues locally, they are both warnings and not actual errors so I'm not sure why Jenkins flags these as problematic. Regardless, one of them according to [stack overflow](https://stackoverflow.com/questions/63939799/what-is-clang-tidy-function-is-within-a-recursive-call-chain-how-to-fix-it) can be ignored if you really intend to call the function recursively

pinkenburg:
The recursive is just a warning (it is a possible infinite loop if you call it the wrong way and clang-tidy assumes the worst). If you add this
// NOLINTNEXTLINE(misc-no-recursion)
right before
void PHG4MvtxDetector::SetDisplayProperty(G4LogicalVolume *lv)
it suppresses it
The other is the use of int active. clang-tidy makes a good point that for bit operations you really should not use ints where the sign bit can mess with you. It looks like active can (should) be an unsigned int. As soon as you change that this error should be gone (depending on what m_IsLayerActive returns, if it's int's there are more problems because bit fiddling using a signed int is error prone as well)

hrjheng:
m_IsLayerActive returns int as `std::array<int, n_Layers> m_IsLayerActive{};` and the std::array is set by `m_IsLayerActive[ilayer] = params->get_int_param("active");`.
I'll just add // NOLINTNEXTLINE(misc-no-recursion) for now

osbornjd:
This is the latest build with the macros changes too
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-tracking-low-occupancy-qa/6165/
which show large changes to the QA. We need to understand why this happens as I thought the flags were all set to use the ideal geometry still? So I thought we should get identical performance

hrjheng:
> This is the latest build with the macros changes too https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-tracking-low-occupancy-qa/6165/ which show large changes to the QA. We need to understand why this happens as I thought the flags were all set to use the ideal geometry still? So I thought we should get identical performance

Are you referring to the tpc, tracking, and vertex QAs that show large discrepancies and low KS-test scores?

osbornjd:
Correct, my understanding was that the flags were turned off which effectively leaves the MVTX in its current unmodified state, meaning we would expect no change in the QA. Is my understanding wrong?

hrjheng:
It's my understanding too that MVTX should be in the ideal and default position. Curiously INTT and MVTX QA do not change (not sure about TPOT). The only place that might affect tracking is MakeActsGeometry, but I only modified the part relevant to MVTX, `void MakeActsGeometry::makeMvtxMapPairs(TrackingVolumePtr &mvtxVolume)`. 
I see a few changes to the macro repository that I have not pulled into my branch. I am not sure if it will fix the issue but just want to make sure everything is the same

osbornjd:
Successful build here
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/Build-Master-gcc12/3891/

PR #3169, opened 2024-10-03
(subsystems/intt) Added functionality to retrieve GL1 file names in the InttOdbcQuery class

josephbertaux:

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/packages/intt/InttOdbcQuery.cc
+++ b/offline/packages/intt/InttOdbcQuery.cc
@@ -78,12 +78,11 @@ int InttOdbcQuery::Query(int runnumber)
   odbc::Statement* statement = dbcon->createStatement();
 
   int iret = 0;
-  iret += (QueryStreaming(statement, runnumber) != 0);
-  iret += (QueryType(statement, runnumber) != 0);
-  for(int which_intt = 0; which_intt < 8; ++which_intt)
-  {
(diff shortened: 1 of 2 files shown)


josephbertaux:
Marking as "ready for review" from "draft"
Sorry for the multiple build triggers--there were some niceties I was tempted to put off for another PR but ended up implementing in this one

pinkenburg:
The search query will time out. It tries a match on millions of entries. Our replicas don't handle long request to be able to survive the onslaught by 200k jobs

josephbertaux:
Can it be fixed by modifying the query, by using a nested one like

SELECT ... FROM (SELECT ... FROM filelist WHERE runnumber = ...) WHERE filename LIKE ...?


josephbertaux:
I think not. Could it be solved by maintaining sections of this information in smaller tables for runnumber ranges, similar to how we have subdirectories for the production DST files?

pinkenburg:
The daq DB is just the wrong place to look this up. The field if a file is on disk or not is not kept up to date. What you see is the state when the file was transferred. If a file gets removed or staged from tape the daq DB does not reflect this. The production hands down lists of files (with full path or filename which then gets looked up in our file catalog), there is no need to have this functionality in our code. We need a catalog for the raw data files which then reflects what is on disk and in hpss, but creating this is currently somewhat lower priority since the production can do without.

PR #3173, opened 2024-10-04
Jet DST skimmer

Shuonli:

## Types of changes

This is the initial commit to the module for making the slimed(skimmed?) DST. Where it will search for max Jet energy and EMCal cluster energy per event and only keep the event that contains high energy jets/clusters.

I think it will also good to have the background event (beam background) rejection implemented in this. I have left a place holder here.

## TODOs (if applicable)

Implement the Beam background ID function.


osbornjd:
Can I suggest that we organize this differently? Let's have a directory structure that looks like
offline
-- packages
---- Skimmers
------- Jets
---------- JetSkimmer.cc/h/makefile/etc

Then we can make other directories for different types of skimmers

bseidlit:
@osbornjd  Don't we want  "slimmer" not "skimmer"?  We are taking out events not removing information for each event.

osbornjd:
I don't think we ever decided on a specific name - every collaboration has their own but I don't think we came to a conclusion. Slimmer and skimmer could in principle refer to the same thing - either you slim the DST down to only include the events you want or you are skimming the larger DST to write out only the ones you want

eumaka:
Sorry I missed the last software meeting. What ever name is decided, these DSTs are meant to skip events that don't meet a certain cut, e.g. jet pT etc...?

osbornjd:
They're intended to be used to write out smaller DSTs that meet certain criteria

bseidlit:
Okay, I don't have strong opinions.  I think in phenix they called them mini DSTs.

bseidlit:
It was decided in the software meeting to call it a skimmer.

PR #3178, opened 2024-10-11
Adding truth matching for simulated version of TowerInfo object, and downstream reco/eval

Shuonli:

## Types of changes

This PR adds the infrastructure for having the truth matching(from the g4eval) for the TowerInfo objects.

Two new set of derived classes are implemented here: 
`TowerInfoSimv1` which is derived from `TowerInfov2` with two maps of energy contributions from `PHG4Hit` and `PHG4Shower` and getter/setter of them.

`TowerInfoSimv2` derived from `TowerInfoSimv1` that contains the waveform information(which is the object that will be constructed in the waveformsim module).

The related g4eval modules are also modified to use towerinfo for truth matching.

`CaloWaveformSim` and `CaloTowerBuilder` are also updated.

### Something to think about
According to [this](https://root-forum.cern.ch/t/tclonesarray-and-tobject-containing-a-pointer-to-std-vector/11344) conversation, it seems like the authors are not a big fan of dynamic allocation in  `TCloneArrays` , but I guess I will give this a try anyways and if there is memory/performance issues we can find a workaround.

I could have used template definition in the towereval, but I didn't for the code simplicity. Also since the ways that we handle RawTowerContainer and TowerInfoContainer are very different, I don't think the code will look much prettier even if I change to template functions(that being said, I can change it if needed).

## TODOs (if applicable)

~~If this seems okay I will propagate this to g4eval~~(done) and test it.

~~I will make this a draft for now before I make the changes in eval modules.~~

I passed on changes to the eval, now I need to test it...

~~I also need to put the truth info in in `CaloWaveformSim`  and `CaloTowerBuilder` module~~ done

This would also need a macro change


Shuonli:
Weird, it seems like the `TowerInfoSimv2.h` used in Jenkins' check is different than the one displayed on git...I will try to run the test again..

PR #3189, opened 2024-10-17
EPD reco update: replace the hardcoded equal distance phi value arrays with human readable calculation

Shuonli:

## Types of changes

I believe

```cpp
float EpdReco::GetTilePhi(int thisphi) {
  static const float tilephi[24] = {
      0.13089969, 0.39269908, 0.65449847, 0.91629786, 1.17809725, 1.43989663,
      1.70169602, 1.96349541, 2.2252948,  2.48709418, 2.74889357, 3.01069296,
      3.27249235, 3.53429174, 3.79609112, 4.05789051, 4.3196899,  4.58148929,
      4.84328867, 5.10508806, 5.36688745, 5.62868684, 5.89048623, 6.15228561};
  return tilephi[thisphi];
}
```

is (almost)equivalent to:

```cpp
float EpdReco::GetTilePhi(int thisphi) {
  return ((2 * thisphi + 1) * M_PI) / 24;
}
```

The later also doesn't segfault if you give some index out of range...

I think this could be a potential improvement of the code 👀

## TODOs (if applicable)

### DIFF
--- a/offline/packages/epd/EpdGeom.h
+++ b/offline/packages/epd/EpdGeom.h
@@ -3,16 +3,12 @@
  Re-written by Ejiro Umaka 03/28/23
 */
 
-#ifndef EPD_GEOM_H
-#define EPD_GEOM_H
+#ifndef EPD_EPDGEOM_H
+#define EPD_EPDGEOM_H
 
 #include <phool/PHObject.h>
 
-#include <vector>
-#include <utility>
-#include <tuple>
-#include <iostream>
-#include <cmath>
+#include <limits>
 
 class EpdGeom : public PHObject
 {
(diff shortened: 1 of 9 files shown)


eumaka:
Closing this PR. Until the sEPD geometry is finalized, I’d like to keep these arrays as is. I would want this to break if something goes out of bound. This will be indicative that something went wrong with our mapping in TowerinfoDefs

Shuonli:
Interesting, if you **really** want to use segmentation fault and undefined behavior to test your mapping that's understandable. And there are indeed many better ways to code this up as oppose to just put a bunch of hard-coded float in a static array for instance:

```cpp
float EpdReco::GetTilePhi(int thisphi) {
  static float tilephi[24] = {0};  // Static array to store precomputed values
  if (tilephi[0] == 0) {  // Initialize only if the array is not yet filled
    for (int i = 0; i < 24; ++i) {
      tilephi[i] = ((2 * i + 1) * M_PI) / 24;  // Precompute the values
    }
  }
  return tilephi[thisphi];  // Return precomputed value
}
```

In this case, even the code will segfault and have undefined behavior as it is(and lack of comments), the code will be still remain understandable for other people in the collaboration...

I really don't think:
```
float EpdReco::GetTilePhi(int thisphi) {
  static const float tilephi[24] = {
      0.13089969, 0.39269908, 0.65449847, 0.91629786, 1.17809725, 1.43989663,
      1.70169602, 1.96349541, 2.2252948,  2.48709418, 2.74889357, 3.01069296,
      3.27249235, 3.53429174, 3.79609112, 4.05789051, 4.3196899,  4.58148929,
      4.84328867, 5.10508806, 5.36688745, 5.62868684, 5.89048623, 6.15228561};
  return tilephi[thisphi];
}
```
without any documentation
is the kind of code that will be helpful for people other than the original author to understand what's going on...

eumaka:
As I said before I closed this PR, those values of the tile locations are bound to change. This is a simplification of the tile locations. When we implemented this two plus years ago (remember this was first implemented in the simulation), Chris asked for the use of arrays, which we all knew will change when there is a survey. We said we will then read in the geometry with a text file (there was no CDBTTree then). Changes to the phi values implementation is moot, I see that I unsuccessfully made that point earlier. I also see that the proposed change leaves the r values array as is. This is a headache that returns no fruit.

PR #3200, opened 2024-10-23
set ZS status for uniform software ZS tower, no magic number for timing and chi2

Shuonli:

## Types of changes

Currently if we are doing the uniform software ZS, the suppression is done in *CaloWaveformFitting* and after suppression the time of the tower will be set to -20 without setting the ZS status. However, the calotower status will check the timing for non-ZS tower.

This PR change the calotowerbuilder and waveformfitting, so the hardware and software ZS towers are set with time and chi2 equals to qnan. Also let calotowerbuilder to check for if the timing and chi2 is qnan(for SZS) to set the ZS status for software zero suppressed towers.

This PR also add case handling for Towerinfov4 where the chi2 is set into log scale, for quite nan case the lnchi2 is set to 0 and we can get the quite nan back while getting chi2.

Something to note that if we change the time for ZS tower this fix will stop working(but I don't think we plan to change it). 

## TODOs (if applicable)

### DIFF
--- a/offline/packages/CaloBase/TowerInfov4.h
+++ b/offline/packages/CaloBase/TowerInfov4.h
@@ -4,6 +4,7 @@
 #include "TowerInfov1.h"
 
 #include <cmath>
+#include <limits>
 
 class TowerInfov4 : public TowerInfo
 {
@@ -30,7 +31,7 @@ class TowerInfov4 : public TowerInfo
     
     if (std::isnan(_chi2))
     {
-      lnChi2 = 1;
+      lnChi2 = 0;
     }
     else if (_chi2 <= 0)
     {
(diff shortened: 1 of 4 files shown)


pinkenburg:
Why is the time and chi2 set to a magic value for ZS towers (-20,0)? Direct float comparisons (time == -20) should always be avoided, why aren't we using nan for those invalid values (like everywhere else where we flag invalid floats/doubles) where this would be replaced by std::isfinite()

bseidlit:
@pinkenburg this looks good to me.  Is this what you intended?

pinkenburg:
> @pinkenburg this looks good to me. Is this what you intended?

yes - it is, the invalid values are preserved and restored during the compression, so there is no chance for mixing this up by accident

PR #3201, opened 2024-10-23
Add Intt Event Mixup QA

MAIKANO1:
bcocheck module make INTT BCO difference plot (BCO_Full - BCO) for getting trigger timing peak.
InttMixupQA module analyze hit mixed up from previous event.
## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/QA/Intt/InttMixupQA.cc
+++ b/offline/QA/Intt/InttMixupQA.cc
@@ -1,0 +1,1108 @@
+#include "InttMixupQA.h"
+
+using namespace std;
+
+InttMixupQA::InttMixupQA(const string &name, const int run_num,const int felix_num)
+  : SubsysReco(name)
+{
+  run_num_ = run_num;
+  felix_num_=felix_num;
+  cout<<"felix_num_="<<felix_num_<<"felix_num="<<felix_num<<endl;
+}
+
+InttMixupQA::~InttMixupQA() {}
+
+int InttMixupQA::Init(PHCompositeNode * /*topNode*/)
+{
+  if (Verbosity() > 5) {
+    std::cout << "Beginning Init in InttMixupQA" << std::endl;
+  }
+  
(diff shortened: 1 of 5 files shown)


osbornjd (changes requested):
This needs to be modified to use the QA framework for file handling so that all QA histograms go to the same place when the module is used in the production. 

See [here](https://github.com/sPHENIX-Collaboration/coresoftware/blob/bcb9d79c241df26e7f4a55259208d2c719763cb9/offline/QA/Calorimeters/CaloValid.cc#L753-L759) for an example of how to register QA histograms with the manager, and then you can get your macro to dump the histograms out using some code like [this](https://github.com/sPHENIX-Collaboration/macros/blob/5aa95507a37c815da98e21398cb0d35af91b6731/TrackingProduction/Fun4All_FieldOnAllTrackers.C#L395-L400)

josephbertaux:
I fixed some of the issues locally--I spoke with Mai and only 4 of the (arrays) of histograms are to be plotted.
I introduced name mangling and added them to a QAHistManager, she's merged the changes to her branch. I'm leaving most of it alone and will helper her with a plotting macro soon

osbornjd (changes requested):
I fixed the clang-tidy issues from the CI but this still needs some work. You register the histograms now with the histogram manager which is great, but the only thing these modules should do is create the histograms, register them with the histogram manager, and then fill those histograms. The histogram manager will take care of the file creation and writing of the histograms for you. Then, the drawing of those histograms is delegated to a different module in [github.com/sPHENIX-Collaboration/QAhtml ](https://github.com/sPHENIX-Collaboration/QAhtml/). That means these modules should not do any drawing or other file creation - that is delegated to QAhtml.

Also in general you should not use `Form` and instead use the `boost::format` to create names of things. When we switch to c++20 soon then we can use `std::format`

PR #3211, opened 2024-10-29
feat: automate mvtx pooling bco range based on strobe length

osbornjd:
This PR generalizes the daq db strobe lookup to a helper class in the mvtx decoder such that both the SinglePoolInput and unpacker can call it. This way the strobe width can be accessed from the daqdb (as is already done in the unpacker) in the pooling, and the bco ranges can be set accordingly. This should completely automate the production of the MVTX data in any mode that currently exists - 89 or 10 mus streaming strobe, or in triggered mode.

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/framework/fun4all/InputFileHandler.h
+++ b/offline/framework/fun4all/InputFileHandler.h
@@ -20,6 +20,7 @@ class InputFileHandler
   void IsOpen(const int i) { m_IsOpen = i; }
   void SetVerbosity(const int i) { m_Verbosity = i; }
   int GetVerbosity() const { return m_Verbosity; }
+  const std::list<std::string>& GetFileList() const { return m_FileList; }
   void UpdateFileList();
   void FileName(const std::string &fn) { m_FileName = fn; }
   const std::string FileName() const { return m_FileName; }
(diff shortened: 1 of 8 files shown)


cdean-github (approved):
I think this looks good to me. 

If I had to suggest a change, it would be to set the BCO ranges based on the ratio of the strobe length to the BCO time (e.g. std::ceil(89 / 0.1065)) but this would return values that arent nice numbers like 100, or may require scaling to fit in the TPC extended readout window of 50 us.

Also, the MVTX can run at other strobe lengths (40, 20 and 5 us). We havent run much at 5 us but if we do this could cause issues in the code as there are catches for a stobe length > 9 us and < 1 us

osbornjd:
> If I had to suggest a change, it would be to set the BCO ranges based on the ratio of the strobe length to the BCO time (e.g. std::ceil(89 / 0.1065)) but this would return values that arent nice numbers like 100, or may require scaling to fit in the TPC extended readout window of 50 us.

I was thinking this too but wasn't sure if we wanted whole nice numbers, since the purpose isn't necessarily to have an exact range but just to find the matching GL1s. I'm happy to make the change if you think it is worthwhile, I just used the current existing (or those we have used) bco ranges for the given strobe windows.

> Also, the MVTX can run at other strobe lengths (40, 20 and 5 us). We havent run much at 5 us but if we do this could cause issues in the code as there are catches for a stobe length > 9 us and < 1 us

If there are suggestions for the bco ranges to use for these, I can just add a catch in for them. Or I can change the logic such that if there is a "new" strobe length (e.g. 40) then the bco ranges default to `std::ceil(89 / 0.1065)`

PR #3213, opened 2024-10-30
Changing location of line lasers to match reality

chughes90:
**Files Affected:**

- simulation/g4simulation/g4tpc/PHG4TpcDirectLaser.cc

**Changes:**

- L419, 425 - adjusting the sign of the angle so that it rotates the laser locations to proper angles

**TODO:**

- check with Ross that SS is right
- rebuild macros and confirm that its right


chughes90:
image

PR #3219, opened 2024-10-31
fix: clean up gl1 pool vectors

osbornjd:
This cleans up the GL1 pool vectors/maps better than before, and appears to lead to more stable CPU processing time as reported in today's tracking meeting.

https://indico.bnl.gov/event/25251/

Marking draft as I don't totally understand *why* this makes the speed more stable...

@hupereir 

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/framework/fun4allraw/SingleGl1PoolInput.cc
+++ b/offline/framework/fun4allraw/SingleGl1PoolInput.cc
@@ -205,7 +205,6 @@ void SingleGl1PoolInput::CleanupUsedPackets(const uint64_t bclk)
       {
         delete pktiter;
       }
-      iter.second.clear();
       toclearbclk.push_back(iter.first);
     }
     else
--- a/offline/framework/fun4allraw/SingleGl1PoolInput.cc
+++ b/offline/framework/fun4allraw/SingleGl1PoolInput.cc
@@ -108,7 +108,6 @@ void SingleGl1PoolInput::FillPool(const unsigned int /*nbclks*/)
 
     Gl1Packet *newhit = new Gl1Packetv2();
     uint64_t gtm_bco = packet->lValue(0, "BCO");
-    m_BeamClockFEE.insert(gtm_bco);
     m_FEEBclkMap.insert(gtm_bco);
     newhit->setBCO(packet->lValue(0, "BCO"));
     newhit->setHitFormat(packet->getHitFormat());
@@ -165,13 +164,7 @@ void SingleGl1PoolInput::FillPool(const unsigned int /*nbclks*/)
 
 void SingleGl1PoolInput::Print(const std::string &what) const
 {
-  if (what == "ALL" || what == "FEE")
-  {
-    for (const auto &bcliter : m_BeamClockFEE)
-    {
-      std::cout << PHWHERE << "Beam clock 0x" << std::hex << bcliter << std::dec << std::endl;
-    }
-  }
+  
   if (what == "ALL" || what == "FEEBCLK")
   {
     for (auto bcliter : m_FEEBclkMap)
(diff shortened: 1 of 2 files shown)


offline/framework/fun4allraw/SingleGl1PoolInput.cc
```diff
@@ -204,30 +197,28 @@ void SingleGl1PoolInput::Print(const std::string &what) const
...
-  for (const auto &iter : m_Gl1RawHitMap)
+  for (auto &iter : m_Gl1RawHitMap)
   {
     if (iter.first <= bclk)
     {
       for (auto pktiter : iter.second)
       {
         delete pktiter;
       }
+      iter.second.clear();
```
hupereir, line 208: Joe, this should really not be necessary, since the actual iterator is removed from the map at line 222.
erasing the iterator will clear both members of the pair, making the call above redundant.

offline/framework/fun4allraw/SingleGl1PoolInput.cc
```diff
@@ -108,7 +108,6 @@ void SingleGl1PoolInput::FillPool(const unsigned int /*nbclks*/)
 
     Gl1Packet *newhit = new Gl1Packetv2();
     uint64_t gtm_bco = packet->lValue(0, "BCO");
-    m_BeamClockFEE.insert(gtm_bco);
```
hupereir, line 111: Joe, do I understand right that you removed m_BeamClockFEE entirely because it is 100% redundant with m_FEEBlckMap ? (that's my impression indeed when reading the code)

offline/framework/fun4allraw/SingleGl1PoolInput.cc
```diff
@@ -204,30 +197,28 @@ void SingleGl1PoolInput::Print(const std::string &what) const
...
   }
-  // for (auto iter :  m_BeamClockFEE)
-  // {
-  //   iter.second.clear();
-  // }
+
 
   for (auto iter : toclearbclk)
   {
+    m_FEEBclkMap.erase(iter);
```
hupereir, line 220: Good catch. Definitly somethign that was missing. It should have created a (small) leak, but more importantly, filling the map would become increasingly slow over time, because new gtm_bco will be inserted in an ever increasing map.

hupereir (commented):
Hi Joe, 
few comments attached. 
Hugo

offline/framework/fun4allraw/SingleGl1PoolInput.cc
osbornjd: I thought I was told or read somewhere (can't remember where now) that vectors should also be cleared and that calling a map.clear does not necessarily properly clean up the memory of a vector that is in the map. I'm probably misremembering... let me see if I can find anything on stack overflow, but I can remove this

offline/framework/fun4allraw/SingleGl1PoolInput.cc
osbornjd: I mean m_BeamClockFEE...

offline/framework/fun4allraw/SingleGl1PoolInput.cc
hupereir: Joe, when calling erase(iterator), the 'object' pointed to by the iterator is deleted. In that case, a std::pair<uint64_t, std::vector<Gl1Packet *>>. Both members of the pair are deleted. When a vector is deleted, it is cleared. Hence the redundancy. We rely on that at many places in our code (I can point you to them on demand).
I would really recommend not calling the clear explicitly here, and revert to const& in the for loop, as this is bad practice.

hupereir: If calling the 'clear' here makes any difference, I would like to see numbers/tests to back it up, because this is against what I learned about c++.

hupereir: oops. I missed that. Sorry for the noise !

PR #3232, opened 2024-11-06
Added Function to skip events with > 100k TPCseeds

pedroanietom:
[comment]: <I implemented the set_max_seeds function that aborts events with more than 100k TPC seeds.> (Please tell us something about this pull request)

## Types of changes

[comment]: <If you call the function cprop->set_max_seeds(true); in the Fun4All_FieldOnAllTrackers.C macro, it aborts events that have more than 100k TPC seeds coming from the seeder process. Discussion in: https://indico.bnl.gov/event/25255/> ( What does this PR do? Linking to talk in software meeting encouraged )

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackreco/PHSimpleKFProp.cc
+++ b/offline/packages/trackreco/PHSimpleKFProp.cc
@@ -241,7 +241,7 @@ int PHSimpleKFProp::process_event(PHCompositeNode* topNode)
     std::cout << "prepared KD trees" << std::endl;
   }
 
-  if (Verbosity()==0)
+  if (Verbosity())
   {
     std::cout << "number of TPC seeds: " << _track_map->size() << std::endl;
   }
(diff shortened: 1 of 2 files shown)


pinkenburg:
Is there a reason why the 100k is hardcoded? The flag to enable it could be used to set this limit, I am sure we need to play with this value

osbornjd:
In the future we won't have this - this is just a stop gap to prevent the production from blowing up in memory and/or slow down. We need to spend time at the lower level seeding algorithm and also in the hit unpacking where there are events that are super noisy, but those are longer term developments.

pedroanietom:
I decided to go with 100k since the normal distribution for p+p collisions averages 1.3k TPC seeds and for Au+Au it can be up to 900 times more for central collisions. So I left it at 100k, that would cover most AuAu events. 10k is not enough maybe.

pinkenburg:
This version allows to set the maximum number of seeds before aborting

PR #3245, opened 2024-11-14
Move cdb call to initrun function of CaloFittingQA. Add QA histograms for calorimeter single sample pedestal.

emclaughlin2:

## Types of changes

This PR moves the call to the CDB for the emcal adc skip mask to the init function from the process event function so that the CDB call happens only once. The PR also adds new QA histograms to monitor the calorimeter single sample pedestal. 

## TODOs (if applicable)

### DIFF
--- a/offline/QA/Calorimeters/CaloFittingQA.cc
+++ b/offline/QA/Calorimeters/CaloFittingQA.cc
@@ -68,6 +68,20 @@ int CaloFittingQA::Init(PHCompositeNode* /*unused*/)
 
   createHistos();
 
+  if (m_debug)
+  {
+    std::cout << "Leaving CaloFittingQA::Init" << std::endl;
+  }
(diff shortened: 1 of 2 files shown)


pinkenburg:
The run number is not known when the Init() is executed (Init() is executed during the se->registerSubsystem() call in the macro). That means the timestamp which is used to look up the calibration is also not set yet and the cdb will return calibrations for time=0. If you move this to the InitRun() method then the timestamp will be set correctly. Marking this PR as draft until this is sorted out

bseidlit:
I forgot about this oddity.  Are there things that really should be init and not initrun?

pinkenburg:
Under normal circumstances I can't think of anything which has to go into Init() rather than InitRun(), at least for this module there is no reason to implement Init()

PR #3259, opened 2024-11-20
Updates to add zero_field utility to the track matching

david-stewart:

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackbase/TrackFitUtils.cc
+++ b/offline/packages/trackbase/TrackFitUtils.cc
@@ -744,8 +744,8 @@ double TrackFitUtils::z_fit_to_pca(const double xy_slope, const double xy_interc
   std::vector<std::pair<double,double>> zd_vec;
   for (auto& glob : glob_pts) {
     auto point = get_line_point_pca(xy_slope, xy_intercept, glob); // point = point on line
-    auto dist = sqrt(square(pca.x()-point.x())+square(pca.y()+point.y()));
-    zd_vec.push_back({dist,glob.z()});
+    double dist = sqrt(square(pca.x()-point.x())+square(pca.y()+point.y()));
+    zd_vec.emplace_back(dist,glob.z());
   }
   auto slope_and_b = line_fit(zd_vec);
   /* double zd_slope; */
--- a/offline/packages/trackbase/TrackFitUtils.cc
+++ b/offline/packages/trackbase/TrackFitUtils.cc
@@ -692,7 +692,7 @@ std::vector<float> TrackFitUtils::fitClustersZeroField(std::vector<Acts::Vector3
 void TrackFitUtils::getTrackletClusters(ActsGeometry* _tGeometry,
                                         TrkrClusterContainer* _cluster_map,
                                         std::vector<Acts::Vector3>& global_vec,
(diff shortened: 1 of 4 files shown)


offline/packages/trackbase/TrackFitUtils.cc
```diff
@@ -862,3 +870,68 @@ float TrackFitUtils::get_helix_surface_pathlength(const Surface& surf, std::vect
...
+// phi, eta, pT, pos_dca, momentum -- note that momentum is such that pT\equiv1
+std::tuple<bool, double, double, double, Acts::Vector3, Acts::Vector3> 
+TrackFitUtils::zero_field_track_params(
+    ActsGeometry* _tGeometry, 
+    TrkrClusterContainer* _cluster_map, 
+    const std::vector<TrkrDefs::cluskey>& cluskey_vec)
+{
+  std::vector<Acts::Vector3> global_vec;
+
+  if (global_vec.size()<2)
```
hupereir, line 891: Dave, 
Nice commit !
I'm confused though: global_vec is declared on line 889, empty. So the test here is always satisfied. No ? What do I miss ?

david-stewart: I just saw this! It is an error, and I have moved it to after the call that is supposed to fix it.

offline/packages/trackbase/TrackFitUtils.cc
```diff
@@ -862,3 +870,68 @@ float TrackFitUtils::get_helix_surface_pathlength(const Surface& surf, std::vect
...
+  Acts::Vector3 cluster = global_vec.back();
+  auto clus_dca_xy = get_line_point_pca(xy_m, xy_b, {cluster.x(),cluster.y(),0});
+  double px = clus_dca_xy.x();
+  double py = clus_dca_xy.y();
+  double pz = xz_m * x + xz_b;
+
+  // correct to direction from the pca to origin
+  px -= x;
+  py -= y;
+  pz -= z;
```
xuanli23, line 926: For this commit, the pz will be always 0, is this in purpose or a typo?

osbornjd:
Marked as draft while other comments are being resolved

osbornjd (commented):
One other question/comment - this PR indicates we will not be actually introducing an object to store zero field track parameters and rather be recalculating them on the fly with the function in `TrackFitUtils` whenever they're needed. I don't have an issue with this because the zero field running is a niche "calibration" case which doesn't have stringent CPU requirements, but I wanted to make sure that was the plan moving forward as we should then replace the evaluation zero field machinery with this call to properly calculate the straight line fit track parameters

sPHENIX-GitHub-l:
Hi Joe,

Yes, that is the plan. The zero field running requires changes in only two places (track matching and vertex finding). It did not seem to justify complicating the seed storage, given that we do not need to process much zero field data.

Tony


offline/packages/trackbase/TrackFitUtils.cc
david-stewart: This is a typo. Line 921 should use `px` instead of `x`. Thanks, this is a good catch

PR #3290, opened 2024-12-09
BCOOffset is loaded from CDB and set it in combiner in order to get p…

gwd213:
…roper time-bucket info in trigger mode

1. Adding the feature to load BCOoffset from CDB database and applying to Combiner in order to get proper Timing OffSet.
2. BCOFiltering is set to OFF by default, since now that is one of offline analysis cut no longer necessary in general. Especially streaming and extended readout trigger mode should not filter the data based on BCO value.

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/packages/intt/InttBCOMap.cc
+++ b/offline/packages/intt/InttBCOMap.cc
@@ -6,6 +6,8 @@
 
 #include <filesystem>
 #include <iostream>
+#include <vector>
+#include <algorithm>
 
 InttBCOMap::InttBCOMap()
 {
@@ -64,7 +66,6 @@ int InttBCOMap::LoadFromFile(std::string const &filename)
 
 int InttBCOMap::LoadFromCDBTTree(CDBTTree &cdbttree)
 {
-  ///////////////
   std::cout << "LoadFromCDBTTree::LoadFromCDBTTree" << std::endl;
 
   uint64_t N = cdbttree.GetSingleIntValue("size");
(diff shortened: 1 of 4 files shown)


osbornjd (commented):
Have we checked how this changes the downstream tracking crossing assignments? For the streaming data, this is particularly important. Is the value of 23 offset read in from the CDB for all streaming runs? I guess this shouldn't actually change the crossing distribution since the time bucket is not modified, there is only an added check on whether or not certain hits should be continued to be processed. Maybe I am misunderstanding and this is only necessary for the timing offset calibration for the triggered data runs.

gwd213:
@osbornjd In summary, it does not affect on streaming run production, but that is necessary to remove bco filtering
 
https://github.com/sPHENIX-Collaboration/coresoftware/pull/3290/files#diff-4ac08cbc75b958f769b7e08b33c3a43cead8ca838fc1503b22a62056ab6509c6L260-R264


As you described, in streaming readout, our offset value 23 is not changed. So, we don't actually need to read CDB for streaming runs. 
Here, if m_triggedMode == true ( false by default), we change the offset from 23 to value from CDB.
https://github.com/sPHENIX-Collaboration/coresoftware/pull/3290/files#diff-4ac08cbc75b958f769b7e08b33c3a43cead8ca838fc1503b22a62056ab6509c6R152-R155


For analysis cut purpose, we have used that cut to check hit only associated with GL1, but it should not apply for pp runs both extended readout(trigger mode) and streaming mode.
So, we put m_bcoFilter (false by default) in order to disable this function, but want to keep it for some special analysis.

Have we checked how this changes the downstream tracking crossing assignments? 
-> I also check it in hitunpacking level so not downstream. But for sure, I'd like to check the detail of it in trigger mode. If you have idea how to check, could you guide me which part I am supposed to check?

gwd213:
[offset_check.pdf](https://github.com/user-attachments/files/18073800/offset_check.pdf)
I just checked the INTT cluster time-bucket distribution for both streaming and trigger data, with and without the new PR.
If we don’t see significant issues with the INTT cluster time-bucket, I think we can safely conclude that the new PR does not harm the downstream processes further.

The first slide shows the cluster time-bucket distribution in streaming readout, without and with the BCO CDB, respectively.
As you also guessed, we don’t see any difference between them since we already set 23 as the default offset, which has not been modified.

The second slide shows the same distribution, but in trigger mode.
Without the BCO CDB and the new PR, the peak of the cluster time-bucket distribution is not properly aligned at 0 because the offset is not calibrated.
With the BCO CDB and the new PR, the peak of the cluster time-bucket distribution shifts to 0, where it is supposed to be.

osbornjd:
Thanks for the quick study! That is very helpful, and thanks for doing the development of this and the subsequent checks also.

PR #3329, opened 2025-01-06
Change Status Check for StreakSidebandFilter

jnclement:
Changes the status that the streak filter checks to determine if a tower is good or bad. Previously, checked for 1 to determine goodness (wrong), now checks for 0 (good) or 32 (zero suppressed and good).
## Types of changes

This is a bug fix. As above, I have changed which statuses indicate that a tower is good to match the definitions currently in core software. Previously this file checked that the status was NOT 1 in order to determine that the tower was bad. It should check for NOT 0 and NOT 32, as these two statuses are good towers.

## TODOs (if applicable)


pinkenburg:
I am not sure about the check for the status. I thought the status is a bit field, but then we shouldn't compare this to a number (32) but rather if a given bit is set.

PR #3347, opened 2025-01-14
feat: segmentized qa files attempt 2

osbornjd:
Hugo said this one worked for him without an issue, let's see what Jenkins says...

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/framework/fun4all/Fun4AllServer.cc
+++ b/offline/framework/fun4all/Fun4AllServer.cc
@@ -750,18 +750,20 @@ int Fun4AllServer::process_event()
   }
   if(!HistoManager.empty() && !eventbad && writing)
   {
-    for( auto histit = HistoManager.begin(); histit != HistoManager.end(); ++histit)
+    for( const auto& histit : HistoManager)
     {
-      if((*histit)->dumpHistoSegments())
+      if((*histit).dumpHistoSegments())
       {
         if(Verbosity() > 0)
         {
-          std::cout << PHWHERE << (*histit)->Name() << " wrote events, closing " << (*histit)->OutFileName() << std::endl;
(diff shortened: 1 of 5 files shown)


osbornjd:
Okay, this PR at least seems to corroborate @hupereir observation that the code does actually work at this point without any of the `#ifdef __CLING__` statements because the jupyter kernels crash, but the actual macros succeed. So there must be some issue in the default behavior that changed from the code addition
https://web.sdcc.bnl.gov/jenkins-sphenix/job/sPHENIX/job/test-tracking-low-occupancy-qa/6669/console

osbornjd:
@hupereir @pinkenburg this worked, so apparently there was some clash with the macros when I originally submitted this PR

offline/framework/fun4all/Fun4AllServer.cc
```diff
@@ -731,6 +734,8 @@ int Fun4AllServer::process_event()
             MakeNodesTransient(runNode);  // make all nodes transient by default
             (*iterOutMan)->WriteNode(runNode);
             (*iterOutMan)->RunAfterClosing();
+            segment = static_cast<Fun4AllDstOutputManager*>((*iterOutMan))->Segment();
```
pinkenburg, line 737: I would prefer to add this method to the base class instead of a static cast here. We can run other output managers here (where one of the raw data output managers also has seg
ments). They would likely crash here

osbornjd: Fair enough, I wasn't sure if there was an explicit e.g. historical reason this was in the derived class. I moved it in this commit
https://github.com/sPHENIX-Collaboration/coresoftware/pull/3347/commits/5bd6fec09dc8881b7ce703e4ea8b2ba54ccb1136

PR #3359, opened 2025-01-17
Remove pT-dependent magnification of track matching windows

david-stewart:
This is a simple change to set constant windows for the track matching between the Silicone and Tpc.
I can put in something more sophisticated, but think this is all that we want at this point.
The only point that I am a bit concerned about is the effect of the bit of code using the 
`silicon_match->set_use_old_matching(true)` in some macros, such as the `calibrations/tpc/TpcDVCalib/Fun4All_FieldOnAllTrackersCalos.C` -- for which the the phi window isn't enlarged.

- also set the phi search windows to use two parameters, an upper and lower bound. (All the other windows use a single bound for |DeltaVal|<bound)

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackreco/PHSiliconTpcTrackMatching.cc
+++ b/offline/packages/trackreco/PHSiliconTpcTrackMatching.cc
@@ -62,7 +62,7 @@ int PHSiliconTpcTrackMatching::InitRun(PHCompositeNode *topNode)
                       "event:sicrossing:siq:siphi:sieta:six:siy:siz:sipx:sipy:sipz:tpcq:tpcphi:tpceta:tpcx:tpcy:tpcz:tpcpx:tpcpy:tpcpz:tpcid:siid");
   }
   // put these in the output file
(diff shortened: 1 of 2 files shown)


osbornjd (commented):
We should discuss what the best way to handle this is, because it kills the low pT matching efficiency in simulation where we have good alignment etc. In principle once we have the detector well aligned we'd probably want to go back to using an inflation factor so I don't know if we should just comment things out here. Maybe we have a `isData` setting which configures things for data that can be set at the macro level? @adfrawley thoughts?

david-stewart:
> We should discuss what the best way to handle this is, because it kills the low pT matching efficiency in simulation where we have good alignment etc. In principle once we have the detector well aligned we'd probably want to go back to using an inflation factor so I don't know if we should just comment things out here. Maybe we have a `isData` setting which configures things for data that can be set at the macro level? @adfrawley thoughts?

I will ask to close this. After talking with Tony, who wanted a pT dependent cut for Z and eta, and then the offset window for phi, then I started over with the new pull request.

PR #3373, opened 2025-01-23
Update matching windows to reflect study on current ranges

david-stewart:
These include a constant (w.r.t. pT) window for x, y A Q and pT dependent window for Z, and eta
A offcentered-window for phi

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackreco/PHSiliconTpcTrackMatching.cc
+++ b/offline/packages/trackreco/PHSiliconTpcTrackMatching.cc
@@ -73,6 +73,21 @@ int PHSiliconTpcTrackMatching::InitRun(PHCompositeNode *topNode)
   std::istringstream stringline(m_fieldMap);
   stringline >> fieldstrength;
 
+  // initialize the WindowMatchers
+  if (_use_legacy_windowing) {
+    window_dx.set_use_legacy(_x_search_win, this);
+    window_dy.set_use_legacy(_y_search_win, this);
+    window_dz.set_use_legacy(_z_search_win, this);
+    window_deta.set_use_legacy(_eta_search_win, this);
+    window_dphi.set_use_legacy(_phi_search_win, this);
+  } else {
+    window_dx.init_bools();
+    window_dy.init_bools();
+    window_dz.init_bools();
+    window_dphi.init_bools();
+    window_deta.init_bools();
+  }
+
   return ret;
 }
 
@@ -85,6 +100,39 @@ void PHSiliconTpcTrackMatching::SetDefaultParameters()
   // http://skipper.physics.sunysb.edu/~prakhar/tpc/HTML_Gases/split.html
 
   return;
+}
+
+void PHSiliconTpcTrackMatching::WindowMatcher::init_bools() {
+  only_fabs = (negL[0]==100. || posL[0] == 100.);
+  all_pos_Q = (negR[0]==100.);
+  if (posR[0]==100.) { use_legacy = true; }
+  negL_b0 = (negL[1]==0.); 
+  negR_b0 = (negR[1]==0.); 
+  posL_b0 = (posL[1]==0.); 
+  posR_b0 = (posR[1]==0.); 
+}
+
+bool PHSiliconTpcTrackMatching::WindowMatcher::in_window(bool posQ, double pt_tpc, double pt_si) 
+{
+  const auto delta = pt_tpc-pt_si;
+  if (use_legacy) {
+    return fabs(delta)*parent_ptr->getMatchingInflationFactor(pt_tpc) < leg_search_win;
+  }
+  if (posQ || all_pos_Q) {
+    if (only_fabs) {
+      return fabs(delta) < fn_exp(posR, posR_b0, pt_tpc);
+    } else {
+      return (delta > fn_exp(posL, posL_b0, pt_tpc) 
+           && delta < fn_exp(posR, posR_b0, pt_tpc));
+    } 
+  } else { // negative Q track
+    if (only_fabs) {
+      return fabs(delta) < fn_exp(negR, negR_b0, pt_tpc);
+    } else {
+      return (delta > fn_exp(negL, negL_b0, pt_tpc) 
+           && delta < fn_exp(negR, negR_b0, pt_tpc));
+    } 
+  }
 }
 
 //____________________________________________________________________________..
--- a/offline/packages/trackreco/PHSiliconTpcTrackMatching.cc
+++ b/offline/packages/trackreco/PHSiliconTpcTrackMatching.cc
@@ -81,11 +81,11 @@ int PHSiliconTpcTrackMatching::InitRun(PHCompositeNode *topNode)
     window_deta.set_use_legacy(_eta_search_win, this);
     window_dphi.set_use_legacy(_phi_search_win, this);
   } else {
-    window_dx.init_bools("dx", Verbosity()   >0);
-    window_dy.init_bools("dy", Verbosity()   >0);
-    window_dz.init_bools("dz", Verbosity()   >0);
-    window_dphi.init_bools("dphi", Verbosity() >0);
-    window_deta.init_bools("deta", Verbosity() >0);
+    window_dx.init_bools("dx", _print_windows || Verbosity()   >0);
+    window_dy.init_bools("dy", _print_windows || Verbosity()   >0);
+    window_dz.init_bools("dz", _print_windows || Verbosity()   >0);
+    window_dphi.init_bools("dphi", _print_windows || Verbosity() >0);
+    window_deta.init_bools("deta", _print_windows || Verbosity() >0);
   }
 
   return ret;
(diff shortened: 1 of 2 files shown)


david-stewart:
I am having trouble running this with with my Fun4All macro -- I will try and get my issues sorted today and run a few simple tests.

david-stewart:
For study of ranges and some documentation on how to set the bounds for the track matching, see [these slides](https://indico.bnl.gov/event/26224/attachments/59575/102352/2025_01_23_StewartTracking.pdf)

offline/packages/trackreco/PHSiliconTpcTrackMatching.h
```diff
@@ -40,6 +49,103 @@ class PHSiliconTpcTrackMatching : public SubsysReco, public PHParameterInterface
   float get_y_search_window() const { return _y_search_win; }
   float get_z_search_window() const { return _z_search_win; }
 
+  // 2024/01/22 update
+  struct WindowMatcher {
+    // --- option to use the legacy method ---
+    bool use_legacy = false;
+    double leg_search_win = 1.; // use if use_legacy == true; set in InitRun
+    PHSiliconTpcTrackMatching* parent_ptr {nullptr};
```
osbornjd, line 57: Why do we need this? Having a struct within a class having a pointer to said class gives me an uneasy feeling

offline/packages/trackreco/PHSiliconTpcTrackMatching.cc
```diff
@@ -87,6 +102,52 @@ void PHSiliconTpcTrackMatching::SetDefaultParameters()
...
+    negHi_b0 = (negHi[1]==0.);
+  }
+}
+
+bool PHSiliconTpcTrackMatching::WindowMatcher::in_window
+(const bool posQ, const double pt_tpc, double pt_si) 
+{
+  const auto delta = pt_tpc-pt_si;
+  if (use_legacy) {
+    return fabs(delta)*parent_ptr->getMatchingInflationFactor(pt_tpc) < leg_search_win;
```
osbornjd, line 130: I'm not sure I see how this will return what we have now. `pt_tpc` can be either the tpc phi, eta, pcax or pcay right? Or am I missing something? The inflation factor was pT dependent

osbornjd: In fact, we should have the `use_legacy` set to true by default for now so that we can confirm in the CI that nothing has changed

offline/packages/trackreco/PHSiliconTpcTrackMatching.cc
```diff
@@ -87,6 +102,52 @@ void PHSiliconTpcTrackMatching::SetDefaultParameters()
...
+  }
+  if (posQ) {
+    double pt = (pt_tpc<min_pt_posQ) ? min_pt_posQ : pt_tpc;
+    if (fabs_max_posQ) {
+      return fabs(delta) < fn_exp(posHi, posHi_b0, pt);
```
osbornjd, line 135: Just to confirm my understanding, these are exponential functions but you have set them to be effectively constant cuts with the appropriate parameter settings of (100,0,0) in the header?

osbornjd:
Also I missed this, how is this PR associated to https://github.com/sPHENIX-Collaboration/coresoftware/pull/3359? Does one need to be merged before the other, are they related in anyway?

david-stewart:
> Also I missed this, how is this PR associated to #3359? Does one need to be merged before the other, are they related in anyway?

I just asked to close that PR. After talking with Tony and the scope expanded to making new bounds, and seeing that the Q/pT bounds were matched well with an exponential function, and needing a way to put in lower and upper bounds (i.e. off center for the phi), then I superseded that old request with this one.

offline/packages/trackreco/PHSiliconTpcTrackMatching.cc
david-stewart: Not quite. The parameters are {a,b,c} as a+b*exp(c/pT), so that b=0 sets it to be a constant. The a=100 is a flag that basically says: "this parameter set is not used". There are four different sets needed (as a maximum).
1. The upper bound for |dX|< fn(a1,b1,c1)
2. The upper and lower limit for fn(a0,b0,c0)<dX<fn(a1,b1,c1)
and then two more if there are separate limits for positive and negative tracks (posQ and negQ)
In the PHSiliconTpcTrackMatching::InitRun(), the ::init_bools() is called on each WindowMatcher where the parameter sets with a==100 are set to be ignored or copied over.

david-stewart:
Good news:
This PR is ready to be merged. See [PDF with images](https://indico.bnl.gov/event/26201/attachments/59606/102400/PR_3373_verification.pdf)
Bad news:
There is some behavior with the residualtree that I don't understand. With both the existing and the updated windows with this code, changing the window size for one of the parameters (I used eta, but it is arbitrary) causes some shuffling of the matches inside of the new cuts.

offline/packages/trackreco/PHSiliconTpcTrackMatching.h
david-stewart: The old method used three member parameters and a member function to find
the envelop growth for each of the five function parameters. It was like this:
    
```
class PHSiliconTpcTrackMatching {

  double _match_function_a = 1.0;
  double _match_function_b = 5.0;
  double _match_function_pow = 1.0;
  // these are used to have a common window for magnification.
  . , , 
  double mag = getMatchingInflationFactor(tpc_pt); // uses _match_function_{a,b,pow};
  // the code then uses mag for all window, like:
  bool match_X = fabs(X_tpc-X_si) < mag*_X_search_win; // X is x,y,z,phi, or eta
}
```

The new method needs a custom function to be set for x,y,z,phi,eta (so up to five)
that can be off centered and, grow as a function of Q/pT (we found
   `a+b*exp(c/pT)` works well).

This motivates the generation of the member struct `WindowMatcher` to get the unique
parameters and to calculate the windows for x,y,z,phi,eta separately. The call points in the code then
become like:

`bool match_X = window_dX.in_window(track_charge, track_pt, X_tpc, X_si);` in which
each `window_dX` owns the parameters for the window growth.

To finally answer the question, if we wish to use the legacy mode, each
`window_dX` needs to either be able to (a) ask the parent class instance
`PHSiliconTpcTrackMatching` to do the legacy comparison, (b) get a copy of the
legacy function `_match_function_{a,b,power}` parameters and copy the legacy
method, or (c) we have to institute an `if (use_legacy)` switch for all five
parameter and call the legacy code outside of the `WindowMatcher` instances, or
(d) make a new function that directs traffic between the legacy mode and the
inidividual calls to `WindowMatcher`.

Option (a) is the simplest of these. To do that, then, the member `WindowMatcher` just 
need a pointer to calling instance of `PHSiliconTpcTrackMatching`, and to also 
remember if it is in legacy mode. Then, if in legacy mode, they
can just ask the parent isntance of `PHSiliconTpcTrackMatching` to use its method to
calculate the windows.

offline/packages/trackreco/PHSiliconTpcTrackMatching.cc
david-stewart: > I'm not sure I see how this will return what we have now. `pt_tpc` can be either the tpc phi, eta, pcax or pcay right? Or am I missing something? The inflation factor was pT dependent

To this first point, you are right -- I caught these when testing the code, and it's ok now.

> In fact, we should have the use_legacy set to true by default for now so that we can confirm in the CI that nothing has changed

To the second point, the PHSiliconTpcTrackMatching module will throw all the WindowMatcher `use_legacy` switches at once with the `PHSiliconTpcTrackMatching::set_use_legacy_windowing()` function (the only one the user should call). I have updated this value to be `true` by default.

osbornjd:
This still changes the default simulation QA, so there is still something that is inconsistent between the current build and this PR

david-stewart:
> This still changes the default simulation QA, so there is still something that is inconsistent between the current build and this PR

OK -- the efficiencies are really hit. This shouldn't be -- the plots I was QA'ing with shoes identical results, so I'll look. Hopefully a quick fix.

david-stewart:
Slow found it -- the PR, when using legacy mode, was magnifying the delta of the tpc and sil seed parameters instead of the window -- essentially it was shrinking instead of inflating the window as pT->0.

offline/packages/trackreco/PHSiliconTpcTrackMatching.h
osbornjd: Can you put the `getMatchingInflationFactor` function into a namespace so that it is accessible in the entire .cc file but nowhere else? e.g. at the top of PHSiliconTpctrackMatching
```
namespace 
{
getMatchingInflationFactor(float tpc_pt)
{
 .... function code ...
}
}
```
Then you should be able to call that function anywhere in that file (i.e. in PHSiliconTpcTrackMatching functions or WindowMatcher functions) - or am I missing something?

david-stewart: That would be ok, too. It would just require that all the fitting parameters are passed in each time -- or re-initialized in each call to `process_event`. Or, they could just be hard-coded in, an we never really use any other values than the default.
So, I did what I think is the simplest thing -- I just hard-coded the `mag=1+5./pT` (along with the min_pt value of `150 MeV) into the WindowMatchers to use in legacy mode, and remove the extra function and the pointers altogether.  I ran a small set to make sure that it gives the same results as the legacy code. The update is pushed.
When it gets merged (🤞), I'll post to MatterMost tracking thread to just add the `->set_use_legacy_windowing(false)` in order to update the windowing to current recommended values.

PR #3389, opened 2025-02-02
Minor bug fix for CentralityReco

hrjheng:

## Types of changes

Initialize m_mbd_total_charge to 0 instead of NaN to prevent the first event processed by Fun4All from having m_mbd_total_charge=nan and CentBin=0 (since adding a float to nan results in nan.)
This minor bug does not affect subsequent events, as ResetEvent sets m_mbd_total_charge = 0.

## TODOs (if applicable)

### DIFF
--- a/offline/packages/centrality/CentralityReco.cc
+++ b/offline/packages/centrality/CentralityReco.cc
@@ -194,6 +194,7 @@ int CentralityReco::FillVars()
       std::cout << scale_factor << "*" << m_centrality_scale << std::endl;
     }
 
+  m_mbd_total_charge = 0; // Reset the total charge
   for (int i = 0; i < 128; i++)
     {
       
--- a/offline/packages/centrality/CentralityReco.cc
+++ b/offline/packages/centrality/CentralityReco.cc
@@ -194,7 +194,6 @@ int CentralityReco::FillVars()
       std::cout << scale_factor << "*" << m_centrality_scale << std::endl;
     }
 
-  m_mbd_total_charge = 0; // Reset the total charge
   for (int i = 0; i < 128; i++)
     {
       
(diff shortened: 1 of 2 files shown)


pinkenburg:
Why does this need to be changed? 0 is an unlikely but valid value - we will not be able to distinguish a failure to set it from it being set to zero

hrjheng:
This needs to be changed because if m_mbd_total_charge is initialized to be NAN the very first event processed will always have m_mbd_total_charge=NAN which then gives a failed centrality value. Before each subsequent events is processed m_mbd_total_charge will be set to 0, so setting m_mbd_total_charge=0 for the first event shouldn't be an issue
```C
for (int i = 0; i < 128; i++)
    {
      
      m_mbd_hit = m_mbd_container->get_pmt(i);

      if ((m_mbd_hit->get_q()) < mbd_charge_cut)
	{
	  continue;
	}
      if (fabs(m_mbd_hit->get_time()) > mbd_time_cut) 
	{
	  continue;
	}
      m_mbd_total_charge += m_mbd_hit->get_q()*scale_factor*m_centrality_scale;
    }
```

hrjheng:
Made a slight adjustment—m_mbd_total_charge is now initialized as NaN but is set to 0 before the for loop where it gets aggregated
Also tagging @danjlis @pjwinnetou for review
```C
m_mbd_total_charge = 0; // Reset the total charge
  for (int i = 0; i < 128; i++)
    {
      
      m_mbd_hit = m_mbd_container->get_pmt(i);

      if ((m_mbd_hit->get_q()) < mbd_charge_cut)
	{
	  continue;
	}
      if (fabs(m_mbd_hit->get_time()) > mbd_time_cut) 
	{
	  continue;
	}
      m_mbd_total_charge += m_mbd_hit->get_q()*scale_factor*m_centrality_scale;
    }
```

pinkenburg:
Point taken - I think it is better to initialize this correctly in the class def., merging it now to get it into tomorrows build

PR #3404, opened 2025-02-05
Add Mvtx hit set helper

ycorrales:
- **ycm - clean and merge MVTX event header**
- **ycm - add new Mvtx trkr hitset helper to map hitsetkey with strobe id**
- **ycm - add default ctors/assigment**
- **ycm - changes to add new mvtx hit set helper**

 This PR add a map of hitsekey per strobe during the MVTX unpacking, this will allow to
run the clustering process for single strobe and Si seeding for a given range of strobes.

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/framework/ffarawobjects/MvtxFeeIdInfo.h
+++ b/offline/framework/ffarawobjects/MvtxFeeIdInfo.h
@@ -8,17 +8,28 @@
 class MvtxFeeIdInfo : public PHObject
 {
  public:
+  //! ctor
   MvtxFeeIdInfo() = default;
-  virtual ~MvtxFeeIdInfo() = default;
+
+  //! cp/mv ctor
+  MvtxFeeIdInfo(const MvtxFeeIdInfo&) = default;
+  MvtxFeeIdInfo(MvtxFeeIdInfo&&) = default;
+
+  //! cp/mv assignment
+  MvtxFeeIdInfo& operator=(const MvtxFeeIdInfo&) = default;
+  MvtxFeeIdInfo& operator=(MvtxFeeIdInfo&&) = default;
+
+  //! dtor
+  ~MvtxFeeIdInfo() override = default;
 
   virtual uint16_t get_feeId() const { return std::numeric_limits<uint16_t>::max(); }
-  virtual void set_feeId(const uint16_t) { return; }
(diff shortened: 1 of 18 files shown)


ycorrales:
thank you @pinkenburg, even benign I've taken care of that. @jdosbo how do you want to proceed did we remove the draft tag and merge the implementation.

pinkenburg:
there is a bd valgrind error (invalid read) in the hitsetunpacker. It's from
mvtx_event_header->add_strobe_BCO(feeIdInfo->get_bco());

ycorrales:
> there is a bd valgrind error (invalid read) in the hitsetunpacker. It's from mvtx_event_header->add_strobe_BCO(feeIdInfo->get_bco());

I think the problem is that jerkins raw input data was created with old MVTXRawEventheader version 1, we changed the raw event header to v2 few month ago in the event combiner but because it was not directly used, because Combiner rely on hit data only,  it was never an issue. Some of the change introduced uses data from the raweventheader which  has a different data member format and might originate the invalid read.

osbornjd (commented):
This looks good to me, I have a small request though given that we just got the mvtx hit unpacking into a condition where we think it is working the way we expect it to at the track seeding level. Can you show a plot of the hit distribution per strobe index before and after these changes just to show that nothing has changed? 

Ideally we would already have a check like this in jenkins so that it was done automatically but... we're working on it

ycorrales:
> This looks good to me, I have a small request though given that we just got the mvtx hit unpacking into a condition where we think it is working the way we expect it to at the track seeding level. Can you show a plot of the hit distribution per strobe index before and after these changes just to show that nothing has changed?
> 
> Ideally we would already have a check like this in jenkins so that it was done automatically but... we're working on it

I have compared the hit distribution using the old and new unpacker method. For a single event It looks ok.
Screenshot 2025-02-15 at 2 30 31 PM
with the difference related to the change of the strobe index definition.
However, merging other events like 100, I noticed a difference not only in the expected hit distribution but in the total number of hits.
Screenshot 2025-02-16 at 9 42 53 AM
The difference is because in the current unpacker there are duplicated hit for single strobe index windows, I can only interprete this case as some special timing cases where the current unpacker merge hits from consecutive strobe in a single strobe index and duplicated hits are removed.

PR #3489, opened 2025-03-23
fix: build mvtx as one layer in Acts geometry

osbornjd:
The current Acts geometry model (Gen1) is not flexible enough to handle the large shift in the MVTX barrel that we see in real life. Currently, the translations are implemented in the geometry context and the representative surfaces in Acts are moved outside of the approach surfaces. This leads to a phi asymmetry in the reconstructed track states as the Kalman Filter expects representative surfaces to be within the approach surfaces to perform filtering.

This PR modifies the Acts geometry layer building to build all MVTX layers within one "layer" such that there are only two approach surfaces for the entire MVTX - one entering layer 0, and one exiting layer 2. See [here](https://indico.bnl.gov/event/26197/) for more discussion.

For this to work with real data, we need to also update the geometry file and alignment file in the CDB. We will also need to update the material map that Acts uses in the KF, although that is not as urgent as the MVTX is the least amount of material in the detector and I suspect the material used in the KF is not currently the largest of our worries.

## Types of changes

## TODOs (if applicable)


osbornjd:
This is the new state phi distribution in the mvtx, which is substantially more uniform than previously was the case
[statephi.pdf](https://github.com/user-attachments/files/19411355/statephi.pdf)

osbornjd:
Merging this now so we can get a new ana build

PR #3506, opened 2025-04-01
(TrackerMillepedeAlignment) Added member, handles for writing TTree output

josephbertaux:
Added member and handles for writing TTree output with residual information
* Currently writes TTree with track local and global residual derivative information
* Will eventually add branches for other fields as they are desired

## Types of changes

## TODOs (if applicable)

### DIFF
--- a/offline/packages/TrackerMillepedeAlignment/MakeMilleFiles.cc
+++ b/offline/packages/TrackerMillepedeAlignment/MakeMilleFiles.cc
@@ -12,7 +12,6 @@
 #include <trackbase/TrkrDefs.h>
 
 #include <trackbase_historic/ActsTransformations.h>
-#include <trackbase_historic/SvtxAlignmentState.h>
 #include <trackbase_historic/SvtxTrack.h>
 #include <trackbase_historic/SvtxTrackMap.h>
 #include <trackbase_historic/SvtxTrackState.h>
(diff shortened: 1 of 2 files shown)


osbornjd (commented):
Maybe I'm misunderstanding - what is the purpose of this? There is already a tree that gets dumped out in helical fitter with a ton of information (including what is added here). Is the purpose to use this with the Acts alignment states?

josephbertaux:
Yes--the idea is to have TTree/TNtuple output with the binary Millepede output also, so we can make similar residual plots when doing alignment iterations. The advantage provided here is that we'd be able to do this with any track states on the node tree, instead of just HelicalFitter's.

I've tested this by running it with HelicalFitter, the output is 1:1 for the branches I've added (though I've given the branches different names here)

josephbertaux:
(For now I've only added the derivatives as an initial sanity track, I'll have to add more branches as we go--I haven't added branches for the track/vertex residuals nor the global r, phi coordinates yet)

josephbertaux:
I just realized the member `m_tree` may be null but the line m_tree->Fill() is not protected by a guard clause, oops (converted to draft)

josephbertaux:
My code had already diverged a bit since I noticed that--I'll close this PR and reopen with a more substantial refactor after QM

PR #3548, opened 2025-04-30
yc - fix reading MVTX RO mode. using isnan function for nan comparison

ycorrales:
This PR will fix the method to read MVTX RO modes.
The current approach is to read the MVTX RO method from OCDB first and in case the
value is not found try with DAQ DB. the latest is never reached becasue the outcome of the
comparison was always true.

## Types of changes

MVTX Offline recosntruction for Run25 must be redone after this PR implementation.

## TODOs (if applicable)

### DIFF
--- a/offline/framework/fun4allraw/MvtxRawDefs.cc
+++ b/offline/framework/fun4allraw/MvtxRawDefs.cc
@@ -1,6 +1,6 @@
 #include "MvtxRawDefs.h"
 
-#include <odbc++/connection.h> // odbc::Connection
+#include <odbc++/connection.h>  // odbc::Connection
 #pragma GCC diagnostic push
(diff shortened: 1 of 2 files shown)


osbornjd:
Does this affect the pp production also or just the recent cosmics where the strobe time wasn't being recorded properly?

ycorrales:
> Does this affect the pp production also or just the recent cosmics where the strobe time wasn't being recorded properly?

@osbornjd this affects all runs (independently whether strobe information was recorded or not) taken after @pinkenburg imaged the DAQ DB info to fix the wrongly recorded information for cosmic runs last period. I think this was done during RHIC shutdown  so pp would be ok. Only Run25 runs are be affected. Please notice cosmic before the fix are still affected because their strobe information are not available neither in the OCDB nor DAQ DB.

PR #3583, opened 2025-05-14
New overload of encode_epd to use sEPD mapping

largewedge:
[comment]: <> 
The existing version of encode_epd(arm,ring,phi) generates keys using a mapping that was defined for simulation but does not match the physical configuration of the detector. To enable generating a key from the (arm,rbin,phibin) coordinates of a tile in data, a new version of the function is introduced, which uses an array that was generated from the CDB entry with url SEPD_CHANNELMAP. An additional flag (isData) is passed to encode_epd to distinguish this overload from the one used in simulation.

## Types of changes

## TODOs (if applicable)


pinkenburg:
I assume this flag is not under user control but set somewhere in the code according to what is being processed. Rather than overloading a method we should either use a default argument or force its use to make sure there are no accidental bad default uses. What is the plan for embedding? This sounds like sims are actually stored differently from real data in the TowerContainer, how can simulated channels be merged with real data if not in the TowerContainer?

largewedge:
> I assume this flag is not under user control but set somewhere in the code according to what is being processed. Rather than overloading a method we should either use a default argument or force its use to make sure there are no accidental bad default uses. What is the plan for embedding? This sounds like sims are actually stored differently from real data in the TowerContainer, how can simulated channels be merged with real data if not in the TowerContainer?


It seems like it would be fine to force the use of isData=true, but maybe it would be best to simply define a separate function for use in simulation so that accidental use of the wrong mapping is less likely. I wouldn't expect simulated towers to be stored differently than real ones--it's just that mapping from (arm,rbin,phibin)->key is different.

largewedge:
I've tested this in simulation. We pass geantinos through each tile (one per event). For each g4hit, we find the corresponding (arm,rbin,phibin) and then call encode_epd(arm,rbin,phibin). We then check the simulated tower using this key, and, if the mapping is correct, then we should find a tower with energy = -1 GeV. The module code is [here](https://github.com/sPHENIX-Collaboration/analysis/blob/master/sEPDMappingScanner/src/CheckEpdMap.cc#L569C1-L569C70). 

The result of this test using the most recent version of encode_epd(arm,rbin,phibin) (no overload) shows that the mapping is consistent in simulation (output attached--matches between g4hits and reconstructed tiles are marked green, and mismatched tiles are marked in red). This is because of how PHG4EPDModuleReco uses encode_epd to fill the simulated TowerInfoContainer (https://github.com/sPHENIX-Collaboration/coresoftware/blob/master/simulation/g4simulation/g4epd/PHG4EPDModuleReco.cc#L161 ). If we attempt to use the isData=True overload without adjusting PHG4EPDModuleReco, then every tile is mapped incorrectly as expected.


sEPD_sim_new_map

largewedge:
Note that I'm still doing some testing on this latest version and may make some changes. I hadn't realized that the changes to my fork would show up in this PR

pinkenburg:
We should stop looping over 768 channels when we have only 744 populated. This just creates confusing code and we should just stop doing it this way. Given that the mapping is final - we should just have an array[744] with array[channel] = TCindex (or key);
Why do we have a huge 3d array now in the TowerInfoDefs.cc instead of the old 2d array? I haven't done the math but I assume this will break the readback of existing simulated DSTs which is something we have to avoid.

largewedge:
> We should stop looping over 768 channels when we have only 744 populated. This just creates confusing code and we should just stop doing it this way. Given that the mapping is final - we should just have an array[744] with array[channel] = TCindex (or key); Why do we have a huge 3d array now in the TowerInfoDefs.cc instead of the old 2d array? I haven't done the math but I assume this will break the readback of existing simulated DSTs which is something we have to avoid.


The large 3d arrays in TowerInfoDefs are leftover from the earlier commits and need to be reverted. I will change the bounds on the channel mapping loop to reflect the actual number of channels in the detector.

largewedge:
> I think what we concluded at the software meeting was to apply the mapping in calotowerbuilder (already addressed by PR [#3609](https://github.com/sPHENIX-Collaboration/coresoftware/pull/3609).

I am implementing the mapping in the way Chris has suggested, which was the conclusion of the software meeting.

PR #3609, opened 2025-05-24
Adjust sEPD towerinfo container mapping 

eumaka:
Can be marked as draft until discussion on Tuesday.

PR makes changes to the processed waveform order for the sEPD to map into towerinfocontainer just as installed in the detector and so as use seamlessly with TowerinfoDefs (w/o adjustments)...

TowerinfoDefs can be used directly to get the correct map (no need for CDB lookup)

E.g.
for ( int i = 0; i < 744; ++i )
 {
     unsigned int key = TowerInfoDefs::encode_epd(i);
     int arm = TowerInfoDefs::get_epd_arm(key);
     int phibin = TowerInfoDefs::get_epd_phibin(key);
}
etc...

As a check, produced plots with this setup are attached below (they are in agreement w/ previous setup and online monitoring (north phi bin is flipped)):
sepd_wheel_adc_map

### DIFF
--- a/offline/packages/CaloReco/CaloTowerBuilder.cc
+++ b/offline/packages/CaloReco/CaloTowerBuilder.cc
@@ -95,7 +95,7 @@ int CaloTowerBuilder::InitRun(PHCompositeNode *topNode)
 
     m_fieldname = "adcskipmask";
 
-    std::string calibdir = CDBInterface::instance()->getUrl(m_calibName);
+    calibdir = CDBInterface::instance()->getUrl(m_calibName);
     if (calibdir.empty())
     {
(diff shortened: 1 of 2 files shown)


bseidlit:
We have worked vary hard to make calotowerbuilder streamlined and detector agnostic to preserve clarity and maintainability.  
Why does the additional mapping need to be incorporated into calotowerbuilder, shouldn't this be delt with in the towerinfo container, like the rest of the calorimeters?

pinkenburg:
My hope is that this remapping will not be needed anymore once the channel mapping is applied during filling of the TowerInfoContainer from the raw data

PR #3620, opened 2025-05-28
Transient alignment transformation distortion corrections

adfrawley:

## Types of changes

Implementation of TPC distortion corrections for Acts fitting using transient alignment transforms. The surface is moved to the distortion corrected cluster position, instead of moving clusters back to the nominal surface position. It works when using direct navigation in the Acts fitter. It does not work when using the default navigation - that is a separate issue that still needs to be understood, and so using the cluster mover is still the default setting for the Acts fitter.

Because the transient alignment transformations are generated independently for each track, and are not kept, SvtxTrackState_v3 is added to make both the local and global track state positions available downstream.

For direct navigation, it give results very similar to those produced by the cluster mover method (identical results are not expected). The transient transform method results in 4-5% more tracks.

## TODOs (if applicable)


osbornjd (commented):
Should we just store the local coordinates and remove the global coordinates like we do with all other objects (I.e. expecting then that one can get the global coordinates from the appropriate transform)

PR #3624, opened 2025-05-29
Track-EMC clus matching added to KFP.

viewolfe321:
I have added track to EMCal cluster matching functionality in KFParticle. I don't anticipate this breaking anything or requiring any other code to need to change in response.

## Types of changes

This pull request will add a feature. My code updates KFParticle_truthAndDetTools.cc and KFParticle_truthAndDetTools.h to add track-calo matching. It also adds one new branch in the TTree KFP produces. 

## TODOs (if applicable)

### DIFF
--- a/offline/packages/KFParticle_sPHENIX/KFParticle_truthAndDetTools.cc
+++ b/offline/packages/KFParticle_sPHENIX/KFParticle_truthAndDetTools.cc
@@ -1,5 +1,4 @@
 #include "KFParticle_truthAndDetTools.h"
-
 #include "KFParticle_Tools.h"  // for KFParticle_Tools
 
 #include <g4eval/SvtxEvalStack.h>   // for SvtxEvalStack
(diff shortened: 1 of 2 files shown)


hupereir:
> I don't think I understand the Jenkins cpp-check complaint as the same code is used elsewhere. Maybe this object is guaranteed to have been checked earlier in the code and that's why it thinks its redundant? Then maybe these lines before `track = getTrack(daughter.Id(), dst_trackmap);` could be removed?
> 
> ```c++
> dst_trackmap = findNode::getClass<SvtxTrackMap>(topNode, m_trk_map_node_name_nTuple);
>   if (!dst_trackmap)
>   {
>     std::cout << "KFParticle truth matching: " << m_trk_map_node_name_nTuple << " does not exist" << std::endl;
>   }
> 
>   track = getTrack(daughter.Id(), dst_trackmap);
> ```

I think the point is that if you feel the need to check for non-null value for dst_trackmap, then you should not call getTrack( ..., ) after that, in case the pointer is null 
So: returning at the end of the if(!dst_trackmap) block, should then quiet cpp check. Makes sense ? 
(or putting the rest of the code in an 'else' block
On the other hand, non_null dst_trackmap is tested inside the ::getTrack method, then the earlier call is redundant and can be removed.

PR #3630, opened 2025-05-31
update GlobalQA 

eumaka:
Short update to GlobalQA for sEPD after merged mapping implementation in #3609


bseidlit:
Can we remove or compress h_GlobalQA_sEPD_tile into a th2.  The 744 histograms clutter the root file and slow down hadding, which we use automated scripts for and rely on timely hadding.

eumaka:
actually don't need those hists anymore. I used them early in the run to get calibration factors. Can't be removed now

PR #3632, opened 2025-06-01
add sEPD psi_2 to jet background 

eumaka:
PR adds sEPD combined arms second order event plane angle to jet background subtraction.


eumaka:
Empty commit to trigger jenkins - eventplaneinfo was built later in packages.txt, hence couldn't find library.

PR #3636, opened 2025-06-03
use new CDB default setup

eumaka:
PR makes use of CDB default setup for default calibrations in event plane calibration

### DIFF
--- a/offline/packages/eventplaneinfo/EventPlaneReco.cc
+++ b/offline/packages/eventplaneinfo/EventPlaneReco.cc
@@ -93,25 +93,22 @@ EventPlaneReco::EventPlaneReco(const std::string &name) : SubsysReco(name) {
 }
 
 int EventPlaneReco::InitRun(PHCompositeNode *topNode) {
-    
+
+    FileName = "EVENTPLANE_CORRECTION_default";
     if(_isSim)
     {
-        m_runNo = 0;
+        FileName = "EVENTPLANE_CORRECTION_SIM_default";
     }
-    if(!_default_calib)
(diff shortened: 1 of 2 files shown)


bseidlit:
Also, in the Calos we add sim calibration to the MDC2 tag which have the payload name as the data. So then we don't need the sim flag.  Would this be possible to implement?

eumaka:
Yep, what you said is what I want.
Find calibration for this run, if not found use default calibration. 

I'm not 100% sure about the new CDB setup. @pinkenburg did you say that the default setting can be switched off on the CDB side and then it will search for a run by run calibration? 

For sim, I don't think it matters what name we pick, if we keep the appended "SIM" or drop it and use the MDC2 tag. As long as the GEANT simulation conditions and the HIJING/FlowAfterBurner settings don't change, the same sim calibration file can be used.

bseidlit:
"For sim, I don't think it matters what name we pick, if we keep the appended "SIM" or drop it and use the MDC2 tag. As long as the GEANT simulation conditions and the HIJING/FlowAfterBurner settings don't change, the same sim calibration file can be used."
- it more, I would like to get rid of the sim flag so we can use identical macros in data and sim

eumaka:
Ok I'll keep that in mind for future PR/commits. I don't want to make multiple commits to an open PR.

eumaka:
Ok. The CDB functionality we want is a search for run by run calibration. If calibration file for a particular run is not available in the CDB, then use the default calibration file.

eumaka:
awesome, thanks!

PR #3638, opened 2025-06-04
Ignore BCO skip in INTT standalone mode

gwd213:

## Types of changes

Adding feature to turn ON/OFF big BCO skip based on GL1 BCO value. This is necessary for INTT local run without GL1 combiner. It will not harm any current combiner since standalone flag is false by default ; bool m_StandaloneMode{false}

## TODOs (if applicable)

### DIFF
--- a/offline/framework/fun4allraw/Fun4AllStreamingInputManager.cc
+++ b/offline/framework/fun4allraw/Fun4AllStreamingInputManager.cc
@@ -1291,14 +1291,13 @@ int Fun4AllStreamingInputManager::FillInttPool()
   }
   for (auto iter : m_InttInputVector)
   {
+	if(!m_gl1_registered_flag)
+	  iter->SetStandaloneMode(true);
 	if (Verbosity() > 0)
 	{
 	  std::cout << "Fun4AllStreamingInputManager::FillInttPool - fill pool for " << iter->Name() << std::endl;
 	}
-	if(!m_gl1_registered_flag)
-	{
-	}
-	iter->FillPool(ref_bco_minus_range,m_gl1_registered_flag);
+	iter->FillPool(ref_bco_minus_range);
 	// iter->FillPool();
 	if (m_RunNumber == 0)
 	{
--- a/offline/framework/fun4allraw/Fun4AllStreamingInputManager.cc
+++ b/offline/framework/fun4allraw/Fun4AllStreamingInputManager.cc
@@ -1291,30 +1291,33 @@ int Fun4AllStreamingInputManager::FillInttPool()
   }
   for (auto iter : m_InttInputVector)
   {
-    if (Verbosity() > 0)
-    {
-      std::cout << "Fun4AllStreamingInputManager::FillInttPool - fill pool for " << iter->Name() << std::endl;
-    }
-    iter->FillPool(ref_bco_minus_range);
-    // iter->FillPool();
-    if (m_RunNumber == 0)
-    {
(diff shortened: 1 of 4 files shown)


osbornjd (changes requested):
Instead of having another flag that needs to be set at the macro level, could we just set the `m_StandaloneMode==true` when there is no GL1 node present? i.e. in the streaming input manager, if there is no GL1 registered, then any pooler registered is assumed to be running in standalone mode (there is a flag already for the GL1 [here](https://github.com/sPHENIX-Collaboration/coresoftware/blob/83a160c67def008b0b63245c1ee4eaea46740557/offline/framework/fun4allraw/Fun4AllStreamingInputManager.cc#L151))

gwd213:
> Instead of having another flag that needs to be set at the macro level, could we just set the `m_StandaloneMode==true` when there is no GL1 node present? i.e. in the streaming input manager, if there is no GL1 registered, then any pooler registered is assumed to be running in standalone mode (there is a flag already for the GL1 [here](https://github.com/sPHENIX-Collaboration/coresoftware/blob/83a160c67def008b0b63245c1ee4eaea46740557/offline/framework/fun4allraw/Fun4AllStreamingInputManager.cc#L151))

I just updated the PR and enabled Standalone mode based on the m_gl1_registered_flag.
This change no longer requires modifications at the macro level.
Could you please confirm whether this update might affect other macros or potentially cause issues in production?

osbornjd (commented):
Instead of passing it as an argument to the FillPool function, I was thinking you could set the flag you originally had as `m_gl1_registered_flag`. For example
```
if(!m_gl1_registered_flag)
{
inttpool->set_standalone(true);
}
```
or something like that - that way the SingleStreamingInput base class didn't need any changes

In any case I don't think the current change will achieve what you want, since at line 1301 the `FillPool` call is outside of the `if` statement

gwd213:
> Instead of passing it as an argument to the FillPool function, I was thinking you could set the flag you originally had as `m_gl1_registered_flag`. For example
> 
> ```
> if(!m_gl1_registered_flag)
> {
> inttpool->set_standalone(true);
> }
> ```
> 
> or something like that - that way the SingleStreamingInput base class didn't need any changes
> 
> In any case I don't think the current change will achieve what you want, since at line 1301 the `FillPool` call is outside of the `if` statement

Thank you for your suggestion, but I am wondering how I can use ```m_gl1_registered_flag``` in SingleInttPoolInput.cc/h since ```Fun4AllStreamingInputManager.cc/h``` uses only base class ```SingleStreamingInput.cc/h```.

I might misunderstand your point or way to implement this feature

osbornjd:
I am saying from the Fun4AllStreamingInputManager you could check if `m_gl1_registered_flag==false` then set the Intt pool input object to run in standalone mode. It may be good to implement this at the base class level `SingleStreamingInput` as other pooling objects could be treated in the same way for standalone running

gwd213:
> I am saying from the Fun4AllStreamingInputManager you could check if `m_gl1_registered_flag==false` then set the Intt pool input object to run in standalone mode. It may be good to implement this at the base class level `SingleStreamingInput` as other pooling objects could be treated in the same way for standalone running

Thanks Joe! I think I got your point :)
I've modified InputManager and put functions at SingleStreamingInput just in case other sub-system might need this feature in future.

osbornjd (approved):
Thanks for making the changes! Looks good to me

PR #3663, opened 2025-06-10
PHActsTrkFitter now uses direct navigation by default.

adfrawley:

## Types of changes

This change makes direct navigation the default for the Acts track fitter. Changing back to default navigation can be done from the macro. 

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackreco/PHActsTrkFitter.h
+++ b/offline/packages/trackreco/PHActsTrkFitter.h
@@ -216,7 +216,7 @@ class PHActsTrkFitter : public SubsysReco
   /// Flag for pp running
   bool m_pp_mode = false;
 
-  bool m_directNavigation = false;
+  bool m_directNavigation = true;
   
   // do we have a constant field
   bool m_ConstField{false};


osbornjd:
Some observations from the QA
1. The low pT efficiency drops a little. Not a problem, but would be interesting to have an answer as to why that is... It is backwards from what my naive expectation would be
2. The track fits with 4 MVTX clusters instead of 3 drop. This is an indication of how Acts deals with multiple clusters on a single layer - apparently the default navigator can deal with it better than the direct navigator. This might be correlated to the low pT efficiency, where we are probably more likely to get 2 staves in one layer on a single track.
3.  The DCA/sigma(DCA) becomes worse at low pT (i.e. width farther from unity). Unclear why that is

adfrawley:
I agree that the drop in efficiency for tracks below 200 MeV is puzzling. 
For these results we should keep in mind that the simulated data are kinder to the default navigator than the real data.

I think it is reasonably consistent with what we have seen in real data. Looking at the Pythia QA:
  The track efficiency at large η is a little better, similar to what we see in real data.
  The DCA values are slightly worse, similar to what we see in real data.

I have never understood the sigmalized DCA plots in the QA. They seem to be inconsistent with the DCA plots.

PR #3685, opened 2025-06-18
Add code to: check runnumber and update m_species flag in CaloValid, set verbosity for CaloValid.

nk7252:
Add code to: check runnumber and update m_species flag in CaloValid, set verbosity for CaloValid.

## Types of changes

Add code that gets the runnumber during CaloValid::InitRun.
The runnumber is compared to run ranges defined in RunnumberRange.h 
This sets the m_species (pp, AuAu) flag so the correct charge scaledowns are applied for correlation QA plots.

Additionally added verbosity method to CaloValid.h
Currently the only couts it applies to are ones that announce the species, lack of a RunHeader node, or if the runnumber is outside of the range defined in RunnumberRange.h.
These are shown if verbosity>0

### DIFF
--- a/offline/QA/Calorimeters/CaloValid.cc
+++ b/offline/QA/Calorimeters/CaloValid.cc
@@ -17,15 +17,16 @@
 #include <qautils/QAHistManagerDef.h>
 
 #include <ffaobjects/EventHeader.h>
+#include <ffaobjects/RunHeader.h>  // for runnumber
 
 #include <ffarawobjects/Gl1Packet.h>
 
 #include <fun4all/Fun4AllHistoManager.h>
 #include <fun4all/Fun4AllReturnCodes.h>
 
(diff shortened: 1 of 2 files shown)


osbornjd:
You can always use clang-format to format the code in the sPHENIX style, as in
```
clang-format -i MyFile.cc MyOtherFile.h 
```

nk7252:
I did just that. Thanks Joe!

pinkenburg:
> clang-format -i MyFile.cc MyOtherFile.h

actually the command is
clang-format -style=file -i MyFile.cc MyOtherFile.h
this will apply our customizations, without the -style it uses the default

pinkenburg:
I'll merge it but we should use uint64_t for scalers, not long long which will go negative when the upper bit is set. Chances of this are very low (non existent) but for 32 bit counters this is a perennial issue

PR #3704, opened 2025-06-26
CD: Silicon seeding update

cdean-github:

## Types of changes

@hrjheng and I altered the silicon seeding method, instead of searching for a matching cluster that is within a specified separation from the projected seed along the INTT face, we now figure out the radial separation between the clusters on the initial layer and the projected layer, and the separation between the cluster on the projected layer and the seed location on said layer. This allows us to calcualte the delta phi between the seed projection and the potential cluster. This avoids search biases coming from the offset of the MVTX. See attached photo

(Note for photo, in the middle drawing I have one projection that has an rφ1 but has its arc defined as Δφ2 and this is reversed for the other projection, this makes the drawing a little confusing but I didnt notice till after I took the picture. The inequality at the bottom is correct)

@adfrawley @jdosbo you could try this on full tracks.

rn_image_picker_lib_temp_210d97ea-4c38-4a00-b33e-9d5897078c86

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trackreco/PHActsSiliconSeeding.cc
+++ b/offline/packages/trackreco/PHActsSiliconSeeding.cc
@@ -45,6 +45,8 @@
 #include <Acts/Seeding/Seed.hpp>
 #include <Acts/Seeding/SeedFilter.hpp>
 
+#include <algorithm>
+
 namespace
 {
   template <class T>
@@ -318,15 +320,13 @@ void PHActsSiliconSeeding::makeSvtxTracks(const GridSeeds& seedVector)
         cluster_keys.push_back(cluskey);
 
         trackSeed->insert_cluster_key(cluskey);
-        auto globalPosition = m_tGeometry->getGlobalPosition(
-            cluskey,
-            m_clusterMap->findCluster(cluskey));
+        auto globalPosition = m_tGeometry->getGlobalPosition(cluskey, m_clusterMap->findCluster(cluskey));
         globalPositions.push_back(globalPosition);
         if (m_seedAnalysis)
         {
           m_mvtxgx.push_back(globalPosition(0));
           m_mvtxgy.push_back(globalPosition(1));
-	  m_mvtxgz.push_back(globalPosition(2));
+	        m_mvtxgz.push_back(globalPosition(2));
         }
         positions.insert(std::make_pair(cluskey, globalPosition));
         if (Verbosity() > 1)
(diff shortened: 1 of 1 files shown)


offline/packages/trackreco/PHActsSiliconSeeding.cc
```diff
@@ -726,11 +767,28 @@ std::vector<TrkrDefs::cluskey> PHActsSiliconSeeding::findMatches(
...
+          //           << " r_prev " << r_prev
+          //           << std::endl;
+          //if ((det == TrkrDefs::TrkrId::mvtxId && rphiresid < m_mvtxrPhiSearchWin &&
+          //     zresid < m_mvtxzSearchWin) ||
+          //    (det == TrkrDefs::TrkrId::inttId && rphiresid < m_inttrPhiSearchWin && zresid < m_inttzSearchWin))
+          //Lets try a phi search instead of rphi
           if ((det == TrkrDefs::TrkrId::mvtxId && rphiresid < m_mvtxrPhiSearchWin &&
                zresid < m_mvtxzSearchWin) ||
-              (det == TrkrDefs::TrkrId::inttId && rphiresid < m_inttrPhiSearchWin && zresid < m_inttzSearchWin))
+              (det == TrkrDefs::TrkrId::inttId && segment_phi < 0.03 && zresid < m_inttzSearchWin)) //0.112 is about 4mm, 0.06 is about 1mm
```
osbornjd, line 787: Instead of hard coding this value, should we just make it configurable? This is probably not the last time we will want to explore this phase space...

cdean-github: I think Hao Ren was still working on the macro level hooks. This was pushed so others could check the tracking output

osbornjd: Interestingly the QA shows it makes the INTT matching worse by a small amount - I'm not sure why that would be with ideal geometry...

hrjheng: I haven't really started changing macro though. Do we want to implement class functions for the threshold at this point yet? The latest test I did was to remove the global shifts to clusters and I didn't see much improvement 
If we think this updated algorithm is the way to go, I can start working on it tomorrow

adfrawley: We need to evaluate the performance before we commit to this. I did not see much effect on the seeding efficiency in pp data.

cdean-github: Agreed that we need to evaluate this. Im leaving it as a draft so we can check it out. Joe, is there a way to disable Jenkins for a specific PR? or should I cancel the PR and people can pull from my local repo?

osbornjd: No, Jenkins just runs automatically. If you expect to keep pushing to the branch but want to hold off I would just close the pr and open a new one. I do share Tony's confusion though, I thought what we originally saw was that shifting the mvtx back to approximately concentric around the origin led to better seeding performance. That was a separate development from the intt projection changes, I thought

PR #3723, opened 2025-07-06
tag events in  jetbackground where flow was not applied in UE subtraction

eumaka:
PR adds a flag to jet background to tag events where flow is not determined because there are no calorimeter strips for v2 determination or when there is no provide psi_2 during the flow modulated jet background subtraction.

Can be used downstream as:

TowerBackground *towerbackground;
towerbackground = findNode::getClass<TowerBackground>(topNode, "TowerInfoBackground_Sub2");
if(towerbackground->get_flow_failure_flag())
{
  these events have no strips for flow, or no provided psi_2
}


pinkenburg:
This PR adds a data member to a versioned class which wreaks havoc if this class was saved on some DST. Are we sure this is not the case?

PR #3772, opened 2025-07-23
add ring by ring event plane determination for sEPD

eumaka:
PR adds the Q vectors  and psi_n calculations for sEPD rings. Correction methodology for new additions to follow...


pinkenburg:
We need to address how you deal with schema evolution. If a data member is added you have to at least bump up the version number of the class in the ClassDef, this will be able to read old versions without issuing a warning. If data members are removed you need a new version

eumaka:
But this class is not stored in a DST. Data members were not removed.

pinkenburg:
> But this class is not stored in a DST. Data members were not removed.

then we are good

PR #3799, opened 2025-08-03
KFP updates: DST, vertex, track-calo matching

yuxdPKU:

## Types of changes

Several changes on KFParticle 
1. In the previous KFParticle, if multiple mother/intermediate particle is reconstructed in an event, trackmap need to be inserted multiple times. Due to duplicate keys, there are lots of warnings. In the current update, I add a check before ``insertWithKey`` to avoid warnings.
2. In track-calo matching study, track states at calo radius will be created, which do not have trkr cluster associated with. Therefore, if we call ``getGlobalPosition`` function, there will have many warnings, like surface not found. In the current update, I check if cluster is valid or not. If not, just continue to avoid warnings.
3. Add ``requireTrackEMCalMatch `` function to require daughter tracks must be matched with EMCal, which helps to minize output nTuples for quarkonium analysis. Plus some cosmetic modification on code. If we disable either ``getCaloInfo``, or ``requireTrackEMCalMatch``, the basic functionality is not affected, and they are set to false by default, so track-only analysis will not be influenced.
4. Update Vertex-related code in the ``KFParticle_Tools``. In the previous code, If enabling GlobalVertexMap, code will crash, because in some functions, we still search for GlobalVertex id in SvtxTrackMap which is not self-consistent. I basically update ``KFParticle_Tools::makeAllPrimaryVertices``, ``KFParticle_Tools ::getTracksFromBC``, ``KFParticle_Tools::getTracksFromVertex``, ``KFParticle_Tools::checkTrackAndVertexMatch`` functions. I also modify ``KFParticle_nTuple`` a bit, make more clear about
4.1 number of vertices for a given bunch crossing -- ``nPrimaryVerticesOfBC``
4.2 number of tracks for a given bunch crossing -- ``nTracksOfBC``
4.3 number of tracks associated with a given vertex -- ``nTracksOfVertex``

@cdean-github Please review it if they are reasonable. Thanks!

## TODOs (if applicable)

### DIFF
--- a/offline/packages/KFParticle_sPHENIX/KFParticle_DST.cc
+++ b/offline/packages/KFParticle_sPHENIX/KFParticle_DST.cc
@@ -177,7 +177,10 @@ void KFParticle_DST::fillParticleNode_Track(PHCompositeNode* topNode, KFParticle
   SvtxTrack *dummyMother = nullptr;  
   while (!dummyMother)
   {
-    dummyMother = m_recoTrackMap->insertWithKey(m_recoTrack, resonanceCounter);
+    if (!m_recoTrackMap->get(resonanceCounter))
+    {
+      dummyMother = m_recoTrackMap->insertWithKey(m_recoTrack, resonanceCounter);
+    }
     --resonanceCounter;
   }
   m_recoTrack->Reset();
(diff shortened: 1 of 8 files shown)


cdean-github (commented):
Hi Xudong, 

In general I think this looks good. I have some small comments below that we can discuss before we merge this. The vertex stuff should certainly safe guard against seg faults.

Cheers,
Cameron

Line 404, offline/packages/KFParticle_sPHENIX/KFParticle_Tools.cc

What happens here if we require a MBD vertex (if this is a bug, I think it's a carry over from my initial coding and I only noticed it now)

Line 675, offline/packages/KFParticle_sPHENIX/KFParticle_nTuple.cc

This feels like it will create confusion if someone isn't doing track/calo matching but I don't have a good suggestion at this time

We should update https://wiki.sphenix.bnl.gov/index.php/KFParticle with the new options

Line 702, offline/packages/KFParticle_sPHENIX/KFParticle_truthAndDetTools.cc

Valerie and Shelby should be informed of this change

viewolfe321:
Hi Xudong. The changes look good to me. I hadn't noticed the reversal of the dEta (cal - track vs track - cal), so I'm glad you caught and modified that. Thank you for adding the function to cut out the non calo-matched tracks!

PR #3828, opened 2025-08-15
Improvements to sEPD software

largewedge:
This PR adds support for applying the sEPD calibration in CaloTowerCalib and for adding the sEPD geometry to the node tree using the RawTowerGeomContainer, in keeping with the conventions followed by calorimeter software.

## Types of changes

As discussed at the BulkTG meeting: https://indico.bnl.gov/event/29269/
Note that this requires the calibrations file to use the tower key to index channels.

## TODOs (if applicable)
Consider moving sEPDGeomMapping functionality into CaloGeomMapping.
Change sEPD calibration files to store (1/MPV) rather than MPV so that no special cases are needed in CaloTowerCalib.

Corresponding changes in production macros at this PR: https://github.com/sPHENIX-Collaboration/macros/pull/1176

### DIFF
--- a/offline/packages/CaloReco/CaloTowerCalib.cc
+++ b/offline/packages/CaloReco/CaloTowerCalib.cc
@@ -237,6 +237,13 @@ int CaloTowerCalib::process_event(PHCompositeNode *topNode)
     float calibconst = cdbttree->GetFloatValue(key, m_fieldname);
     bool isZS = caloinfo_raw->get_isZS();
 
+    if (calibconst <= 0)
+    {
+      _calib_towers->get_tower_at_channel(channel)->set_energy(0.0);
+      _calib_towers->get_tower_at_channel(channel)->set_isNoCalib(true);
+      continue;
+    }
+
     if (isZS && m_doZScrosscalib)
     {
       float crosscalibconst = cdbttree_ZScrosscalib->GetFloatValue(key, m_fieldname_ZScrosscalib);
@@ -249,11 +256,6 @@ int CaloTowerCalib::process_event(PHCompositeNode *topNode)
     else
     {
       _calib_towers->get_tower_at_channel(channel)->set_energy(raw_amplitude * calibconst);
-    }
-   
-    if (calibconst == 0)
-    {
-      _calib_towers->get_tower_at_channel(channel)->set_isNoCalib(true);
     }
     if(m_dotimecalib)
     {


bseidlit:
The changes to calotowercalib seem largely unnecessary and undesirable.   
I am fine with changing isNoCalib to be applied to all calibrations <= 0 for all subsystems. 

But can you just use the convention for the CDB tree  name already in place? Putting in this code  just because this is the BDC tree already in the CDB is just laziness and add to code bloat.

I fully support integrating the EPD into the main calo software but the whole point is to not use a bunch of flags for EPD when avoidable. 

if I am looking at this right you should only need to change a line or two of calotowercalib.

largewedge:
@bseidlit Agreed. I have reverted those changes. I will fix the calibration file to use the expected convention

eumaka:
I have some significant technical concerns about the proposed changes to retire EpdReco and replace it with sEPDGeomMapping:

Architectural Issues: The current approach introduces unnecessary complexity by duplicating geometry calculations. The workflow now becomes: copy phi, r, z values from EpdReco into sEPDGeomMapping with hardcoded phi values → convert to Cartesian coordinates → store in RawTowerGeomv1 → recalculate phi in downstream EventPlaneReco code. This circular conversion adds computational overhead and potential precision loss without clear benefit.

Geometric Modeling Concerns: The EPD detector has a disk geometry, which differs fundamentally from cylindrical calorimeter geometries. The existing EpdGeom class was specifically designed to handle the r, φ, z coordinate system appropriate for disk detectors. Forcing EPD into the calorimeter tower model is precisely why the geometry functionality had to be duplicated from EpdReco into the new sEPDGeomMapping class - it's an indication that we're working against the natural geometry rather than with it.

Simulation Impact: The proposed changes would retire the EpdGeom class, which is currently used by simulation code. This creates a significant risk of breaking existing simulation workflows that depend on the current geometry implementation.

Recommendation: Before proceeding with this major refactoring, I suggest we:
Evaluate whether the benefits of calorimeter-style geometry justify the added complexity and code duplication
Ensure simulation compatibility is maintained throughout any transition
Consider a phased migration approach that preserves existing functionality

largewedge:
I do not completely disagree with @eumaka --I definitely think that whether the sEPD geometry handling should be done in CaloGeomMapping or in its own, separate module is worth further discussion, since CaloGeomMapping does seem to assume a cylindrical geometry; however, I do think that it's worthwhile to adhere more closely to the conventions set by the calorimeters, which would mean using RawTowerGeom objects when possible. 

Note that RawTowerGeomv1 itself does not assume anything about the geometry of the detector: it allows us to use cartesian coordinates to locate towers (https://github.com/sPHENIX-Collaboration/coresoftware/blob/dd264ae362a1b9f3d60aacd945552bc1258c6ca5/offline/packages/CaloBase/RawTowerGeomv1.h#L44 ), which should be as inexpensive as storing arm, ring, phi. Either way, we store three values.

Regarding the workflow: the first step ("copy phi, r, z values from EpdReco into sEPDGeomMapping with hardcoded phi values") was a part of the process of refactoring itself and will not be done every time the code is run, so this does not add anything to the computational cost. The remaining steps contribute a marginal amount of additional computing time and are done in other calorimeters without issue (for example, the HCal: https://github.com/sPHENIX-Collaboration/coresoftware/blob/dd264ae362a1b9f3d60aacd945552bc1258c6ca5/simulation/g4simulation/g4calo/HcalRawTowerBuilder.cc#L197 and the EMCal: https://github.com/sPHENIX-Collaboration/coresoftware/blob/dd264ae362a1b9f3d60aacd945552bc1258c6ca5/offline/packages/jetbase/TowerJetInput.cc#L497 ).

Regarding the simulation: I do not think that this would be that great of a challenge. There are relatively few places in the simulation code that I can see making use of EpdGeom (specifically here https://github.com/sPHENIX-Collaboration/coresoftware/blob/dd264ae362a1b9f3d60aacd945552bc1258c6ca5/simulation/g4simulation/g4epd/PHG4EPDModuleReco.cc#L82 ) although please correct me if I am wrong.

bseidlit:
GreatYes, I think we don't want to change the core functionality of the EPD for like 2-3 weeks.  Is the code that produces these calibrations on git.  If not, you could work on getting that into coresoftware or macros PR'ed and merged b.c. that won't effect the reconstruction but is also necessary.

PR #3843, opened 2025-08-22
Laser cluster identifier update

bkimelman:

## Types of changes

Flags events as able to have laser Gl1 trigger bit or not based on runnumber. If a run can't have this trigger bit, use original method for identifying and selecting/rejecting laser events; if it can have the trigger bit, use that flag to select/reject laser events. Also added BCO to LaserEventInfo for easy selection

## TODOs (if applicable)


osbornjd (changes requested):
Just saw that this was open. Given that this is adding member variables to LaserEventInfov2, this will render all previously created DSTs unreadable. You should either create a new v3 or risk not being able to access any existing DST

offline/packages/tpc/LaserEventInfov2.h
```diff
@@ -17,17 +17,25 @@ class LaserEventInfov2 : public LaserEventInfov1
...
+
  protected:
 
+  bool m_canHaveGl1Data{false};
   bool m_isGl1LaserEvent{false};
   bool m_isGl1LaserPileupEvent{false};
-
+  uint64_t m_bco{0};
+  
   ClassDefOverride(LaserEventInfov2, 1);
```
osbornjd, line 39: Okay after discussing with Chris we might be able to get away with just incrementing the class def override by 1 (I guess this should really be 2 currently...). So I would try increasing this and then attempting to readback one of our existing DSTs using any standard macro and see what happens

bkimelman: Removed these changes in favor of using runnumber instead. The Gl1 tigger bit was added at a fixed runnumber, so we now use that for any run after that run, or the old peak finding method for anything before that. This means no updates are needed in the LaserEventInfov2 class

PR #3881, opened 2025-09-19
Feat(JetBackgroundCut): Improve input flexibility and readability

Steepspace:
## Types of changes

This commit refactors the `JetBackgroundCut` module to enhance its flexibility and code clarity without altering the default behavior.

Key changes include:

- Configurable Inputs: Introduced setter methods (`set_input_cemc_tower_node` and `set_input_ohcal_tower_node`) to allow users to specify the input tower node names. The module now retrieves tower containers using these configurable names instead of hardcoded strings. The default node names remain the same, ensuring backward compatibility.

- Improved Readability: Replaced hardcoded integer values for jet component sources with their respective `Jet::SRC` enum members (e.g., `Jet::CEMC_TOWERINFO_RETOWER`, `Jet::HCALOUT_TOWERINFO`). This makes the source checking logic more explicit and maintainable.

- Support for Subtracted Jets: Added checks for the `Jet::CEMC_TOWERINFO_SUB1` and `Jet::HCALOUT_TOWERINFO_SUB1` enums. This enables the module to correctly process jets that have undergone underlying event subtraction and use the "SUB1" tower representation.

## TODOs (if applicable)

### DIFF
--- a/offline/packages/jetbackground/JetBackgroundCut.h
+++ b/offline/packages/jetbackground/JetBackgroundCut.h
@@ -90,14 +90,14 @@ class JetBackgroundCut : public SubsysReco
     _cutParams.set_int_param("failsAnyJetCut", 0);
   }
 
-  void set_input_cemc_tower_node(std::string input_cemc_tower_node)
+  void set_input_cemc_tower_node(const std::string &input_cemc_tower_node)
   {
-    m_input_cemc_tower_node = std::move(input_cemc_tower_node);
+    m_input_cemc_tower_node = input_cemc_tower_node;
   }
 
-  void set_input_ohcal_tower_node(std::string input_ohcal_tower_node)
+  void set_input_ohcal_tower_node(const std::string &input_ohcal_tower_node)
   {
-    m_input_ohcal_tower_node = std::move(input_ohcal_tower_node);
+    m_input_ohcal_tower_node = input_ohcal_tower_node;
   }
 
  private:
--- a/offline/packages/jetbackground/JetBackgroundCut.h
+++ b/offline/packages/jetbackground/JetBackgroundCut.h
@@ -90,6 +90,16 @@ class JetBackgroundCut : public SubsysReco
     _cutParams.set_int_param("failsAnyJetCut", 0);
   }
 
+  void set_input_cemc_tower_node(std::string input_cemc_tower_node)
+  {
+    m_input_cemc_tower_node = std::move(input_cemc_tower_node);
+  }
+
+  void set_input_ohcal_tower_node(std::string input_ohcal_tower_node)
+  {
+    m_input_ohcal_tower_node = std::move(input_ohcal_tower_node);
+  }
+
  private:
   bool _doAbort;
   std::string _name;
@@ -99,6 +109,9 @@ class JetBackgroundCut : public SubsysReco
   GlobalVertex::VTXTYPE _vtxtype;
   int _sysvar;
   PHParameters _cutParams;
+
+  std::string m_input_cemc_tower_node{"TOWERINFO_CALIB_CEMC_RETOWER"};
+  std::string m_input_ohcal_tower_node{"TOWERINFO_CALIB_HCALOUT"};
 };
 
 #endif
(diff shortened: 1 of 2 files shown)


offline/packages/jetbackground/JetBackgroundCut.h
```diff
@@ -90,6 +90,16 @@ class JetBackgroundCut : public SubsysReco
     _cutParams.set_int_param("failsAnyJetCut", 0);
   }
 
+  void set_input_cemc_tower_node(std::string input_cemc_tower_node)
```
pinkenburg, line 93: I am not opposed to new ideas - I welcome them. But please check if there is an advantage to using them - compared to our standard passing a reference to a string:
set_input_cemc_tower_node(std::string &input_cemc_tower_node)
{
m_input_cemc_tower_node = input_cemc_tower_node;
}

I got help but it sounds that passing the address is 1 copy, your construct adds a move to this and is therefore more expensive (in our use where we pass hardcoded strings from the macro)
--------------------------------------------------------------------------
Here is what chatgpt says:

Great question 👌 — this is a subtle but very important C++ API design point.

You’re comparing:

```cpp
// Version A: const reference
void set_string(const std::string& bla)
{
    std::string mystring = bla;
}
```

vs.

```cpp
// Version B: pass by value + move
void set_string(std::string bla)
{
    std::string mystring = std::move(bla);
}
```

---

## 🔎 Differences

### **Version A (const reference)**

* **Caller side:** No copy at the call site.
* **Inside function:** Always makes a copy (`mystring = bla`).
* **Total copies:** Always **1 copy**, no matter what argument you pass.
* **Pros:**

  * Fast if caller has an lvalue (already existing string).
  * No move semantics needed.
* **Cons:**

  * If caller passes a temporary (rvalue), you *still copy* instead of moving.

Example:

```cpp
std::string s = "hello";
set_string(s);            // 1 copy
set_string("temporary");  // 1 copy (no move)
```

---

### **Version B (by value + move)**

* **Caller side:**

  * If caller passes an **lvalue** → the parameter `bla` is copy-constructed.
  * If caller passes an **rvalue** → the parameter `bla` is move-constructed.
* **Inside function:** `mystring = std::move(bla)` moves from that local parameter.
* **Total copies/moves:**

  * lvalue arg → **1 copy + 1 move**
  * rvalue arg → **1 move**
* **Pros:**

  * Optimized for temporaries (rvalues): only 1 move, no copy.
  * Makes your API more flexible.
* **Cons:**

  * Slightly more expensive for lvalues (copy + move vs. just copy).
  * Might be misleading if you expect “always cheap.”

Example:

```cpp
std::string s = "hello";
set_string(s);            // 1 copy + 1 move
set_string("temporary");  // 1 move (fast!)
```

---

## ✅ Which One Should You Use?

* If **your function always copies** into a local variable anyway (like in your example), then **Version B (by value)** is often preferred in modern C++.

  * It gives you the best of both worlds: efficient moves for temporaries, and acceptable cost for lvalues.
* If **your function sometimes doesn’t need the value** (e.g., just reads or logs it), use **Version A (const ref)** — it avoids unnecessary copies/moves.

---

## 🚀 General Rule of Thumb

* **Use `const std::string&`** for read-only access.
* **Use `std::string` by value + `std::move`** when you need to *make a copy anyway*.

This is exactly the **“pass by value, then move”** idiom that Scott Meyers and Herb Sutter recommend for setters, constructors, etc.

---

👉 Example of a modern setter following this idiom:

```cpp
class Foo {
    std::string mystring;

public:
    void set_string(std::string s) {
        mystring = std::move(s); // efficient for rvalues, fine for lvalues
    }
};
```

---

Would you like me to also show you the **third option** (overload with both `const std::string&` and `std::string&&`) and explain when that’s better than just the pass-by-value version?

Steepspace: > I got help but it sounds that passing the address is 1 copy, your construct adds a move to this and is therefore more expensive (in our use where we pass hardcoded strings from the macro)

Ok, I can revert back to the traditional way of going about things which is using the Version A (const reference) approach.

PR #3884, opened 2025-09-19
feat(MinimumBiasClassifier): Add PHParameters for detailed cut tracking

Steepspace:
## Types of changes

This PR introduces `PHParameters` to the `MinimumBiasClassifier` to provide detailed, event-by-event information about why the minimum bias selection passes or fails. Previously, the module only set a single boolean, making it difficult to debug or perform quality assurance checks.

Key changes include:

- A `PHParameters` object named `"MinBiasParams"` is now created and stored on the `PAR` node.
- Individual boolean flags are set to `true` if an event fails a specific selection criterion, such as:
    - `minbias_background_cut_fail`
    - `minbias_two_hit_min_fail`
    - `minbias_zdc_energy_min_fail`
    - `minbias_mbd_total_energy_max_fail`
- Key calculated values (e.g., MBD north/south charge, vertex scale) are also saved to this object for downstream analysis and QA.
- The total MBD charge cut is no longer a hardcoded constant and can be configured via the new `set_mbd_total_charge_cut()` method. (Note: The default total MBD charge cut of 2100 remains unchanged.)

## TODOs (if applicable)

### DIFF
--- a/offline/packages/trigger/MinimumBiasClassifier.cc
+++ b/offline/packages/trigger/MinimumBiasClassifier.cc
@@ -91,6 +91,17 @@ int MinimumBiasClassifier::InitRun(PHCompositeNode *topNode)
 
   CreateNodes(topNode);
 
+  // Create Space on NodeTree to save Minimum Bias Params
+  PHNodeIterator parIter(topNode);
+  m_parNode = dynamic_cast<PHCompositeNode *>(parIter.findFirst("PHCompositeNode", "PAR"));
+  if (!m_parNode)
+  {
+    std::cout << "No RUN node found; cannot create PHParameters. Aborting run!";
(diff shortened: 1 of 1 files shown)


Steepspace:
> PHParameters are very expensive to use on an event by event basis. They are meant to be used for configurations and have a DB backend. This is just the wrong class to handle event by event parameters in this case,

I will note that PHParameters is currently used in the JetBackgroundCut module that also does something similar on an event by event basis: https://github.com/sPHENIX-Collaboration/coresoftware/blob/9d03281ebaecd5e37e533df03e9836e73602d5aa/offline/packages/jetbackground/JetBackgroundCut.cc#L283-L293 

However, since this is not recommended, do you suggest any other approach that will be ideal to pass the information regarding the cuts so that QA downstream can log that different cases?

PR #4263, opened 2026-05-07
Subdet track fits ii

gregottino:

## Types of changes

## TODOs (if applicable)

### Motivation & Context
This PR extends sPHENIX track reconstruction to support dedicated fitting modes for individual subdetectors. Previously, the Acts track fitter could combine Silicon + Micromegas (MM) hits in a single fit. This PR adds the capability to fit tracks using **Silicon hits only** or **TPC hits only**, enabling:
- Improved track quality when one subdetector has degraded performance
- Dedicated reconstruction modes for physics analyses requiring specific detector acceptance
- Integration with the track pruner module for improved track selection

### Key Changes
- **ActsSurfaceMaps**: Added `isSiSurface()` method and `m_siVolumeIds` member to classify silicon detector surfaces, parallel to existing TPC/Micromegas checks
- **MakeActsGeometry**: Populates silicon volume IDs from silicon cluster surface maps during geometry initialization  
- **PHActsTrkFitter Configuration**: 
  - Added `forceTpcOnlyFit(bool)` public setter (complementing existing `forceSiOnlyFit()`)
  - Both flags default to `false`
- **Track Fitting Logic**:
  - Initial position computation from silicon seed now conditional on silicon availability and `!m_forceTpcOnlyFit`
  - Seed phi/theta/eta now depend on silicon seed validity and fit mode
  - `filterSourceLinks()` now excludes Micromegas + Silicon surfaces when TPC-only mode is enabled; excludes TPC + Micromegas when silicon-only mode is enabled
  - State clearing modified to skip when forcing silicon-only/TPC-only fits
  - Momentum parameters (px/py/pz) copied from TPC seed rather than Acts fit result when `m_forceSiOnlyFit` is enabled

### Potential Risk Areas
- **Reconstruction Behavior Changes**: Track quality, momentum resolution, and angular resolution will vary significantly depending on fit mode. Physics analyses must carefully select the appropriate mode.
- **Track Parameters**: The conditional logic for seed usage and state clearing could produce different track parameter sets for the same input depending on configuration flags. Validation against Monte Carlo truth is recommended.
- **Surface Filtering Logic**: The new source-link filtering in `filterSourceLinks()` is critical for proper fit convergence in subdetector-only modes. Incorrect surface exclusion could lead to ghost tracks or fit failures.
- **IO Format**: No changes to persistent formats detected in this diff, but downstream code using track parameters should validate numerical stability.
- **Thread Safety**: No obvious thread-safety concerns in the added code, but validation against multi-threaded event loops recommended.

### Possible Future Improvements
- Add validation/documentation for physics behavior in each fit mode (Si-only, TPC-only, Si+MM)
- Consider exposing fit mode through macro configuration rather than only programmatic setters
- Extend track diagnostics to report which subdetectors contributed to each fit

---
**Note**: AI-assisted analysis was used in code examination; please verify critical logic paths manually, particularly the surface filtering and seed parameter logic.

### DIFF
--- a/offline/packages/trackreco/PHActsTrkFitter.cc
+++ b/offline/packages/trackreco/PHActsTrkFitter.cc
@@ -525,11 +525,11 @@ void PHActsTrkFitter::loopTracks(Acts::Logging::Level logLevel)
 
       // position comes from the silicon seed, unless there is no silicon seed
       Acts::Vector3 position(0, 0, 0);
-      if (siseed)
+      if (siseed && !m_ignoreSilicon)
       {
         position = TrackSeedHelper::get_xyz(siseed) * Acts::UnitConstants::cm;
       }
-      if (!siseed || !is_valid(position) || m_ignoreSilicon)
+      if (!siseed || !is_valid(position) || m_forceTpcOnlyFit)
       {
         position = TrackSeedHelper::get_xyz(tpcseed) * Acts::UnitConstants::cm;
       }
@@ -573,6 +573,13 @@ void PHActsTrkFitter::loopTracks(Acts::Logging::Level logLevel)
               continue;
             }
           }
+          //else if (m_forceTpcOnlyFit)
+          //{
+          //  if (surface_apr->geometryId().volume() < 14)
+          //  {
+          //    continue;
+          //  }
+          //}
           bool pop_flag = false;
           if (surface_apr->geometryId().approach() == 1)
           {
@@ -642,7 +649,7 @@ void PHActsTrkFitter::loopTracks(Acts::Logging::Level logLevel)
       float seedphi = 0;
       float seedtheta = 0;
       float seedeta = 0;
-      if (siseed)
+      if (siseed && !m_forceTpcOnlyFit)
       {
         seedphi = siseed->get_phi();
         seedtheta = siseed->get_theta();
@@ -980,6 +987,9 @@ SourceLinkVec PHActsTrkFitter::filterSourceLinks(const SourceLinkVec& sourceLink
     if (m_forceSiOnlyFit && (m_tGeometry->maps().isMicromegasSurface(surf) || m_tGeometry->maps().isTpcSurface(surf)) )
     { continue; }
 
+    if (m_forceTpcOnlyFit && (m_tGeometry->maps().isMicromegasSurface(surf) || m_tGeometry->maps().isSiSurface(surf)) )
+    { continue; }
+
     // update vectors
     filtered.push_back(sl);
   }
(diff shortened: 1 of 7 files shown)


offline/packages/trackreco/PHActsTrkFitter.cc
```diff
@@ -1097,9 +1107,20 @@ void PHActsTrkFitter::updateSvtxTrack(
...
+  
+  if(!m_forceSiOnlyFit)
+  {
+    track->set_px(params.momentum()(0));
+    track->set_py(params.momentum()(1));
+    track->set_pz(params.momentum()(2));
+  }
+  else
+  {
+    track->set_px(seed->get_px());
```
osbornjd, line 1116: I'm not sure I understand this, why would we set the output track to have momentum from the seed parameters?

offline/packages/trackreco/PHActsTrkFitter.cc
```diff
@@ -984,6 +991,9 @@ SourceLinkVec PHActsTrkFitter::filterSourceLinks(const SourceLinkVec& sourceLink
     if (m_forceSiOnlyFit && (m_tGeometry->maps().isMicromegasSurface(surf) || m_tGeometry->maps().isTpcSurface(surf)) )
     { continue; }
 
+    if (m_forceTpcOnlyFit && (m_tGeometry->maps().isMicromegasSurface(surf) || m_tGeometry->maps().isSiSurface(surf)) )
```
osbornjd, line 990: This seems to be only skipping the non TPC surfaces, but the source links are still added right?

offline/packages/trackreco/PHActsTrkFitter.cc
gregottino: In the case of the Si only track fit, the pT resolution is so bad it actually is a huge improvement to use the TPC seed pT for the final value of pT instead of the Si fitted pT.

offline/packages/trackreco/PHActsTrkFitter.cc
gregottino: Yeah I didn't try and skip adding sourcelinks, I just ignored them in the fitting process.

offline/packages/trackreco/PHActsTrkFitter.cc
osbornjd: So what happens if the silicon-TPC match was wrong? I'm not sure we can just unilaterally do this for all attempted track fits

offline/packages/trackreco/PHActsTrkFitter.cc
osbornjd: That means this will only work for the direct navigator then, I think

PR #4266, opened 2026-05-12
CaloTowerStatus: Prioritize Direct URL Overrides

Steepspace:
## Types of changes

Updated the logic in `InitRun` to ensure that URLs provided via `set_directURL_hotMap` and `set_directURL_chi2` take precedence over the default `CDBInterface` lookups. Previously, the code checked the CDB first, which caused user-specified URLs to act only as fallbacks rather than overrides.

The logic now:
1. Checks for a user-specified direct URL.
2. Falls back to the `CDBInterface` if no direct URL is provided.
3. Aborts or disables the masking if neither is found.

## TODOs (if applicable)

### Motivation
CaloTowerStatus previously queried the CDB first and treated user-provided direct URLs as fallbacks. That made it awkward to run with local calibration files when a CDB entry existed. This PR fixes the lookup order so explicit direct URLs reliably override CDB entries, enabling straightforward local testing and intended override behavior.

### Key Changes
- Direct-URL precedence: InitRun now checks the direct URL string members (m_directURL_chi2, m_directURL_hotMap) first; non-empty strings are used directly.
- Removed boolean flags: Eliminated use_directURL_chi2 and use_directURL_hotMap; presence of a non-empty direct URL string is now the single signal to use an override.
- CDB fallback: If no direct URL is provided, the code queries CDBInterface for the appropriate entry and constructs CDBTTree only when a URL is returned.
- Abort/disable behavior preserved: If neither a direct URL nor a CDB entry is found, behavior remains — abort if m_doAbortNoChi2 / m_doAbortNoHotMap are set; otherwise disable the corresponding masking and log the outcome.
- Logging improved: Clearer messages distinguish direct-URL usage, CDB lookups, and the fallback/abort paths.

### Potential Risk Areas
- IO/data formats: None — no changes to calibration file formats or downstream reconstruction objects; only the lookup source selection changed.
- Reconstruction behavior: Low risk — the same CDBTTree objects are produced when a source is available; behavior differs only in which source is chosen.
- Thread-safety: No new shared mutable state introduced; risk is unchanged.
- Performance: Negligible — simple string emptiness checks replace the previous boolean-flag logic.
- Configuration pitfalls: Users who previously relied on the boolean flags must ensure they provide non-empty direct URL strings; otherwise CDB will be consulted.

### Possible Future Improvements
- Validate direct-URL/file existence at InitRun and emit actionable errors (instead of only logging) to avoid silent misconfiguration.
- Add unit/integration tests covering direct-URL overrides, CDB fallbacks, and abort vs disable behavior.
- Consolidate and simplify related configuration handling to a single helper that returns the resolved URL (direct or CDB) to reduce duplication.

Notes
- Reviewer feedback favored removing boolean flags (to avoid a true flag with no filename); this PR implements that suggestion.
- AI-generated summaries can be imperfect — please review the diff to confirm behavior and messaging details.

[Review Change Stack](https://app.coderabbit.ai/change-stack/sPHENIX-Collaboration/coresoftware/pull/4266)

### DIFF
--- a/offline/packages/CaloReco/CaloTowerStatus.cc
+++ b/offline/packages/CaloReco/CaloTowerStatus.cc
@@ -133,7 +133,7 @@ int CaloTowerStatus::InitRun(PHCompositeNode *topNode)
       m_cdbttree_hotMap = new CDBTTree(calibdir_hotMap);
       if (Verbosity() > 1)
       {
-        std::cout << "CaloTowerStatus::Init " << m_detector << "  hot map found " << m_calibName_hotMap << " Ddoing isHot" << std::endl;
+        std::cout << "CaloTowerStatus::Init " << m_detector << "  hot map found " << m_calibName_hotMap << " Doing isHot" << std::endl;
       }
     }
     else
--- a/offline/packages/CaloReco/CaloTowerStatus.cc
+++ b/offline/packages/CaloReco/CaloTowerStatus.cc
@@ -81,22 +81,23 @@ int CaloTowerStatus::InitRun(PHCompositeNode *topNode)
   m_calibName_chi2 = m_detector + "_hotTowers_fracBadChi2";
   m_fieldname_chi2 = "fraction";
 
-  std::string calibdir = CDBInterface::instance()->getUrl(m_calibName_chi2);
-  if (!calibdir.empty())
+  std::string calibdir_chi2;
+  if (use_directURL_chi2)
   {
-     m_cdbttree_chi2 = new CDBTTree(calibdir);
-    if (Verbosity() > 0)
-    {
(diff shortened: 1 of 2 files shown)


pinkenburg:
The proliferation of flags is really concerning. Here an approach of setting the filename and only call the cdb if that filename is empty is more appropriate. That will simplify the code a lot and prevent bad usages where the flag is set to true but the filename is not set.

Steepspace:
> The proliferation of flags is really concerning. Here an approach of setting the filename and only call the cdb if that filename is empty is more appropriate. That will simplify the code a lot and prevent bad usages where the flag is set to true but the filename is not set.

@pinkenburg , thank you for the feedback. I have simplified the module by removing the additional boolean flags. If you could confirm it looks good on your end that would be great.

PR #4269, opened 2026-05-18
Streaming Lumi Counting and BCO Tagging

dillfitz:

## Types of changes

(1) Add Subsys reco module (`StreamingBcoLumiReco`) for producing DST from `BcoLumiReco` output that contains the gl1 bco, the corresponding streaming window, a boolean to encode if the gl1 bco is usable for analysis, and the luminosity in the streaming data. (2) Add corresponding PHObject classes for storing the luminosity and gl1 info in the DST. (3) Add Subsys reco module to test new DST output.

- First presentation to HF TG group [03/11/2026](https://indico.bnl.gov/event/32010/contributions/121414/attachments/68750/119649/lumi_counting_streaming_analysis_HF_TG_final_fixed.pdf)
- DST content proposed to HF TG group [04/15/2026](https://indico.bnl.gov/event/32496/contributions/123205/attachments/69483/119652/lumi_counting_streaming_analysis_update3_HF_TG_v2_fixed.pdf#page=4)

## TODOs (if applicable)

**PENDING DISCUSSION WITH THE BCO LUMI STREAMING GROUP** -- we should discuss the current structure of the code and if this is what we want moving forward. For example, there are 2 passes over the gl1 raw data which might not be necessary. Other options to consider are:

1. If we save the info we need for the lumi calculation to the DST created from `BcoLumiReco`, then we could just load the info directly from the DST. The relevant part of `StreamingBcoLumiReco` is [here](https://github.com/dillfitz/coresoftware/blob/master/offline/packages/bcolumicount/StreamingBcoLumiReco.cc#L128-L139).
2. We could consider moving all logic for saving the streaming DST upstream so that two passes are not done at all (e.g. handle everything in what is now called `BcoLumiReco` probably renaming to `StreamingBcoLumiReco` such that it outputs the equivalent of what is now output from `StreamingBcoLumiReco`). I avoided this as the use of `SyncReco` and the way the bco info was being accessed in `BcoLumiReco` was not as straightforward to me to work with for calculating the necessary quantities for the streaming bco lumi DST. However, now that the logic is in place in `StreamingBcoLumiReco`, it could be moved upstream if desired. 

- Corresponding PR in macros repo: https://github.com/sPHENIX-Collaboration/macros/pull/1327

### Motivation / Context
Add streaming DST output for BCO-based luminosity counting and GL1 BCO tagging so downstream analysis can use per-BCO streaming luminosity and GL1 tagging information. Presented to the HF TG group (2026-03-11) and DST content proposed 2026-04-15. Corresponds to macros PR sPHENIX-Collaboration/macros#1327.

### Key changes
- New reconstruction SubsysReco:
  - StreamingBcoLumiReco — reads GL1 packet 14001, extracts per-bunch and per-event BCO information, determines usable BCO tags, computes streaming windows, accumulates per-bunch counts, and computes raw/live/scaled luminosities. Writes results to DST nodes STREAMINGBCOINFO and STREAMINGLUMIINFO and produces monitoring histograms.
- New DST PHObject classes:
  - StreamingBcoInfo / StreamingBcoInfov1 — store GL1 BCO, evtno, usable BCO flag, and streaming window.
  - StreamingLumiInfo / StreamingLumiInfov1 — store raw, live, and scaled luminosity totals.
- DST writer / validation:
  - StreamingBcoLumiReco writes DST entries; StreamingBcoLumiCheck — SubsysReco test module validates/logs STREAMINGBCOINFO and STREAMINGLUMIINFO contents.
- Build and packaging:
  - Makefile.am updates to add new sources, ROOT dictionaries, and link dependency (-lfun4all).
- Cleanup:
  - Removal of the older BcoLumiCheck implementation.

### Potential risk areas
- IO/format: New DST nodes (STREAMINGBCOINFO, STREAMINGLUMIINFO) change output—downstream tools must be updated and code recompiled for new dictionaries.
- Reconstruction correctness: Per-bunch accumulation, bunch-number wrap-around (0–119) and abort-gap (>110) handling, and cross-section/scaler normalization are subtle — verify for off-by-one or normalization errors.
- Performance: Current workflow performs two passes over GL1 raw data (existing BcoLumiReco and the new streaming path), increasing CPU/IO for streaming workflows.
- Thread-safety / concurrency: Stateful per-run/per-bunch arrays and histogram writes may not be safe in multi-threaded contexts; review if used under parallel processing.
- Fixed sizing: Uses fixed 120-bunch arrays; unusual beam configurations could hit limits.

### Possible future improvements
- Eliminate double-pass by either persisting intermediate lumi data in BcoLumiReco-produced DST or moving streaming DST creation into BcoLumiReco to do counting once.
- Add unit/integration tests and validation macros exercising wrap-around, abort-gap, and scaler-normalization edge cases.
- Harden thread-safety (per-thread accumulators, protected histogram access) and make bunch-array sizing configurable or validated from run configuration.
- Improve logging/diagnostics and expose monitoring histograms to dashboards.

AI disclaimer: AI-generated summaries can contain mistakes—please verify the code and logic (especially wrap-around, abort-gap handling, and luminosity normalization) rather than relying solely on this summary.


offline/packages/bcolumicount/StreamingBcoInfov1.cc
```diff
@@ -0,0 +1,24 @@
...
+#include <phool/phool.h>
+
+#include <iostream>
+
+void StreamingBcoInfov1::Reset()
+{
+  set_bco(0);
+  set_usable_bco_tag(false);
+  set_bco_streaming_window(std::make_pair(0, 0));
+  return;
```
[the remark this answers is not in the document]
osbornjd: This is a relevant comment from the rabbit, we should reset all member variables of the EDM object. I think. Or perhaps the idea is to only set the event number once and have it be incremented on each process_event call? We may consider seeing if there is a "global" event number owned by fun4all that could be set here each reset call, so that it is guaranteed that this object is always in sync with Fun4all

offline/packages/bcolumicount/StreamingBcoLumiReco.cc
```diff
@@ -0,0 +1,281 @@
...
+      uint64_t bco_prev = bcoinfo->get_previous_bco();
+      uint64_t bco_futu = bcoinfo->get_future_bco();
+      uint64_t bco_diff_prev = m_bco - bco_prev;
+      uint64_t bco_diff_futu = bco_futu - m_bco;
+
+      // TODO: Set BCO window length in the macro 
+      // TODO: Set BCO Negative window length in the macro
+      // window length: 360, negative window length: 20
+      // Therefore, should check if BCO is within 340
+      // special case if BCO is within 20 of previous BCO?
```
osbornjd, line 189: This should never happen, I think Cameron and I found there is a minimum allowed bco diff of 17 for triggered events which Chris can confirm - this is a real condition in the DAQ

offline/packages/bcolumicount/StreamingLumiInfov1.cc
```diff
@@ -0,0 +1,26 @@
...
+
+void StreamingLumiInfov1::Reset()
+{
+  // Double check that this is only called once per run!! Or... it should just be called in the InitRun hook?
+  set_lumi_raw(0.);
+  set_lumi_live(0.);
+  set_lumi_scaled(0.);
+
+  return;
+}
```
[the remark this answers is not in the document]
osbornjd: Reset is called per event, so I don't think this is what you want (or at least, this won't do anything since you only fill these values within `StreamingBcoLumiReco::EndRun` which is correct)

offline/packages/bcolumicount/StreamingBcoLumiReco.cc
```diff
@@ -0,0 +1,281 @@
...
+
+      StreamingBcoInfo *streaming_bco_info = findNode::getClass<StreamingBcoInfo>(topNode, "STREAMINGBCOINFO");
+      m_bco = bcoinfo->get_current_bco();
+      if (gtm_bco != m_bco) { std::cout << "BCO MISMATCH!!! : gtm_bco : " << gtm_bco << " m_bco " << m_bco << std::endl;}
+      uint64_t bco_prev = bcoinfo->get_previous_bco();
+      uint64_t bco_futu = bcoinfo->get_future_bco();
+      uint64_t bco_diff_prev = m_bco - bco_prev;
+      uint64_t bco_diff_futu = bco_futu - m_bco;
+
+      // TODO: Set BCO window length in the macro
```
osbornjd: Why don't we just do this now? We can set the default value to 340 and then have the necessary setter so that it can be modified if necessary

offline/packages/bcolumicount/StreamingBcoLumiReco.h
```diff
@@ -0,0 +1,78 @@
...
+  TH1I *h_bco_diff_trigbits[40] = {nullptr};
+  TH1I *h_bco_tag = nullptr;
+
+  uint64_t m_bco{0};
+  int m_bunches = 120;
+  int m_evtno{0};
+  bool m_usable_bco_tag = false;
+  std::pair<uint64_t, uint64_t> m_bco_streaming_window;
+
+  double m_xsec_MBDNS = 24.07*1e9; //convert to pb from Vernier scan DOUBLE CHECK VALUE!
```
osbornjd, line 59: We may consider having this be a global value in https://github.com/sPHENIX-Collaboration/coresoftware/blob/master/offline/framework/phool/sphenix_constants.h so that all code can use the same value and we only have to update one place

offline/packages/bcolumicount/StreamingBcoLumiReco.cc
```diff
@@ -0,0 +1,281 @@
...
+      }
+      return Fun4AllReturnCodes::ABORTEVENT;
+    }
+    uint64_t gtm_bco = packet->lValue(0, "BCO");
+    uint64_t gl1_scaledvec = packet->lValue(0, "ScaledVector");
+    //uint64_t gl1_livevec = packet->lValue(0, "TriggerVector");   
+
+    int bunchno = packet->lValue(0,"BunchNumber");
+
+    // SYNTAX TAKEN FROM ZHIWANS CODE, why = and not +=? If this is correct it seems like a waste to call it for every event (would just need it for the last event in a particular crossing?)
```
osbornjd, line 141: @pinkenburg can you comment on this?

offline/packages/bcolumicount/StreamingLumiInfov1.cc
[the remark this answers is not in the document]
dillfitz: Should the reset function be removed entirely? I was following a template of a class filled in the process_event loop.

offline/packages/bcolumicount/StreamingBcoLumiReco.cc
dillfitz: @pinkenburg just pinging again now that I am picking this back up 😎

offline/packages/bcolumicount/StreamingBcoLumiReco.h
dillfitz: I think this is a good idea. Would you like me to prepare a separate PR? Then this one can be updated to read the cross section value from the constants file.

offline/packages/bcolumicount/StreamingBcoLumiReco.cc
dillfitz: Does this imply that there can still be a BCO diff of 18, 19, or 20 for triggered crossings? I think that would still qualify for the special case where a BCO occurs during the negative time window of the next triggered crossing. I had in mind to do a cross check for the special case to ensure the trigger bias algorithm works properly for this condition, but maybe it is a moot point.

offline/packages/bcolumicount/StreamingBcoLumiReco.cc
```diff
@@ -0,0 +1,281 @@
...
+
+int StreamingBcoLumiReco::EndRun(int /*runnumber*/)
+{
+  uint64_t rawgl1scalers_per_bunch = m_rawgl1scaler/120.;
+  for (int i=0; i<m_bunches; i++)
+  {
+    m_bunchnumber_lumi_raw[i] = ( m_bunchnumber_crossings[i] * m_bunchnumber_MBDNS_raw[i] ) / ( rawgl1scalers_per_bunch * m_xsec_MBDNS );
+    m_bunchnumber_lumi_live[i] = ( m_bunchnumber_crossings[i] * m_bunchnumber_MBDNS_live[i]) / ( rawgl1scalers_per_bunch * m_xsec_MBDNS );
+    m_bunchnumber_lumi_scaled[i] = ( m_bunchnumber_crossings[i] * m_bunchnumber_MBDNS_scaled[i] ) / ( rawgl1scalers_per_bunch * m_xsec_MBDNS );
+
```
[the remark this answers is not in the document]
dillfitz: @jdosbo some of these suggestions seem like overkill. I could put a check on the raw scalers I suppose but a check on the cross section seems like a waste of CPU. Do you generally advise following all of code rabbits suggestions that have not been flagged as incorrect by a reviewer?

offline/packages/bcolumicount/StreamingLumiInfov1.cc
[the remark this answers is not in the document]
osbornjd: You can, I just was mentioning that the code you have in the Reset function here doesn't actually do anything since these values are not processed/accessed/set in any way on an event by event basis

offline/packages/bcolumicount/StreamingBcoLumiReco.h
osbornjd: Sure, it should be a straightforward PR. And we will have to cross check it with the vernier scan value

offline/packages/bcolumicount/StreamingBcoLumiReco.cc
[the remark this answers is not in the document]
osbornjd: Definitely not. Everything AI generated should be approached with caution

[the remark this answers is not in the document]
dillfitz: Thanks for clarifying. I implemented all suggestions it had above this one as they seemed legitimate. I also implemted the suggestions below (aside from the comment on the StreamingLumiInfo Reset function). I did not implement this suggestion.

offline/packages/bcolumicount/StreamingLumiInfov1.cc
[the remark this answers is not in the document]
dillfitz: The Reset function has been removed from this class.

offline/packages/bcolumicount/StreamingBcoLumiReco.h
dillfitz: I made the PR here (https://github.com/sPHENIX-Collaboration/coresoftware/pull/4282)

offline/packages/bcolumicount/StreamingBcoLumiReco.cc
dillfitz: This is taken care of with the recent commits.

offline/packages/bcolumicount/StreamingBcoLumiReco.h
dillfitz: Thanks for merging this PR! Just a note that I did not get a chance to update the `m_xsec_MBDNS` value to read from `sphenix_constants.h` yet. The correpsonding PR (https://github.com/sPHENIX-Collaboration/coresoftware/pull/4282) is also merged, but I did not see the updated header in `$OFFLINE_MAIN/include/phool/`, so I was waiting for this to verify everything works as expected. 

We can easily update this in the future though. Looking back I would have preferred to rename `m_xsec_MBDNS` to `xsec_MBDNS` in (https://github.com/sPHENIX-Collaboration/coresoftware/pull/4282), but I forgot to make this change. I can prepare a new PR soon to take care of both small changes simultaneously. I just wanted to flag this for now.

PR #4287, opened 2026-06-05
Another try.

Ishangoel11:
[comment]: <> (Adding more variables for TPC Debugger.)

## Types of changes

## TODOs (if applicable)

### Motivation / context
Add a richer cluster representation (v6) to support TPC debugging and detailed offline diagnostics, and standardize cluster field types across existing versions (v1–v5) for tighter, explicit widths and consistent sentinel handling.

### Key changes
- New TrkrClusterv6 class (header, implementation, LinkDef, Makefile entries)
  - ~30+ new stored fields for debugging: centroid ADC, pad/TBin centers & maxima, directional edge flags (8 directions), mixing metrics, phi/tbin bin bounds, pad/tbin phase, RPhi/Z errors, subsurface key, local (rphi,z) position storage and bounds-checked accessors
  - PHObject/TrkrCluster overrides: identify, Reset, isValid, CloneMe, CopyFrom, get/set for all new fields
  - identify() prints a summary; isValid() checks positions and ADC sentinel (UINT16_MAX)
  - ROOT dictionary and install/listing updated (Makefile.am, TrkrClusterv6_Dict.cc entry)
- Base interface extended (TrkrCluster.h)
  - 20–30+ new virtual accessors/stubs added (getCenAdc, getPadCen/Max, getTBinCen/Max, SLEdge/SREdge/TLEdge/TREdge/DLEdge/DREdge/HLEdge/HREdge, mixing values, phi/tbin bin lo/hi, pad/tbin phase, getRSize, etc.)
  - Return types and sentinel defaults standardized (uint16_t/uint8_t/float; quiet_NaN for floats)
- Type narrowing and standardization across v1–v5 cluster classes and implementations
  - ADC fields: unsigned int → uint16_t; sentinel values updated to std::numeric_limits<uint16_t>::max()
  - Size/edge/overlap fields: char → uint8_t
  - Float NaN sentinel handling: use std::numeric_limits<float>::quiet_NaN()
  - ClassDefOverride version bumps where appropriate (v1–v5)
- New TrkrClusterv6 implementation includes CopyFrom to populate new fields from TrkrCluster interface

### Potential risk areas
- IO / on-disk / ROOT compatibility
  - Narrowing ADC from 32→16 bits and changing types for multiple fields can break ROOT serialization/reading of older files or change binary layouts; validate backward/forward compatibility and ROOT dictionary versions before wide deployment.
- Reconstruction / algorithm behavior
  - Downcasting to uint16_t may truncate values if any code assumed 32-bit ADCs; algorithms that read/write cluster fields or compute aggregates may behave differently.
- API / compilation breakage
  - Method signature changes in v1–v5 are public API changes and will require rebuilding dependent code; external code relying on previous signatures must be updated.
- Missing container/version support
  - No TrkrClusterContainerv6 was added — persistence/containers for v6 clusters are not provided here; this limits how v6 clusters are stored or transported unless intended to use existing containers.
- Logging / runtime noise
  - Deprecated global getters print to stdout directly (not using logging), which can pollute logs in large-scale reconstruction or batch jobs.
- Thread-safety & performance
  - Added fields and deeper CopyFrom may increase memory footprint and copy cost; inspect hot paths for performance regressions. No explicit thread-safety changes, but added mutable state increases the surface area; ensure usages are thread-safe where needed.

### Possible future improvements
- Add TrkrClusterContainerv6 (and LinkDef) if persistent storage of v6 clusters is desired.
- Replace stdout deprecation messages with the project logging framework and/or a compile-time deprecation mechanism.
- Provide migration/readback tests: validate reading older files, and a compatibility shim or upgrade path if needed.
- Consider making many new TrkrCluster virtual setters optional or removing obsolete setters from the virtual base class (per reviewer suggestion) to avoid propagating no-op/obsolete API.
- Add unit/integration tests for serialization, CopyFrom correctness, and reconstruction end-to-end validation; measure memory and CPU impact.

⚠️ This summary was generated with AI assistance—verify type-change impacts, ROOT dictionary/versioning, and storage/compatibility implications with actual code reviewers before merge.


Ishangoel11:
So when will be added?

offline/packages/trackbase/TrkrClusterv6.h
```diff
@@ -0,0 +1,276 @@
+/**
+ * @file trackbase/TrkrClusterv6.h
+ * @author Ishan Goel
+ * @date May 2026
+ * @brief Version 6 of TrkrCluster
+ */
+#ifndef TRACKBASE_TRKRCLUSTERV6_H
+#define TRACKBASE_TRKRCLUSTERV6_H
+
+#include <iostream>
```
pinkenburg: The include file ordering is really important:
https://wiki.sphenix.bnl.gov/index.php?title=Codingconventions#Include_Files
I see other sources here having the same problem - don't worry about that - we'll deal with those later.

offline/packages/trackbase/TrkrClusterv6.h
```diff
@@ -0,0 +1,276 @@
...
+  TrkrDefs::subsurfkey getSubSurfKey() const override { return m_subsurfkey; }
+  void setSubSurfKey(TrkrDefs::subsurfkey id) override { m_subsurfkey = id; }
+
+  //
+  // cluster info
+  //
+  unsigned int getAdc() const override { return m_adc; }
+  void setAdc(unsigned int adc) override { m_adc = adc; }
+
+  unsigned int getMaxAdc() const override { return m_maxadc; }
```
pinkenburg: Is it unsigned int or uint16_t? Please make this consistent. Likely m_adc should be the same data type

offline/packages/trackbase/TrkrClusterv6.h
```diff
@@ -0,0 +1,276 @@
...
+  float getZError() const override { return m_zerr; }
+
+  void setPhiError(float phierror) { m_phierr = phierror; }
+  void setZError(float zerror) { m_zerr = zerror; }
+
+  /// deprecated global funtions with a warning
+  float getX() const override
+  {
+    std::cout << "Deprecated getx trkrcluster function!" << std::endl;
+    return NAN;
```
pinkenburg: use td::numeric_limits<float>::quiet_NaN() instead of NAN

offline/packages/trackbase/TrkrClusterv6.h
```diff
@@ -0,0 +1,276 @@
...
+  float m_zerr;
+  unsigned short int m_adc;     //< cluster sum adc 16
+  unsigned short int m_maxadc;  //< cluster max adc 16
+  unsigned short int m_cenadc;  //< cluster centroid adc 16
+  float m_padcen;
+  float m_tbincen;
+  float m_padmax;
+  float m_tbinmax;
+  unsigned char m_rsize;        // 8bit
+  char m_phisize;               // 8bit
```
pinkenburg: A lot of the char's look like unsigned int's and all are initialized to 0, not to some invalid negative value. , So they might better be unsigned chars

offline/packages/trackbase/TrkrClusterv6.h
```diff
@@ -0,0 +1,276 @@
...
+  char m_tredge;                // 8bit
+  char m_dledge;                // 8bit
+  char m_dredge;                // 8bit
+  char m_hledge;                // 8bit
+  char m_hredge;                // 8bit
+  char m_slmix;                 // 8bit
+  char m_srmix;                 // 8bit
+  char m_tlmix;                 // 8bit
+  char m_trmix;                 // 8bit
+  float m_phibinlo;
```
pinkenburg: bin sounds like it should be an integer data type

offline/packages/trackbase/TrkrClusterv6.h
```diff
@@ -0,0 +1,276 @@
...
+  // float getZSize() const override
+  //{std::cout << "Deprecated size function" << std::endl; return NAN;}
+  // float getPhiError() const override
+  //{ std::cout << "Deprecated getPhiError function"<< std::endl; return NAN;}
+
+ private:
+  float m_local[2]{std::numeric_limits<float>::quiet_NaN(), std::numeric_limits<float>::quiet_NaN()};
+  //< 2D local position [cm] 2 * 32 64bit  - cumul 1*64
+  TrkrDefs::subsurfkey m_subsurfkey;  //< unique identifier for hitsetkey-surface maps 16 bit
+  float m_phierr;
```
pinkenburg: move the long list of initializers from the .cc to here

offline/packages/trackbase/.TrkrClusterv6.h.swo
pinkenburg, line 1: remove the binary, it must not go into our repo

offline/packages/trackbase/TrkrClusterv6.cc
```diff
@@ -0,0 +1,141 @@
...
+int TrkrClusterv6::isValid() const
+{
+  for (int i = 0; i < 2; ++i)
+  {
+    if (std::isnan(getPosition(i)))
+    {
+      return 0;
+    }
+  }
+  if (m_adc == 0xFFFF)
```
pinkenburg: use std::numeric_limits<unsigned short>::max() instead of 0xFFFF

offline/packages/trackbase/TrkrClusterv6.cc
```diff
@@ -0,0 +1,141 @@
...
+{
+  // square convenience function
+  template <class T>
+  constexpr T square(const T& x)
+  {
+    return x * x;
+  }
+}  // namespace
+
+TrkrClusterv6::TrkrClusterv6()
```
pinkenburg: once you move the initializers to the .h file, the ctor is empty and can be declared default in the .h file

pinkenburg (changes requested):
let's try this again - I cannot see a trace of these comments which I made before in the history

pinkenburg:
This is more of a general comment for this package - the v6 contains a lot of implementations of seemingly obsolete methods. If they are are really obsolete the setters should be removed from the virtual base class. this will leave any readback of old versions functional but it will stop the propagation of the use of obsolete set methods which do not have any effect

Ishangoel11:
I hope this is good!

pinkenburg:
I fear this is completely broken now. We cannot change any data members of existing classes, these changes will render existing DSTs unreadable. I would suggest to preserver the v6 and start a fresh PR with those. We probably should discuss this in person first - I see a few ways to do this but it depends on what you actually need. Also your environment for some reason adds hidden files like .TpcClusterizer.cc.swo to the PR - you really need to change that.

osbornjd (commented):
I would agree with Chris, this will break all of our DST readback. Perhaps we should chat about what you are trying to do - at the very least we should just start with only the v6 version and getting that in

PR #4289, opened 2026-06-05
speed up by x1000

pinkenburg:

## Types of changes
- [X ] New feature (non-breaking change which adds functionality)

codex managed to speed up SvtxTruthRecoTableEval by x1000. The resulting root files are binary identical with the previous version. This needs to be reviewed

## TODOs (if applicable)

## Motivation and Context
SvtxTruthRecoTableEval builds mapping tables between simulation truth particles and reconstructed SVTX tracks for downstream evaluation. This PR consolidates truth→reco and reco→truth mapping into a single pass to drastically reduce runtime (~1000× reported) while preserving the ROOT output format.

## Key Changes
- Single-pass mapping: Replaces separate fillTruthMap()/fillRecoMap() with fillTruthRecoMaps(topNode, trackeval, verbosity) that populates m_truthMap and m_recoMap in one traversal of SvtxTrackMap.
- Verbosity & eval stack: process_event caches Verbosity(), sets it on m_svtxevalstack and SvtxTrackEval (asserting the latter), and issues a single combined log message when verbose.
- Momentum pre-filtering: Selects truth particles (primary vs full per m_scanForPrimaries) and filters by squared momentum against m_minMomentumTruthMap^2 to skip low-momentum particles early.
- Cluster-based aggregation: Uses SvtxClusterEval::all_truth_particles() and per-seed cluster keys to count contributions per truth track id and build weighted mappings for reco→truth and aggregated truth→reco.
- Fast-sim handling: Preserves special-case SvtxTrack_FastSim behavior (uses get_truth_track_id() and trackeval->get_nclusters_contribution()).
- API/header update: Header now forward-declares SvtxTrackEval and declares fillTruthRecoMaps(...); removed separate fillTruthMap/fillRecoMap definitions from the .cc file and updated includes (added SvtxClusterEval).
- Logging reduction: Eliminates per-map repeated logging in favor of a single combined message (verbosity gated).

## Potential Risk Areas
- Output equivalence: Author reports binary-identical ROOT files; reviewers must still validate with regression tests and event-sample comparisons to confirm semantic equivalence.
- Fast-sim correctness: Fast-sim path preserved but verify weighting/cluster-count semantics against legacy behavior with fast-sim testcases.
- Memory usage: The single-pass algorithm builds temporary per-event maps keyed by truth id — check memory use for high-multiplicity events.
- Thread-safety/concurrency: No explicit thread-safety changes; consolidated temporary structures should be reviewed before running in multi-threaded processing.
- Dependency stability: Changes rely on SvtxClusterEval::all_truth_particles() and seed-cluster iteration semantics; interface changes there could silently alter mapping results.

## Possible Future Improvements
- Cache cluster→truth results within the event to avoid repeated clustereval lookups.
- Profile remaining hotspots (cluster evaluation, unordered_map/set operations) and consider parallelizing safe per-track work.
- Add automated performance and regression tests (including fast-sim cases) to guard against regressions and ensure binary-identical outputs.
- Consider memory-limited or streaming strategies for very high-multiplicity events.

Note: AI summaries can miss subtle details—please review cluster-matching, weight accumulation, and the fast-sim path carefully during code review.

### DIFF
--- a/simulation/g4simulation/g4eval/SvtxTruthRecoTableEval.cc
+++ b/simulation/g4simulation/g4eval/SvtxTruthRecoTableEval.cc
@@ -2,7 +2,8 @@
 #include "SvtxTruthRecoTableEval.h"
 #include "SvtxEvalStack.h"
 #include "SvtxTrackEval.h"
-#include "SvtxTruthEval.h"
+
+#include "SvtxClusterEval.h"
 
 #include <fun4all/Fun4AllReturnCodes.h>
 #include <phool/PHCompositeNode.h>
@@ -19,10 +20,18 @@
 #include <trackbase_historic/PHG4ParticleSvtxMap_v1.h>
 #include <trackbase_historic/SvtxPHG4ParticleMap_v1.h>
 #include <trackbase_historic/SvtxTrack.h>
+#include <trackbase_historic/SvtxTrack_FastSim.h>
 #include <trackbase_historic/SvtxTrackMap.h>
 
-#include <CLHEP/Vector/ThreeVector.h>
-
+#include <cassert>
+#include <cstddef>
+#include <iostream>
+#include <map>
+#include <set>
+#include <unordered_map>
+#include <unordered_set>
+#include <utility>
+#include <vector>
 
 //____________________________________________________________________________..
 SvtxTruthRecoTableEval::SvtxTruthRecoTableEval(const std::string &name)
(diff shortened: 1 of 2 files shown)


simulation/g4simulation/g4eval/SvtxTruthRecoTableEval.cc
```diff
@@ -102,111 +111,137 @@ int SvtxTruthRecoTableEval::End(PHCompositeNode * /*unused*/)
...
-                                Hep3Vector(g4particle->get_px(), g4particle->get_py(), g4particle->get_pz())
-                                    .mag();
+    const double px = g4particle->get_px();
+    const double py = g4particle->get_py();
+    const double pz = g4particle->get_pz();
+    const double momentum2 = px * px + py * py + pz * pz;
 
     // only record particle above minimal momentum requirement.
-    if (momentum < m_minMomentumTruthMap)
+    if (momentum2 < minMomentumTruthMap2)
```
osbornjd, line 143: Not sure why codex decided to change this from pT to pT^2 but in principle I don't think it matters, other than being potentially confusing when reading the code later

osbornjd (approved):
I tested this on the oo sims and with some default sims and inspected the output from the module with the current new build and with this PR, and the results look identical for a single event. Reading through the code it looks to do the same thing in just a much more clever way rather than going track by track and using the evaluation machinery. We may consider seeing what else codex can improve in the evaluation machinery because it all uses stl containers probably not in the most simplified way...

simulation/g4simulation/g4eval/SvtxTruthRecoTableEval.cc
pinkenburg: I added a comment in the code - to point this out. It does save us a slow sqrt operation which is likely the reason why codex implemented the p cut this way

PR #4290, opened 2026-06-07
INTT hit duplication implementation

hrjheng:

## Types of changes

This PR introduces an optional INTT hit-duplication implementation in `Fun4AllStreamingInputManager` as a possible mitigation for the hit carry-over issue. When enabled, raw hits found at a later BCO, (N + M * 120), are copied into the raw-hit container with their BCO reassigned to N.
The duplication is disabled by default. The maximum shift multiple, controlled by the variable `m_InttHitCarryOverShiftMaxMultiple`, is configurable, while the carry-over shift is currently fixed at 120.

[Update 2026.06.11] The idea is to introduce an additional "second" duplication pass, in which hits in the next (or subsequent) strobes with FPHX BCO = 0 are also duplicated. I suspect this is necessary in addition to the "first" duplication pass because your previous slides indicate a non-negligible population of hits whose FPHX BCO is reset, particularly near the end of a strobe (two spikes appear in the subsequent strobe: one at FPHX BCO = 0 and another at the nominal +120 shift)
To avoid duplicating the same hits twice, the "second" pass is skipped whenever the target BCO (i.e. the BCO to which hits are being duplicated) corresponds to FPHX BCO = 0. In that case, the first duplication has already recovered all possible carried-over hits

(This strategy still requires careful study and discussion, so the PR is currently marked as a draft)

## TODOs (if applicable)

# INTT hit carry-over mitigation: optional hit-duplication in `Fun4AllStreamingInputManager`

## Motivation / Context
INTT streaming can exhibit “hit carry-over” behavior where hits associated with later BCOs are effectively carried into earlier strobe windows (linked to FPHX BCO resets). This PR adds an *optional*, off-by-default mitigation that duplicates affected raw INTT hits into earlier BCOs to recover the intended hit timing.

**AI note:** AI-generated summaries can contain mistakes—please use best judgment and validate details during review.

## Key changes
- **New configuration API (off by default) in `Fun4AllStreamingInputManager.h`:**
  - `EnableInttHitDuplication(bool b = true)`
  - `InttHitCarryOverShiftMaxMultiple(int i)`
  - Internal defaults: `m_InttHitDuplication{false}`, carry-over shift base fixed at **120**, and max multiple defaulting to **4**
- **Extended INTT raw hit filling in `Fun4AllStreamingInputManager.cc` (`FillIntt()`):**
  - When `m_InttHitDuplication` is enabled:
    - **Pre-reading/arming step:** repeatedly calls `FillInttPool()` until the highest stashed INTT BCO reaches `select_crossings + (120 * maxMultiple)`, including a guard to break if `FillInttPool()` doesn’t advance the highest BCO (avoids infinite loops).
    - **Two-pass duplication into `InttRawHitContainer`:**
      - **Pass 1:** for each target BCO `N`, copy hits found at `N + k*120` for `k=1..maxMultiple`, and reassign the copied hit BCO back by `-k*120` (with additional checks on hit BCO relative to the shift).
      - **Pass 2 (FPHX reset recovery):** for the same `N`, attempt an additional copy targeting the case where the copied hits correspond to **FPHX BCO = 0**.
        - The second-pass copy is **skipped** when `target_refbco == 0` (derived from `N % 120`) to avoid double-duplicating the FPHX-bco=0 hits already covered by Pass 1.
        - Second-pass copies only proceed if `source_hit->get_FPHX_BCO() == 0` and the hit BCO is not earlier than the computed `reset_shift`.
  - If duplication is disabled, behavior is unchanged aside from minor non-semantic iterator formatting.

- **Minor code-style changes (no semantic intent):**
  - INTT/MVTX map iteration uses `auto &` in updated loops.

## Potential risk areas
**AI note:** AI can be wrong—please verify with targeted tests/QA.

- **Reconstruction / physics behavior:** duplication increases INTT raw hit multiplicity and modifies BCO assignments; this can affect downstream hit building and any analysis relying on raw-hit time structure.
- **Correctness/edge cases:** the combined “pre-read to threshold + two-pass duplication + skip condition when `N % 120 == 0`” is subtle; off-by-one/BCO-boundary conditions could lead to missed or extra duplicates near strobe boundaries.
- **Performance & memory:** pre-reading extends pooling until a higher BCO threshold, and duplication can add up to `maxMultiple` (plus second-pass candidates) copies per hit; this may increase CPU time and memory use.
- **Thread-safety:** changes operate on shared instance state (`m_InttRawHitMap`, `m_RefBCO`, etc.); if this component were used concurrently, risks remain (though existing design appears single-threaded).

- Add/extend **QA validation** (e.g., compare reconstructed/selected hit rates and timing distributions with duplication on vs off).
- Consider **unit/regression tests** for the duplication logic (especially around FPHX BCO reset transitions and the `target_refbco==0` skip).
- Explore configurability of the base **120-BCO shift** (currently fixed) and document recommended values per running mode.
- Optimize duplication loops (e.g., reduce repeated map lookups) if performance becomes limiting.


osbornjd (commented):
This seems like a reasonable implementation to me, we should test it out and see if it improves the cluster/hit drop off towards the end of the strobe width

hrjheng:
Closing this PR because Cheng-Wei has created a new one: https://github.com/sPHENIX-Collaboration/coresoftware/pull/4324

Cheng-Wei can provide the full explanation, but here is my understanding of the differences between the two implementations:
1. Both implementations use the same two-pass duplication strategy. The first duplication pass is identical (at least conceptually).
2. Cheng-Wei pointed out that, although my second duplication pass correctly reassigns the FPHX BCO, it does not update `BCO_full`. As a result, the duplicated hits are not actually assigned to the intended strobe. Cheng-Wei's implementation corrects this by updating `BCO_full` together with the FPHX BCO.
3. Cheng-Wei's implementation also duplicates hits with FPHX BCO values from 0 through 5, rather than only FPHX BCO = 0. The hypothesis is that when the FPHX BCO is reset, it may be reset to any value between 0 and 5, randomly or not. This hypothesis has not yet been verified.

To test 3, Cheng-Wei will compare the INTT doublet distributions as a function of FPHX BCO using two configurations:
* duplicating only hits with FPHX BCO = 0 (with the corrected `BCO_full` assignment), and
* duplicating hits with FPHX BCO values from 0 to 5.
If the two approaches produce significantly different results, that would provide evidence supporting the hypothesis that the reset FPHX BCO can take values other than 0. If there is little or no difference, the additional duplication may not be necessary.

PR #4317, opened 2026-06-25
Separate luminosity calculation

dillfitz:
Separate luminosity calculation into its own Subsys reco module (StreamingLumiReco)

## Types of changes

It is a refactoring of an existing feature -- separating the streaming luminosity calculation from the trigger bco validation code.

This follows work started in http://github.com/sPHENIX-Collaboration/coresoftware/pull/4269.

## TODOs (if applicable)

- Discuss with team if the current refactoring is as desired. It is still a bit unclear to be if the streaming windows calculated in `StreamingBcoReco` will be passed unmodified to the lumi calculation (as they are now). My understanding is that they need to go through a QA step first, so we may need to modify the sequence in the future. 
- Check if removal of `process_event` loop in `StreamingLumiCheck` is okay. In principal it is not needed, the relevant checks take place right in the `InitRun` sequence. If so, remove commented out block (it was left in for visibility)

Here is the current sequence:

```mermaid
graph TD;
    BcoLumiReco-->StreamingBcoReco;
    StreamingBcoReco-->StreamingLumiReco;
    StreamingBcoReco-->Analyst;
    StreamingLumiReco-->Analyst;
```

I believe in the future it will be

```mermaid
graph TD;
    BcoLumiReco-->StreamingBcoReco;
    StreamingBcoReco-->StreamingDetectorQA;
    StreamingDetectorQA-->StreamingLumiReco;
    StreamingBcoReco-->Analyst;
    StreamingLumiReco-->Analyst;
```

https://github.com/sPHENIX-Collaboration/macros/pull/1347

Motivation / context  
- Refactors the streaming luminosity workflow into a dedicated `StreamingLumiReco` subsystem, separating luminosity calculation from BCO validation logic.
- Follows earlier streaming reconstruction work and aligns the BCO/lumi pipeline with the new module split.

Key changes
- Added `StreamingLumiReco` as a new reconstruction module for luminosity calculation.
- Updated `StreamingLumiReco` to consume `StreamingBcoReco` output instead of `BcoLumiReco`.
- Split out `StreamingBcoReco` / `StreamingBcoCheck` and removed `InitRun` from `StreamingBcoReco`.
- Updated `StreamingLumiCheck` / `StreamingBcoCheck` headers and implementations to reflect the new module boundaries.
- Extended `StreamingLumiInfo` / `StreamingLumiInfov1` with per-bunch lumi storage/accessors.
- Updated build/install lists in `Makefile.am` for the new/renamed headers and sources.

Potential risk areas
- IO / data model changes: new node names and changed producer/consumer relationships may require downstream macro updates.
- Reconstruction behavior changes: moving lumi accumulation out of the BCO validation path can alter event/run timing and output content.
- Performance: per-bunch array handling and additional module separation may affect runtime slightly.
- Thread-safety: no explicit changes indicated, but new shared-node read/write paths should be reviewed carefully.

Possible future improvements
- Decide whether streaming windows should pass through QA before luminosity calculation.
- Consider removing the `process_event` loop in `StreamingLumiCheck` if it is no longer needed.
- Update macros and any downstream analysis code to match the new module interfaces and node layout.

AI-generated summaries can make mistakes; please use best judgment when reviewing the changes.


offline/packages/bcolumicount/StreamingBcoReco.cc
```diff
@@ -59,7 +57,8 @@ int StreamingBcoLumiReco::Init(PHCompositeNode *topNode)
   return iret;
 }
 
-int StreamingBcoLumiReco::InitRun(PHCompositeNode * topNode)
+// Do we even need to include this now that the lumi calculatin has been separated? Or should I remove `InitRun` entirely?
```
osbornjd: It doesn't do anything, I suggest removing it

offline/packages/bcolumicount/StreamingBcoReco.h
```diff
@@ -0,0 +1,53 @@
...
+#include <fun4all/SubsysReco.h>
+#include <fun4all/Fun4AllHistoManager.h>
+
+#include <string>
+#include <utility>
+#include <array>
+
+class TH1;
+
+class StreamingBcoReco : public SubsysReco
```
osbornjd, line 13: I think this is replacing `StreamingBcoLumiReco.h` but you didn't do a `git mv StreamingBcoLumiReco.h StreamingBcoReco.h` - is my understanding right? Then we should `git rm` the old file

offline/packages/bcolumicount/StreamingLumiReco.cc
```diff
@@ -0,0 +1,217 @@
...
+    std::cout << PHWHERE << " DST Node is missing doing nothing" << std::endl;
+    return Fun4AllReturnCodes::ABORTRUN;
+  }
+  return Fun4AllReturnCodes::EVENT_OK;
+}
+
+int StreamingLumiReco::process_event(PHCompositeNode *topNode)
+{
+  StreamingBcoInfo *streaming_bcoinfo = findNode::getClass<StreamingBcoInfo>(topNode, "STREAMINGBCOINFO");
+  Event *evt = findNode::getClass<Event>(topNode, "PRDF");
```
osbornjd, line 83: I just realized we have a problem here. If we read the raw PRDFs then we will never be able to delegate this to analyzers, because there is not enough disk space to keep the PRDFs disk resident. So we need to modify this to read from the `GL1Packet` object - I think the information content should be exactly analogous it is just reading from the "offline" GL1 object instead of the PRDF version. For reference, [here](https://github.com/sPHENIX-Collaboration/coresoftware/blob/c89585ee3c9764fe796763a093b9b704918a46f4/offline/framework/fun4allraw/SingleGl1PoolInput.cc#L111-L137) is where the GL1 object gets filled in the event combiner - I think that should be enough to show you how to read the relevant information from this

That way the input for this module can be
1. STREAMINGBCOINFO object in the "specially" made DST
2. The GL1RAWHIT object, which we explicitly pass to every single DST so that the trigger/GL1 information is available at every level

osbornjd (changes requested):
A few comments

1. Is `StreamingBcoCheck` just intended to be a diagnostic module? Or am I missing its purpose? Same with `StreamingLumiCheck`?
2. Related to the above, I guess it is replacing `StreamingBcoLumiCheck`? Why only delete the header file and not also the `.cc` file?

I left some individual comments also. Some are rather large (e.g. moving from PRDF -> GL1RAWHIT) which we could delegate to another PR instead of adding it to this one

offline/packages/bcolumicount/StreamingBcoReco.h
dillfitz: I think I did not use the cleanest workflow here. I started with the `StreamingBcoLumiReco*` files, made copies of them (just with `cp`), renamed the old and new set of files (just with `mv`), and then dealt with all of the git stuff when all was said and done just using `git rm` and `git add`. I must have missed one, and I don't know how git decided which files I renamed since I never used `git mv`. I can be more careful about this in the future. 

I believe you are right and we can `git rm StreamingBcoLumiReco.h`

offline/packages/bcolumicount/StreamingLumiReco.cc
dillfitz: Yes this makes sense to me. We can also consider saving the trigger counters in the `StreamingBcoInfo` DST since it has to do a pass over the GL1 raw data, but I am not sure if that workflow makes as much sense.

offline/packages/bcolumicount/StreamingBcoReco.h
dillfitz: After double checking, it looks like all the correct files are already there with no extras, though it looks very confusing how git decided to rename vs remove / add files. Ignoring the small changes to the `StreamingLumiInfo` class, I had to create 4 new files, and modify 4 existing files, basically by copying the 4 old `StreamingBcoLumiReco` and `StreamingBcoLumiCheck` files and renaming them accordingly. In the end, git decided that I added 5 files, removed 1, and renamed 3, which is effectively the same as renaming 4 and adding 4. I think git decided some files were just renamed based on how closely they resembled the old files -- but I only used `git add` and `git rm`.

offline/packages/bcolumicount/StreamingBcoReco.cc
dillfitz: Done

offline/packages/bcolumicount/StreamingLumiReco.cc
dillfitz: Thanks for the review. I have addressed your other comments. Do you prefer we address this issue in this PR or a separate one? 

For reference, the corresponding macros PR has already been merged (https://github.com/sPHENIX-Collaboration/macros/pull/1347)

osbornjd: > We can also consider saving the trigger counters in the StreamingBcoInfo DST since it has to do a pass over the GL1 raw data, but I am not sure if that workflow makes as much sense.

That seems unnecessary since this info is already stored in the `GL1RAWHIT` object, so it would be duplicating the information.

We can make the switch from the prdf object to the gl1rawhit object in another PR

PR #4342, opened 2026-07-10
Adding new TrkrHitv3 for the INTT timing payload

ChengWeiShih:
The PR is to associate the INTT timing information with the TrkrHit, which can be important for the study of the INTT carried-over-hit issue in streaming.

## Types of changes

[comment]: 
Three files are added 
`offline/packages/trackbase/TrkrHitv3.cc`
`offline/packages/trackbase/TrkrHitv3.h`
`offline/packages/trackbase/TrkrHitv3LinkDef.h`, and 
`offline/packages/trackbase/Makefile.am` is updated to include them

`offline/packages/trackbase/TrkrHit.h` is updated to include `SetFPHXBCO()` and `SetBCO()` and so on.

And  `offline/packages/intt/InttCombinedRawDataDecoder.cc` is updated to assign the INTT_FPHX_BCO and INTT_long_BCO (BCO_full) to Trkrhit.

## TODOs (if applicable)

[comment]: <>

## Motivation
Preserve INTT timing (FPHX BCO and full BCO) on tracking hits by introducing `TrkrHitv3`, enabling studies of “carried-over” hit issues in streaming reconstruction.

## Key changes
- Added `TrkrHitv3` (`TrkrHitv3.h/.cc` and `TrkrHitv3LinkDef.h`): a `TrkrHitv2`-derived hit class that stores
  - `FPHXBCO` (`uint16_t`)
  - full `BCO` (`uint64_t`)
  with overridden setters/getters, `CopyFrom`, and ROOT dictionary support.
- Extended `TrkrHit` interface (`TrkrHit.h`) with new virtual timing payload APIs:
  - `setFPHXBCO/getFPHXBCO` (default `getFPHXBCO()` returns `std::numeric_limits<uint16_t>::max()`)
  - `setBCO/getBCO` (default `getBCO()` returns `0`)
- Updated `InttCombinedRawDataDecoder.cc` to construct `TrkrHitv3` hits and populate timing fields when creating new hits (in addition to existing ADC assignment logic).
- Updated `offline/packages/trackbase/Makefile.am` to include/install `TrkrHitv3` headers and build the `TrkrHitv3` ROOT dictionary/library alongside existing `TrkrHitv2` components.

## Potential risk areas
- **Serialization/IO compatibility:** introducing a new ROOT-serializable hit concrete type/version (`TrkrHitv3`) may affect reading old data or mixed-version workflows.
- **Reconstruction behavior changes:** downstream code that assumed only `TrkrHitv2` may need to handle `TrkrHitv3` and the presence/meaning of timing payloads.
- **Data completeness/validity:** callers must correctly interpret default/“unset” timing values (e.g., `FPHXBCO` defaulting to `max()`).

- Add targeted tests for `TrkrHitv3` copy/propagation, ROOT dictionary read/write round-trips, and mixed `TrkrHitv2`/`TrkrHitv3` compatibility.
- Add validation/monitoring paths to detect missing or inconsistent INTT timing information during decoding.

*Note: AI-generated summaries can contain mistakes—please use best judgment and cross-check against the actual implementation and integration tests.*

### DIFF
--- a/offline/packages/trackbase/TrkrHit.h
+++ b/offline/packages/trackbase/TrkrHit.h
@@ -15,6 +15,7 @@
 #include <climits>
 #include <cmath>
 #include <iostream>
+#include <limits>
 
 /**
  * @brief Base class for hit object
@@ -60,7 +61,7 @@ class TrkrHit : public PHObject
   // optional per-hit timing payload used by detectors that need to retain
   // the frontend bunch-counter value alongside the digitized hit.
   virtual void setFPHXBCO(const uint16_t) {}
-  virtual uint16_t getFPHXBCO() const { return 0; }
+  virtual uint16_t getFPHXBCO() const { return std::numeric_limits<uint16_t>::max(); }
   virtual void setBCO(const uint64_t) {}
   virtual uint64_t getBCO() const { return 0; }
   /*
--- a/offline/packages/trackbase/TrkrHit.h
+++ b/offline/packages/trackbase/TrkrHit.h
@@ -11,6 +11,7 @@
 
 #include <phool/PHObject.h>
 
+#include <cstdint>
 #include <climits>
 #include <cmath>
 #include <iostream>
@@ -55,6 +56,13 @@ class TrkrHit : public PHObject
   // after digitization, these are the adc values
   virtual void setAdc(const unsigned int) {}
   virtual unsigned int getAdc() const { return 0; }
+
+  // optional per-hit timing payload used by detectors that need to retain
+  // the frontend bunch-counter value alongside the digitized hit.
+  virtual void setFPHXBCO(const uint16_t) {}
+  virtual uint16_t getFPHXBCO() const { return 0; }
+  virtual void setBCO(const uint64_t) {}
+  virtual uint64_t getBCO() const { return 0; }
   /*
   virtual void setCrossing(const short int) {}
   virtual short int getCrossing() { return 0;}
--- a/offline/framework/frog/CreateFileList.pl
+++ b/offline/framework/frog/CreateFileList.pl
@@ -82,7 +82,10 @@ my %proddesc = (
     "41" => "Herwig Jet ptmin = 12 GeV",
     "42" => "Herwig Jet ptmin = 20 GeV",
     "43" => "Herwig Jet ptmin = 40 GeV",
-    "44" => "Herwig Jet ptmin = 50 GeV"
+    "44" => "Herwig Jet ptmin = 50 GeV",
+    "45" => "Herwig Photonjet ptmin = 5 GeV",
+    "46" => "Herwig Photonjet ptmin = 10 GeV",
+    "47" => "Herwig Photonjet ptmin = 20 GeV"
     );
 
 my %pileupdesc = (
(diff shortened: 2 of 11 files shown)


offline/packages/trackbase/TrkrHit.h
```diff
@@ -55,6 +56,13 @@ class TrkrHit : public PHObject
   // after digitization, these are the adc values
   virtual void setAdc(const unsigned int) {}
   virtual unsigned int getAdc() const { return 0; }
+
+  // optional per-hit timing payload used by detectors that need to retain
+  // the frontend bunch-counter value alongside the digitized hit.
+  virtual void setFPHXBCO(const uint16_t) {}
+  virtual uint16_t getFPHXBCO() const { return 0; }
```
osbornjd: 0 is a valid FPHXBCO, right (it can go from 0-120, or do I remember incorrectly)? In that case we should return something here like `std::numeric_limits<uint16_t>::max()`. Having 0 returned for `getBCO()` is probably fine since the actual BCO 0 was outside the lifetime of sPHENIX

osbornjd (commented):
Just a general question for my own understanding, where will this actually be used downstream? Is the idea to propagate this into the cluster object, or something else?

I guess the question is, you can either run over raw hits or clusters. Adding this into the offline TrkrHit object does not buy us anything unless it is needed in the clusterizer, for which maybe there are further development plans

offline/packages/trackbase/TrkrHit.h
ChengWeiShih: This would make it possible to make a plot such as the number of silicon seeds as a function of FPHXBCO (corresponding to the event interval), which would not be trivial in the past, as when we moved from INTTRawHit to TrkrHit, we lost the FPHXBCO information. 

So when I have a seed, I could then get the FPHXBCO from the INTT cluster(s) of a seed.

osbornjd: That is my confusion - the seeds point to clusters, not hits. So ultimately the FPHXBCO would also have to be passed up to the clusters in addition to being passed to the `TrkrHit`

ChengWeiShih: I think that would not be necessary; the following is what I did from my end. Or are you saying that we will delete the TrkrHit DSTs in the future? (The original idea is seed -> cluster -> TrkrHit)


```C++
if (TrkrDefs::getTrkrId(cluskey) == TrkrDefs::inttId)
            {
                ++nIntt;
                int first_hit_fphx_bco = std::numeric_limits<int>::min();
                if (clusterhitassoc && hitsetcontainer)
                {
                    TrkrDefs::hitsetkey hitsetkey_fromClus = TrkrDefs::getHitSetKeyFromClusKey(cluskey);
                    TrkrHitSet *hitset = hitsetcontainer->findHitSet(hitsetkey_fromClus);
                    if (hitset)
                    {
                        TrkrClusterHitAssoc::ConstRange hitrange = clusterhitassoc->getHits(cluskey);
                        if (hitrange.first != hitrange.second)
                        {
                            TrkrDefs::hitkey hitkey = hitrange.first->second;
                            TrkrHit *hit = hitset->getHit(hitkey);
                            if (hit)
                            {
                                first_hit_fphx_bco = static_cast<int>(hit->getFPHXBCO());
                            }
                        }
                    }
                }

                seed_INTT_timebucket.push_back(InttDefs::getTimeBucketId(cluskey));
                seed_INTT_FPHX_BCO.push_back(first_hit_fphx_bco);

                // silseed_cluster_FPHXBCO[silseed_cluster_FPHXBCO.size() - 1].push_back(first_hit_fphx_bco);
            }
```

osbornjd: What DSTs are you running over when you do this code? I assume it must be DSTs like `DST_STREAMING_EVENT_*`?

ChengWeiShih: In my case, my macro reads .evt files for testing the duplication algorithm. 

In this PR, the `InttCombineRawDataDecoder` is updated accordingly. If this PR can be merged, the TrkrHit DST from further production, using the updated `InttCombineRawDataDecoder`, would have the FPHXBCO information associated with the TrkrHit itself; reading `DST_STREAMING_EVENT_*` would not be needed.

osbornjd: It doesn't matter as long as you read in either `.evt` files or `DST_STREAMING` files, my concern was whether or not this will be needed downstream. i.e. at some point we will delete the .evt and streaming DSTs from disk, in which case you would lose access to this. I think for development purposes it is fine though

PR #4352, opened 2026-07-21
Geantino ionization and Acts fitter update

yuxdPKU:

- Adds `PHG4GeantinoIonization`, which assigns detector-dependent mean MIP energy deposition and ionization to charged-geantino G4 hits in MVTX, INTT, TPC, and TPOT before digitization.
- Uses configurable silicon stopping powers and gas-mixture-dependent stopping powers for TPC and TPOT.
- Keeps the normal downstream digitization fluctuations while intentionally neglecting Geant4 energy-loss straggling, delta electrons, and primary-ionization cluster structure.
- Updates `PHActsTrkFitter` to support directed fitting with an empty ACTS material map by falling back to measurement surfaces when no material surfaces are available.
- Adds an empty/single-surface guard and corrects the next-surface position used by the surface-ordering check.

This allows charged-geantino tracks to be digitized and fitted with ACTS while disabling ACTS material corrections.

## Types of changes

## TODOs (if applicable)

## Motivation

Charged-geantino hits were missing detector-appropriate mean MIP ionization before digitization, and ACTS directed fitting could fail when the material surface map was empty. This PR adds a geantino ionization pre-step and hardens `PHActsTrkFitter` so charged-geantino digitization/fitting works without relying on material surfaces.

## Key changes

- **Added `PHG4GeantinoIonization`** (`simulation/g4simulation/g4detectors`)
  - Introduces a `SubsysReco` that targets **charged-geantino** hits (matching truth particle name) and sets **detector-dependent mean** `edep/eion` in **MVTX, INTT, TPC, and Micromegas/TPOT**.
  - **Idempotent behavior**: if `edep/eion` are already finite and nonnegative, the hit is left unchanged (preserves existing downstream digitization fluctuations).
  - Computes `edep/eion` as **(mean MIP dE/dx) × path length**, using configurable **silicon MIP values** for MVTX/INTT and **TPC gas-mixture-dependent MIP** derived from TPC geometry parameters.
  - If `G4GEO_TPC` is missing, attempts to **rebuild `G4GEO_TPC` from `G4GEOPARAM_TPC`** (under `PAR/TPC`); **aborts the run** if required inputs/nodes are missing.
  - Intentionally omits higher-fidelity Geant4 effects (e.g., straggling/delta electrons/cluster structure), retaining downstream digitization fluctuations.

- **Build integration**
  - Updated `simulation/g4simulation/g4detectors/Makefile.am` to install the new header and compile/link `PHG4GeantinoIonization`.

- **Updated `PHActsTrkFitter`** (`offline/packages/trackreco/PHActsTrkFitter.cc`)
  - For **directed navigation**, if the ACTS material-based surface construction yields an **empty** surface list, it **falls back to measurement-derived surfaces** (`surfaces_tmp`).
  - Hardened `checkSurfaceVec`: it now **returns immediately when fewer than two surfaces** are provided.
  - Fixed/adjusted directed navigation surface ordering logic to avoid incorrect next-surface position handling during ordering checks.

## Potential risk areas

- **Reconstruction/physics behavior changes**
  - Changing charged-geantino `edep/eion` from sentinel behavior to mean MIP-based values can affect digitization and downstream tracking/QA performance.
- **Geometry/config dependency**
  - TPC gas-fraction ingestion (and potential rebuild of `G4GEO_TPC` from `G4GEOPARAM_TPC`) can fail or produce unexpected values if geometry parameter names/structure differ from assumptions.
- **Runtime behavior**
  - The TPC initialization path can **abort the run** if required TPC geometry inputs/nodes are missing.
- **Tracking efficiency impact**
  - The ACTS directed-fitting fallback + stricter surface-vector guards can alter fit behavior in edge cases; combined with the new geantino ionization, this could contribute to the reported QA efficiency reduction—worth investigating correlation with surface-map emptiness and with geantino hit composition.
- **Performance**
  - The geantino ionization module iterates through hit containers per enabled subsystem; even with idempotency checks, it adds extra per-event work.

- Add **unit/integration tests** for:
  - geantino idempotency and correct `edep/eion` synthesis,
  - TPC gas-fraction loading vs rebuild-from-parameters behavior,
  - ACTS directed fitting when material surfaces are empty and when only 0–1 surfaces are available.
- Validate default MVTX/INTT/TPC/Micromegas mean dE/dx constants against the digitization/hit-reco conventions used elsewhere in the simulation chain.
- Investigate QA efficiency regression by correlating:
  - which tracks used directed navigation/material surfaces,
  - how often the geantino ionization modified hits (vs left idempotent sentinels intact).

**Note:** AI can make mistakes—use best judgment and verify the summary against the actual diffs/implementation details during review.

### DIFF
--- a/simulation/g4simulation/g4detectors/PHG4GeantinoIonization.cc
+++ b/simulation/g4simulation/g4detectors/PHG4GeantinoIonization.cc
@@ -7,6 +7,9 @@
 
 #include <fun4all/Fun4AllReturnCodes.h>
 
+#include <phparameter/PHParameters.h>
+#include <phparameter/PHParametersContainer.h>
+
 #include <phool/getClass.h>
 #include <phool/phool.h>
 
@@ -16,15 +19,6 @@
 
 namespace
 {
-  struct GasFractions
-  {
-    double neon = 0;
-    double argon = 0;
-    double cf4 = 0;
-    double nitrogen = 0;
-    double isobutane = 0;
-  };
-
   // Pure-gas MIP stopping powers in keV/cm. They match the values used by
   // the current TPC and Micromegas hit-reconstruction modules.
   constexpr double neonMipDedx = 1.56;
@@ -39,17 +33,10 @@ namespace
   double mvtxMipDedx = 0.00384;
   double inttMipDedx = 0.00387;
 
-  GasFractions tpcGasFractions{0.00, 0.75, 0.20, 0.00, 0.05};
-  GasFractions tpotGasFractions{0.00, 0.90, 0.00, 0.00, 0.10};
-
-  double calculateMipDedx(const GasFractions& fractions)
-  {
-    return fractions.neon * neonMipDedx +
-           fractions.argon * argonMipDedx +
-           fractions.cf4 * cf4MipDedx +
-           fractions.nitrogen * nitrogenMipDedx +
-           fractions.isobutane * isobutaneMipDedx;
-  }
+  // PHG4MicromegasDetector and PHG4MicromegasHitReco both use a fixed
+  // Ar/isobutane 90/10 gas mixture.
+  constexpr double tpotMipDedx =
+      1e-6 * (0.9 * argonMipDedx + 0.1 * isobutaneMipDedx);
 }  // namespace
 
 PHG4GeantinoIonization::PHG4GeantinoIonization(const std::string& name)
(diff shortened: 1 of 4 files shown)


simulation/g4simulation/g4detectors/PHG4GeantinoIonization.cc
```diff
@@ -0,0 +1,229 @@
...
+#include <phool/getClass.h>
+#include <phool/phool.h>
+
+#include <cmath>
+#include <cstddef>
+#include <iostream>
+
+namespace
+{
+  struct GasFractions
```
osbornjd: Is there a reason to have a separate struct for gas fractions? We have these set at the macro level, which ultimately sets them in the actual geometry object

osbornjd: See e.g. [here](https://github.com/sPHENIX-Collaboration/macros/blob/690578c83fd8b24b405dfab7104290a424ecae39/common/G4_TrkrVariables.C#L177) which should override the values in the TPC geometry object (you have to rebuild it then, not read it from the CDB) [here](https://github.com/sPHENIX-Collaboration/coresoftware/blob/0e1300e316b6fe00ffd3a3eded16f96d35710447/simulation/g4simulation/g4tpc/PHG4TpcSubsystem.cc#L266-L270)

yuxdPKU: Thanks Joe. Now module reads gas fraction directly from parameter node tree

osbornjd:
This kills the tracking efficiency in the QA, we need to understand why

PR #4355, opened 2026-07-21
updated MB classifier for O+O and species dependence

pjwinnetou:

## Types of changes

## TODOs (if applicable)

## Motivation / context

Update the minimum-bias (MB) classifier to better support O+O collisions with species-dependent selection criteria, including optional pileup rejection, and address event-centrality consistency issues when an event fails the MB classifier.

## Key changes

- **MinimumBiasClassifier (offline/packages/trigger/MinimumBiasClassifier.cc / .h)**
  - **Species-dependent cut initialization** in `InitRun` using an `if/else-if` split:
    - For **O+O**: sets `m_max_charge_cut=400` (was 300), enables additional vertex/MBd-related thresholds (including `m_z_vtx_cut`), and initializes MBD charge/time cuts.
    - For **p+p**: keeps the simpler baseline setup while initializing the relevant MBD charge/time cuts.
  - **OO-only pileup rejection** in `FillMinimumBiasInfo`: when `m_reject_pileup` is enabled and total summed MBD charge exceeds `m_pileup_charge_cut`, `minbiascheck` is forced false.
  - **Correct Fun4All return-code propagation**: `process_event()` now returns `GetNodes(topNode)` non-`EVENT_OK` codes directly.
  - **Reduced node-missing logging**: removed an unnecessary log message when the `GLOBAL` node is absent.
  - **New configuration knob**: added `setRejectPileup(bool)` and the associated per-instance state (`m_reject_pileup`, `m_pileup_charge_cut`).
  - **Cut values moved to per-instance members** (instead of `static constexpr` constants) to support per-instance configuration.

- **Centrality consistency fixes**
  - `CentralityReco::process_event()` now calls **`m_central->Reset()` before** the MB check, preventing centrality from carrying over from the previous event when MB fails.
  - `CentralityInfov2::Reset()` is now properly implemented to reset base state and **clear `_centrality_bin_map`**, ensuring v2 centrality bin info doesn’t persist across resets.

## Potential risk areas

- **Physics/event selection impact**: changed MB thresholds for O+O and introduced OO pileup veto behavior can alter accepted event fractions and downstream physics yields.
- **Centrality output behavior changes**: centrality is now reset earlier—events failing MB will no longer retain previous-event centrality values.
- **Configuration/state safety**: per-instance cut parameters and the new `setRejectPileup` setter should be reviewed for any shared-instance/threading assumptions.
- **Behavior on missing nodes**: `process_event()` now propagates `GetNodes()` return codes directly, which may affect how failures are handled/aborted in the running framework.

- Move remaining “magic-number” MB cut defaults (e.g., pileup/MBD/vertex thresholds) into **CDB/calibration inputs** as requested by reviewers, rather than hardcoded per-instance values.
- Add **regression/unit tests** covering:
  - OO pileup rejection boundary conditions (`m_pileup_charge_cut` just above/below)
  - centrality reset behavior for MB-failing events
  - return-code handling for missing/invalid detector nodes.
- Validate MB efficiency and rejection rates for O+O against reference data/sim.

*Note: AI-generated summaries can contain errors—please use these details as a starting point and rely on direct code/physics validation when making decisions.*

### DIFF
--- a/offline/packages/centrality/CentralityReco.cc
+++ b/offline/packages/centrality/CentralityReco.cc
@@ -258,10 +258,13 @@ int CentralityReco::process_event(PHCompositeNode *topNode)
   }
 
   // Get Nodes from the Tree
-  if (GetNodes(topNode))
+  int ret = GetNodes(topNode);
+  if (ret != Fun4AllReturnCodes::EVENT_OK)
   {
-    return Fun4AllReturnCodes::ABORTRUN;
+    return ret;
   }
+
+  m_central->Reset();
 
   if (!m_mb_info->isAuAuMinimumBias())
   {
--- a/offline/packages/centrality/CentralityReco.cc
+++ b/offline/packages/centrality/CentralityReco.cc
@@ -264,8 +264,6 @@ int CentralityReco::process_event(PHCompositeNode *topNode)
     return ret;
   }
 
-  m_central->Reset();
-
   if (!m_mb_info->isAuAuMinimumBias())
   {
     return Fun4AllReturnCodes::EVENT_OK;
--- a/offline/packages/centrality/CentralityInfov2.cc
+++ b/offline/packages/centrality/CentralityInfov2.cc
@@ -11,6 +11,12 @@ void CentralityInfov2::identify(std::ostream &os) const
   return;
 }
 
+void CentralityInfov2::Reset()
+{
+  CentralityInfov1::Reset();
+  _centrality_bin_map.clear();
+}
+
 bool CentralityInfov2::has_centrality_bin(const PROP prop_id) const
 {
   return _centrality_bin_map.contains(prop_id);
--- a/offline/packages/centrality/CentralityInfov2.h
+++ b/offline/packages/centrality/CentralityInfov2.h
@@ -13,7 +13,7 @@ class CentralityInfov2 : public CentralityInfov1
   ~CentralityInfov2() override = default;
 
   void identify(std::ostream &os = std::cout) const override;
-  void Reset() override {}
+  void Reset() override;
 
   PHObject* CloneMe() const override { return new CentralityInfov2(*this); }
   void CopyTo(CentralityInfo *info) override;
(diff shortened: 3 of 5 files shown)


pinkenburg:
The GetNodes(topNode) method uses Fun4All return codes which then get reinterpreted to EVENT_OK in the process_event(). That is confusing and likely not what you want. In this case the classifier does not work and returning ok will just spew warnings for every single event (blowing up condor logs). If Fun4All return codes are used as return from GetNodes() they have to be passed up, otherwise change the return code to some classifier specific one and return the proper Fun4All return code from process_event()

many of the hardcoded parameters in this list
    m_useZDC          = true;
    m_box_cut         = true;
    m_hit_cut         = 2;
    m_max_charge_cut  = 2100;
    m_mbd_charge_cut  = 0.5;
    m_mbd_time_cut    = 25.;
    m_z_vtx_cut       = 60.;
    m_mbd_north_cut   = 10.;
    m_mbd_south_cut   = 150.;
    m_zdc_cut         = 60.;
are valid for all species but they are set individually for every species. That invites cut and paste errors - initialize them to a common default and only change the ones which are different. I would feel more comfortable if this would be offloaded to the cdb instead of using magic numbers
The printout in line 379 is not relevant, this is default behavior (if the detector node does not exist, create it) and shouldn't fill our logs. If this is unexpected - it should be reflected in the printout

bseidlit:
Can you fix the bug in the centrality reco while you are at it? The fact that the centrality reports the previous events centrality if the current event does not pass the minbias classifier.

pjwinnetou:
@pinkenburg I updated the structure. Could you take a look and see if this is okay to go? 
@bseidlit it seems it was a problem in the reset. I updated the centrality reco module

offline/packages/centrality/CentralityReco.cc
```diff
@@ -258,11 +258,14 @@ int CentralityReco::process_event(PHCompositeNode *topNode)
...
   // Get Nodes from the Tree
-  if (GetNodes(topNode))
+  int ret = GetNodes(topNode);
+  if (ret != Fun4AllReturnCodes::EVENT_OK)
   {
-    return Fun4AllReturnCodes::ABORTRUN;
+    return ret;
   }
 
+  m_central->Reset();
```
pinkenburg: This is not necessary. The Reset() is called for all objects on the node tree after an event is processed. The problem here was that the centrality info object implemented an empty reset method so the node tree reset did not reset it. This line here should be removed to avoid confusion

PR #4360, opened 2026-07-27
Acts extrapolation

hupereir:

This PR improves acts extrapolation in the TPC for measuring distortion corrections, ultimately aligning the code to what is done in GENFIT. 
4 extrapolation modes are introduced: 
1/ default - the current setup which uses the track parameters at origin for extrapolating
2/ forward - uses the track parameters closest, before the target extrapolation layer (in the TPC case this is the last layer of the INTT)
3/ bacward - uses the track parameters closest, after the target extrapolation layer, with backward extrapolation (in the TPC case this is the first TPOT layer, if available).
4/ bidirectional - uses the weighted average of the above two. This is consistent with what is done with GENFIT. 

For now the default is still 1/ so there should be no change to the users. 
I'll make a presentation at tomorrows Tracking meeting. (July 28)
Overall performances are illustrated in the plot below:
image

In addition: changed the meaning of "set_use_micromegas" in the GENFIT track fitter, to make it consistant with ACTS. 
When micromegas are not used, the corresponding clusters do not participate to the fit at all, whether present or not. 
@yuxdPKU this is redundant with a change you discussed last week and which you wanted to commit. 
Sorry if this create conflicts; I needed this for my tests. 

## Types of changes

## TODOs (if applicable)

## Motivation

Improve ACTS extrapolation for TPC distortion-correction measurements by adding mode options (matching GENFIT behavior more closely) while keeping the existing default behavior unchanged.

## Key changes

- Introduced four ACTS silicon-MM extrapolation modes in the TPC fitting flow:
  - **Default:** unchanged behavior using track parameters at the origin.
  - **Forward / Backward:** extrapolate using the nearest smoothed trajectory parameters before/after the target layer (with backward propagation in the backward mode).
  - **Bidirectional:** combine successful forward and backward results via a weighted average of parameters (with validation of surface/hypothesis and finite/valid covariance inputs) and consistent path-length handling.
- Updated ACTS propagation plumbing to support directional propagation and clearer layer/volume mapping:
  - `ActsPropagator::propagateTrack` now takes an `Acts::Direction`.
  - Centralized logging-level selection and refactored SPhenix↔ACTS layer/volume mapping with range-checked reverse mapping.
- Updated GENFIT Micromegas handling to exclude Micromegas from the fit when disabled:
  - Added/implemented `PHGenFitTrkFitter::set_use_micromegas(bool)` to disable the relevant Micromegas fit layers when `false`.
- Additional refactors/cleanup:
  - Minor formatting/initialization and internal construction changes (no intended logic changes) in related ACTS transformation/fitter code paths.

## Potential risk areas

- **Reconstruction behavior changes:** non-default extrapolation modes can alter fitted track states/resolutions; bidirectional mode introduces additional propagation attempts and parameter/path-length averaging.
- **Layer/volume mapping correctness:** updated ACTS↔SPhenix layer mapping and the new reverse mapping helper must be validated for all relevant geometries.
- **Performance/edge cases:** bidirectional mode performs extra propagation/averaging, which may increase runtime and fail more often in pathological configurations.
- **Thread-safety:** if extrapolation mode or Micromegas settings are modified concurrently with reconstruction, behavior may be sensitive (best validated in the actual job model).
- **No I/O format changes** are indicated by the provided changes.

- Add targeted validation for:
  - propagation success/failure rates by mode,
  - bidirectional averaging correctness (including covariance validity paths),
  - and mapping consistency across detector regions.
- Benchmark runtime impact across modes and confirm stability on large datasets.
- Provide documentation on recommended mode selection for distortion-correction workflows.

*AI-generated summaries can contain mistakes; please use best judgment and validate against the actual code changes and physics/regression tests.*

### DIFF
--- a/offline/packages/trackreco/ActsPropagator.cc
+++ b/offline/packages/trackreco/ActsPropagator.cc
@@ -17,40 +17,56 @@
 #include <Acts/Surfaces/PerigeeSurface.hpp>
 #include <Acts/Definitions/Units.hpp>
 
-/// local aborter class, used to tell acts to end track propagation when a given layer is used
-/** for the time being, the class is defined locally only, because it has no usage outside of ActsPropagator */
-struct ActsAborter
+namespace
 {
 
-  /// (ACTS) layer id at which propagation should stop
-  unsigned int abortlayer = std::numeric_limits<unsigned int>::max();
-
-  /// (ACTS) voulme id at which propagation should stop
-  unsigned int abortvolume = std::numeric_limits<unsigned int>::max();
-
-  /// called at each extrapolation step, by acts, to verify whether to stop propagation or not
-  template <typename propagator_state_t, typename stepper_t, typename navigator_t>
-    bool checkAbort(
-    propagator_state_t& state, const stepper_t& /*stepper*/,
-    const navigator_t& navigator, const Acts::Logger& /*logger*/) const
+  /// local aborter class, used to tell acts to end track propagation when a given layer is used
+  /** for the time being, the class is defined locally only, because it has no usage outside of ActsPropagator */
+  struct ActsAborter
   {
 
-    if (!navigator.currentSurface(state.navigation))
-    { return false; }
+    /// (ACTS) layer id at which propagation should stop
+    unsigned int abortlayer = std::numeric_limits<unsigned int>::max();
 
-    const auto& volumeno = state.navigation.currentSurface->geometryId().volume();
-    const auto& layerno = state.navigation.currentSurface->geometryId().layer();
-    const auto& sensitive = state.navigation.currentSurface->geometryId().sensitive();
+    /// (ACTS) voulme id at which propagation should stop
+    unsigned int abortvolume = std::numeric_limits<unsigned int>::max();
 
-    /// Check that we are in the proper layer and that we've also reached
-    /// a sensitive surface
-    if (layerno == abortlayer && volumeno == abortvolume && sensitive != 0)
-    { return true; }
+    /// called at each extrapolation step, by acts, to verify whether to stop propagation or not
+    template <typename propagator_state_t, typename stepper_t, typename navigator_t>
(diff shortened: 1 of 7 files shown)


pinkenburg:
resolved the merging conflict from my identical except an empty space clang-tidy fix

offline/packages/trackreco/ActsPropagator.cc
```diff
@@ -164,19 +182,16 @@ ActsPropagator::propagateTrack(const Acts::BoundTrackParameters& params,
   // initialize aborter
   options.actorList.get<ActsAborter>().abortlayer = actslayer;
   options.actorList.get<ActsAborter>().abortvolume = actsvolume;
+  options.direction = direction;
 
   auto result = propagator.propagate(params, options);
-
-  if (result.ok())
+  if( result.ok() && result.value().endParameters )
```
osbornjd, line 188: Did you find a situation where `result.ok()` returned true but there were no resulting track parameters returned?

osbornjd (commented):
Not for right now, but I wonder if there is a way to delegate some of the additional extrapolation code to another module (e.g. PHActsTrackStateProjector or something). It might not be possible if we need access to the MultiTrajectory which is only available in the KF, but I could see this being generically useful and it would then be delegated to some standalone module

offline/packages/trackreco/ActsPropagator.cc
hupereir: Nope but clang-tidy was complaining.

hupereir:
> Not for right now, but I wonder if there is a way to delegate some of the additional extrapolation code to another module (e.g. PHActsTrackStateProjector or something). It might not be possible if we need access to the MultiTrajectory which is only available in the KF, but I could see this being generically useful and it would then be delegated to some standalone module

I agree. Especially if we also want to update the extrapolation to the calorimeters.

PR #4388, opened 2026-08-11
add projected deltaZ(track,EMCal) in KFP

yuxdPKU:
Previous deltaZ(track,EMCal) uses Z at different radius for track and EMCal which makes two peaks in the deltaZ plot (first figure). To correct this bias, I add an additional branch track_EMCAL_DeltaZ_Projected which is calculated using EMCal Z projected to the same radius as track state linearly. deltaZ cut is applied on the projected delta Z. Second figure shows deltaZ distribution after projection. @cdean-github @viewolfe321 

conversion_track_EMCAL_DeltaZ_Raw
conversion_track_EMCAL_DeltaZ_Projected

## Types of changes

## TODOs (if applicable)

## Motivation

Raw track and EMCal Z values correspond to different radii. This creates a two-peak raw Δz distribution. The change projects the EMCal Z position to the track-state radius before applying the Δz cut. The projected method can be selected while preserving the previous method for comparison.

## Key changes

- Added the `track_EMCAL_DeltaZ_Projected` output branch.
- Added cylindrical projection of EMCal clusters.
- Added configurable projected-Δz cut setters.
- Added rejection of invalid intersections.
- Initialized unmatched projected Δz values to `NaN`.
- Expanded the Δz cut range to `[-100, 100]`.

## Potential risk areas

- The new branch changes the output I/O format.
- The projected method can change matching efficiency and reconstruction behavior.
- Projection edge cases can affect match selection.
- The additional calculation can increase processing time.
- Thread-safety should be verified for per-track storage.

- Add projection unit tests.
- Compare raw and projected matching efficiencies and physics distributions.
- Document the selection flag, geometry, and branch semantics.
- Validate downstream output compatibility.
- Study the projected method before making it the default.

AI-generated summaries can contain mistakes. Contributors should verify the implementation and reconstruction behavior against the code and requirements.

### DIFF
--- a/offline/packages/KFParticle_sPHENIX/KFParticle_truthAndDetTools.cc
+++ b/offline/packages/KFParticle_sPHENIX/KFParticle_truthAndDetTools.cc
@@ -64,9 +64,9 @@
 namespace
 {
 bool projectClusterZToCylinder(
-    const double vx, const double vy, const double vz,
-    const double cluster_x, const double cluster_y, const double cluster_z,
-    const double target_radius_sq, double &projected_cluster_z)
+    const float vx, const float vy, const float vz,
+    const float cluster_x, const float cluster_y, const float cluster_z,
+    const float target_radius_sq, float &projected_cluster_z)
 {
   // With delta = cluster - V, parameterize the ray from the primary vertex
   // V = (vx, vy, vz) through the EMCal cluster as
@@ -86,13 +86,13 @@ bool projectClusterZToCylinder(
   //   A = delta_x^2 + delta_y^2,
   //   B = 2 * (vx * delta_x + vy * delta_y),
   //   C = vx^2 + vy^2 - R_target^2.
-  const double delta_x = cluster_x - vx;
-  const double delta_y = cluster_y - vy;
-  const double delta_z = cluster_z - vz;
-  const double quadratic_a = delta_x * delta_x + delta_y * delta_y;
-  const double quadratic_b = 2. * (vx * delta_x + vy * delta_y);
-  const double quadratic_c = vx * vx + vy * vy - target_radius_sq;
-  const double discriminant = quadratic_b * quadratic_b - 4. * quadratic_a * quadratic_c;
+  const float delta_x = cluster_x - vx;
+  const float delta_y = cluster_y - vy;
+  const float delta_z = cluster_z - vz;
+  const float quadratic_a = delta_x * delta_x + delta_y * delta_y;
+  const float quadratic_b = 2. * (vx * delta_x + vy * delta_y);
+  const float quadratic_c = vx * vx + vy * vy - target_radius_sq;
+  const float discriminant = quadratic_b * quadratic_b - 4. * quadratic_a * quadratic_c;
   if (!(quadratic_a > 0.) || !(discriminant >= 0.))
   {
     return false;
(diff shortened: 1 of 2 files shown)


viewolfe321:
If I understand the changes correctly, this changes the default behavior. I think it would be good to make this change optional, at least until we have more time to study this method. Could you add a flag so that would allow us to apply the dz cut in the same manner as before?

yuxdPKU:
I add two new setters ``set_dz_projected_cut_low`` and ``set_dz_projected_cut_high``. Previous ``set_dz_cut_low`` and ``set_dz_cut_high`` still cut raw delta_z. But I think we should use projected delta_z because comparing z at different radius it not reasonable, as you can see the first plot which shows two peaks

viewolfe321:
I have given this a lot of thought this past week, and I think I disagree with the scope of this PR, and I don't think it should be merged. I think instead of removing a bias, it actually adds one by making the matching look better than it actually is.

The separation of peaks in the dz is an important QA tool, just as the separation in the dPhi peak is an important QA tool. They tell us how good or not good our projections are, and as we update how we do the track projection, these distributions will be more important than ever.

I might think differently if the phi was also projected. But as it is, I think this PR is forcing the dz to look better than it is. If we are worried about efficiency, we should just increase the range of allowed values in dz. 

I am happy to hear additional perspectives, but I have real concerns about this PR for the reasons stated above.

cdean-github:
HI Xudong, I'm curious but would we get a similar result if we project the track to the same radius? My guess if the track projection follows the curvature and should be more correct. At high pT the track is effectively straight but I'd think there would be a small effect especially at lower pT

yuxdPKU:
Hi Valerie and Cameron, thanks for your comments. I think they are helpful.
1. The reason why I choose 102.9 cm is based on the attached top left plot -- mean value of Gaussian fit to the delta-phi(track - EMCal) vs various projection radius, which was made one year ago. You can find that the discrepancy disappear between different charge (read and blue) if we use 102.9 cm as projection radius. For my understanding, the remaining global offset is due to TPC distortion which shift both charges to the same side.
2. Based on this tuning/test, to make delta-z comparison between track and EMCal reasonable, an additional projection is needed to project EMCal cluster to the same radius. Linear projection is a rough approximation.
3. As we know electron/positron hit EMCal and creates EM shower, the shower development and shape depends on track incident angle and energy. Not sure if this is needed to be taken into account in the track-calo matching, and how.
4. In the attached plot, I also show three addional plot -- sigma value of Gaussian fit to the delta-phi (track - EMCal) vs projection radius, and mean/sigma value of Gaussian fit to the delta-z (track - EMCal) vs projection radius. To calculate delta-z here, I do linear projection of EMCal cluster. This provides us more information about the choice of projection radius.
5. Can I propose to store both projected track state x/y/z on EMCal and matched EMCal x/y/z in the KFParticle ntuple, rather than only delta-phi and delta-z? In this way, we can study what is the best way to do the matching and decide if we do linear projection or not.

I also convert this PR to draft for now, and happy to hear your insights.

image

viewolfe321:
I think that the reasoning for 102.9 cm makes sense, even if more study of this is warranted. I nonetheless object to projecting the cluster for the dz comparison; I think doing so that would remove a diagnostic tool. I still think the shape of the dz distribution tells us information about the matching, the state of tracking, projection quality of the track state, etc. 

I agree with your suggestion to add additional info to the TTree. I am working on a PR now that happens to also address this. I will push it to the repo today, I just need to make sure it runs smoothly after adding some additional branches. 

Thank you for addressing my comment and for providing these plots! :)

yuxdPKU:
I close this PR and let Valerie to add additional information in another PR

PR #4422, opened 2026-09-08
CaloReco: support PhotonClusterBuilder BDT loading from CDB

JustinBen0131:
## Summary

- Add `set_bdt_model_cdb_domain` so PhotonClusterBuilder can load the BDT using the job's existing CDB global tag and timestamp.
- Preserve explicit file loading and BDT-disabled behavior; abort initialization if enabled BDT loading has no feature list or cannot load the requested model.
- Document both loading modes and the PPG12 feature order.

No changes to feature calculation, scoring, selection or default BDT enablement.

## Testing

Tested on SDCC with 1,000 pp data events. CDB and local-file loading gave identical scores for the 7 selected candidates. Also checked data/MC model loading and missing-model handling.

### DIFF
--- a/offline/packages/CaloReco/PhotonClusterBuilder.cc
+++ b/offline/packages/CaloReco/PhotonClusterBuilder.cc
@@ -26,6 +26,8 @@
 
 #include <fun4all/Fun4AllReturnCodes.h>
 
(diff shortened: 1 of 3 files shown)


pinkenburg:
The valgrind failure is not from this PR - please ignore

PR #4423, opened 2026-09-08
yc: use move directive to fix wrong memory allocation expanding vector

ycorrales:
This PR try to fix a memory issue seem recently in valgrind. 

## Motivation

Use move construction when adding `GBTLink` objects to prevent incorrect copies and address the Valgrind memory allocation issue observed during vector expansion.

## Key changes

- Added move constructor and move assignment to `mvtx::GBTLink`.
- Deleted copy constructor and copy assignment for `GBTLink`.
- Added `ClassDefNV(GBTLink, 1)` for ROOT dictionary support.
- Updated `setupLinks()` to construct a local `GBTLink` and move it into `mGBTLinks`.
- Applied namespace, include-order, and formatting cleanup.

## Potential risk areas

- Vector reallocation now depends on correct move semantics for all `GBTLink` members.
- Copy operations are no longer available for code that previously copied `GBTLink`.
- No intentional IO format or reconstruction behavior changes are indicated.
- The move path should be checked for thread-safety and performance under realistic event loads.

- Add focused tests for `GBTLink` move construction, move assignment, and vector reallocation.
- Run Valgrind and reconstruction tests before and after the change.
- Review ROOT dictionary generation and downstream `GBTLink` usage.

AI-generated summaries can contain mistakes. Contributors should verify these points against the complete diff and test results.

### DIFF
--- a/offline/framework/fun4allraw/mvtx_decoder/GBTLink.h
+++ b/offline/framework/fun4allraw/mvtx_decoder/GBTLink.h
@@ -6,130 +6,143 @@
 #ifndef MVTXDECODER_GBTLINK_H
 #define MVTXDECODER_GBTLINK_H
 
-#define _RAW_READER_ERROR_CHECKS_ // comment this to disable error checking
+#define _RAW_READER_ERROR_CHECKS_  // comment this to disable error checking
 
-#include "mvtx_utils.h"
-#include "RDH.h"
-#include "PayLoadCont.h"
-#include "PayLoadSG.h"
 #include "DecodingStat.h"
 #include "GBTWord.h"
 #include "InteractionRecord.h"
+#include "PayLoadCont.h"
+#include "PayLoadSG.h"
+#include "RDH.h"
 #include "StrobeData.h"
+#include "mvtx_utils.h"
 
+#include <iomanip>
 #include <iostream>
 #include <memory>
-#include <iomanip>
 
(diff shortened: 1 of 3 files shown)


pinkenburg:
I don't know if this was communicated - we have two versions of the mvtx decoder (each in an mvtx_decode subdir). One in this repo under offline/framework/fun4allraw and one in the online_distribution repo under newbasic. The filenames are identical but the sources differ. It looks like the memory problem came from the binary using the "wrong" version (likely the newbasic one) because some library ordering has changed. In my understanding there is a good reason why they have to be different but this needs to be sorted out (simple renaming of classes here?)

pinkenburg:
I'll merge this since it passed valgrind

ycorrales:
> I don't know if this was communicated - we have two versions of the mvtx decoder (each in an mvtx_decode subdir). One in this repo under offline/framework/fun4allraw and one in the online_distribution repo under newbasic. The filenames are identical but the sources differ. It looks like the memory problem came from the binary using the "wrong" version (likely the newbasic one) because some library ordering has changed. In my understanding there is a good reason why they have to be different but this needs to be sorted out (simple renaming of classes here?)

maybe we should change the name in newbasic? as you mentioned there are reason to have difference source, the core is the same but the difference rely in the random of event analyzed in the online case. if you think the issue is the same name then we can change the naming or use different namespace name. I think that would be more simple.

PR #4426, opened 2026-09-09
Fix mover bug

adfrawley:

## Types of changes

Simplify the logic for finding a new subsurface key after moving TPC clusters to a surface.
Tested with CA seeding with survey geometry, and polyseeding with ideal geometry.

## TODOs (if applicable)

## Motivation / context

This PR fixes subsurface-key lookup after TPC clusters move to a surface. It replaces iterative key updates with simpler, explicit handling.

## Key changes

- `processTrack` performs a single subsurface-key check after movement.
- `get_moved_position` returns the key for the new coordinates without modifying the cluster key.
- Update the cluster key only when the returned key changes.
- Add a warning when the updated key is inconsistent with the surface from the new coordinates.
- Tested with CA seeding and survey geometry, and with polyseeding and ideal geometry.

## Potential risk areas

- No IO format or public API changes are indicated.
- Reconstruction behavior can change when a moved cluster receives a different subsurface key.
- Geometry or key-mapping inconsistencies now produce warnings.
- No direct thread-safety change is apparent.

- Add regression tests for changed and unchanged subsurface keys.
- Define handling for key-to-surface consistency failures.
- Monitor warnings in reconstruction workflows.

AI-generated summaries can contain mistakes. Contributors should verify the implementation and reconstruction results.

### DIFF
--- a/offline/packages/tpc/TpcClusterMover.cc
+++ b/offline/packages/tpc/TpcClusterMover.cc
@@ -168,6 +168,7 @@ std::vector<std::pair<TrkrDefs::cluskey, Acts::Vector3>> TpcClusterMover::proces
     Acts::Vector3 global_new = global;
     TrkrDefs::subsurfkey new_subsurfkey = sskey;
     bool ret = get_moved_position(cluskey, cluster, fitpars, global, global_new, new_subsurfkey);
+
     if (!ret)
     {
       global_moved.emplace_back(cluskey, global);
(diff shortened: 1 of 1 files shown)


offline/packages/tpc/TpcClusterMover.cc
```diff
@@ -267,19 +245,10 @@ bool TpcClusterMover::get_moved_position(TrkrDefs::cluskey cluskey, TrkrCluster*
...
-    if (new_surf && (new_subsurfkey != sskey))
-    {
-      cluster->setSubSurfKey(new_subsurfkey);
-    }
-  }
-
+  // get the subsurface key for this new position and return it 
+  TrkrDefs::hitsetkey hkey = TrkrDefs::getHitSetKeyFromClusKey(cluskey);
+  auto new_surf = _tGeometry->get_tpc_surface_from_coords(hkey, global_new, new_subsurfkey);
+
```
hupereir, line 251: correct me if I am wrong: with the code above remove, neither hkey nor new_surf are used. Do I miss something ? Could this two lines be removed  ?

adfrawley:
The purpose of those two lines is to return "new_subsurfkey", which is checked in the caller to see if it changed. That is why the compiler did not complain ....

offline/packages/tpc/TpcClusterMover.cc
```diff
@@ -178,53 +179,30 @@ std::vector<std::pair<TrkrDefs::cluskey, Acts::Vector3>> TpcClusterMover::proces
...
-	if(!ret)
+    if(new_subsurfkey == sskey)
+      {
+	// we are done with this cluster, add the new position and surface to the return object
+	global_moved.emplace_back(cluskey, global_new);
+      }
+    else
+      {
+	// sskey changed, update the subsurface in the cluster key
+	cluster->setSubSurfKey(new_subsurfkey);
```
hupereir, line 191: I'm afraid that this line (necessary upstream, in e.g. MakeSourceLink, still introduces an inconsistency between the cluster internal, local coordinates and the stored subsurface, so that the next time one converts back to a global coordinates one ends up with an incorrect position. Or do I miss something ? In any case I should be able to test it soone.

hupereir:
> The purpose of those two lines is to return "new_subsurfkey", which is checked in the caller to see if it changed. That is why the compiler did not complain ....

Thanks ! I missed that the call to get_tpc_surface_from_coords indeed modifies new_subsurfkey

